PR #14, opened 2025-06-05
add tool for moving files

eyurtsev:
- need to add support for updating notebooks


pipeline/commands/build.py
```diff
@@ -9,7 +9,7 @@
 logger = logging.getLogger(__name__)
 
 
-def build_command(
+async def build_command(
```
eyurtsev:
```suggestion
def build_command(
```

PR #20, opened 2025-06-05
swap to yaml format for docs

eyurtsev:


pipeline/core/builder.py
```diff
@@ -101,6 +105,35 @@ def build_all(self) -> None:
             skipped_count,
         )
 
+    def _convert_yaml_to_json(self, yaml_file_path: Path, output_path: Path) -> None:
```
eyurtsev, line 108: Will generalize the handler later

src/docs.yml
```diff
@@ -0,0 +1,181 @@
+$schema: https://mintlify.com/docs.json
```
eyurtsev, line 1: cc @lnhsingh

PR #34, opened 2025-06-11
run lint on scripts

eyurtsev:


eyurtsev:
cc @lnhsingh running lint on the scripts

PR #109, opened 2025-08-04
LangGraph migration

lnhsingh:


pipeline/core/builder.py
```diff
@@ -127,6 +134,39 @@ def _convert_yaml_to_json(self, yaml_file_path: Path, output_path: Path) -> None
...
+            """Rewrite a single link match."""
+            pre = match.group(1)  # Everything before the URL
+            url = match.group(2)  # The URL
+            post = match.group(3)  # Everything after the URL
+
+            # Only rewrite absolute /oss/ paths
+            if url.startswith("/oss/"):
+                parts = url.split("/")
+                # Insert full language name after "oss"
+                parts.insert(2, self.language_url_names[target_language])
```
lnhsingh, line 160: is this deducing the target language from the fenching? what if a link is outside of a fence?

pipeline/core/builder.py
```diff
@@ -127,6 +134,39 @@ def _convert_yaml_to_json(self, yaml_file_path: Path, output_path: Path) -> None
...
+        if not target_language:
+            return content
+
+        def rewrite_link(match: re.Match) -> str:
+            """Rewrite a single link match."""
+            pre = match.group(1)  # Everything before the URL
+            url = match.group(2)  # The URL
+            post = match.group(3)  # Everything after the URL
+
+            # Only rewrite absolute /oss/ paths
```
lnhsingh: Noticed that this is also rewriting image paths, which it should ignore

pipeline/core/builder.py
beklapko: Links work inside and outside of the fencing. I think we kind of pre-set the languages and just swap them out depending on which version of the docs we're building at the time.

pipeline/core/builder.py
beklapko: Thank you for flagging this, pushing a fix now!

beklapko: Updated preview: https://langchain-5e9cc07a-preview-migrat-1755112825-437303e.mintlify.app/oss/python/overview

PR #116, opened 2025-08-05
better document what the first argument to a RemoteGraph can be

samecrowder:
when using RemoteGraph for a couple weeks, I noticed unexpected behavior that I could give either an assistant ID or a graph name as the first argument when instantiating a RemoteGraph.

Turns out this is by design, but it was not documented.

### DIFF
--- a/src/langgraph-platform/use-remote-graph.mdx
+++ b/src/langgraph-platform/use-remote-graph.mdx
@@ -14,7 +14,7 @@ sidebarTitle: Interact with the deployment using RemoteGraph
 
 When initializing a `RemoteGraph`, you must always specify:
 
-* `name`: the name of the graph you want to interact with. This is the same graph name you use in `langgraph.json` configuration file for your deployment. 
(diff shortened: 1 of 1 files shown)


src/langgraph-platform/use-remote-graph.mdx
```diff
@@ -14,7 +14,7 @@ sidebarTitle: Interact with the deployment using RemoteGraph
 
 When initializing a `RemoteGraph`, you must always specify:
 
-* `name`: the name of the graph you want to interact with. This is the same graph name you use in `langgraph.json` configuration file for your deployment. 
+* `name`: the name of the graph you want to interact with OR an assistant ID. If you specify a graph name, the default assistant will be used. If you specify an assistant ID, that specific assistant will be used. The graph name is the same name you use in `langgraph.json` configuration file for your deployment.
```
katmayb:
```suggestion
* `name`: the name of the graph you want to interact with **or** an assistant ID. If you specify a graph name, the default assistant will be used. If you specify an assistant ID, that specific assistant will be used. The graph name is the same name you use in `langgraph.json` configuration file for your deployment. 
```

PR #134, opened 2025-08-07
Base LangSmith content migration

katmayb:
This is the LangSmith docs migration over to Mintlify site.

A migration to the mintlify site of the current content, no big content changes.


katmayb:
The `legacy-trace-with-vercel-ai` page, I have moved over to the new site. Jacob mentioned that people may still need to be directed to this content, but we don't necessarily want it to be something that customers user generally.

katmayb:
Most up to date preview: https://langchain-5e9cc07a-preview-langsm-1755022380-de12be9.mintlify.app/langsmith/home

@lnhsingh can you review this? Mainly, so there is nothing included that is breaking 🙃 
I'm going to open a PR for hiding the nav as a follow-up now.

lnhsingh (approved):
Clicked through most of the nav links and some cross-reference links. Looks good from a high level!!

katmayb:
OK, thanks, going to merge this into the `main` branch and then check with the nav removal before pushing to `prod`

PR #138, opened 2025-08-07
Update LGP Deploy to Hybrid instructions

andrewnguonly:

PR #142, opened 2025-08-08
fix: remove mention of ECS support coming soon

samecrowder:
agreed with the infra folks (Mukil, Will) that we should remove this, bc we actually have no concrete plans to build this


katmayb:
@samecrowder  I think you can close your PR
@andrewnguonly has actually removed this ecs content in this PR: https://github.com/langchain-ai/docs/pull/138/files

samecrowder:
ah great!

PR #166, opened 2025-08-15
fix: delete wheel file that doesnt actually work

samecrowder:
my initial thought was users could download the .whl file from this page (since we don't want to publicize that on PyPi yet) but turns out that doesn't actually work


src/oss/langchain-agent-auth.mdx
```diff
@@ -10,17 +10,7 @@ LangChain Agent Auth solves the challenge of secure authentication for autonomou
...
-```bash
-pip install https://docs.langchain.com/langchain_auth-0.1.0-py3-none-any.whl[langgraph]
-```
-
-For use without LangGraph:
-
-```bash
-pip install https://docs.langchain.com/langchain_auth-0.1.0-py3-none-any.whl
-```
+The client library is not yet publicly available. To request access, please reach out to LangChain support. It will be released on PyPI soon.
```
lnhsingh: Should we include the support email here?

samecrowder: done

lnhsingh (approved):

PR #167, opened 2025-08-15
feat: language aware config step + hana changes

hntrl:
This adds a build step that has smarter handling for docsets with code fences, and adds a build step to docs.json to more intuitively organize content

* We now automatically expand docsets that have code fenced content (meaning we don't need to hardcode two different references for python + js in docs.json)
* When we encounter a directory that has code fenced content, instead of flagging it as a version we add it as a dropdown to better align with the design enhancements we want to make
* When we encounter a doc that has code fences, we flag the top level directory as needing conditional parsing (instead of hard coding which directories should be parsed)
  * If we were to add code fences to lgp or LS docs today, we would automatically start building js + python docs the same way we do for LG


beklapko (approved):

lnhsingh:
I think we can close this PR for now since CSS changes have been merged into main. The builder script changes can be added later

PR #162, opened 2025-08-14
Test auto preview branch creation

beklapko:
https://linear.app/langchain/issue/DOC-54/automate-preview-build-in-docs

PR #169, opened 2025-08-15
Move PR write permissions

beklapko:
Incorporating changes from Eugene: https://github.com/langchain-ai/docs/pull/162/files#r2278029839

### DIFF
--- a/.github/workflows/create-preview-branch.yml
+++ b/.github/workflows/create-preview-branch.yml
@@ -36,8 +36,10 @@ jobs:
         id: validate
         run: |
           set -euo pipefail
+          echo "[INFO] Validating current branch context"
           echo "[INFO] Current ref: ${{ github.ref }}"
           echo "[INFO] Branch name: $SOURCE_BRANCH"
+          # Validate source branch contains only safe characters and isn't too long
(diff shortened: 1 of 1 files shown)


.github/workflows/create-preview-branch.yml
```diff
@@ -14,10 +14,11 @@ concurrency:
 permissions:
   contents: write    # Required to checkout, create branches, and push
   actions: read      # Required to run the workflow
-  pull-requests: write  # Required to comment on PRs
 
 jobs:
   create-preview:
+    permissions:
+      pull-requests: write # Required to comment on PRs
```
beklapko: @eyurtsev I moved this to the job level because I didn't think there was a way to set permissions on an individual step. Is this what you were thinking or was there a different solution?

beklapko: Not sure this change is the cause yet since it didn't fail on the step for commenting on the PR but the preview workflow failed: https://github.com/langchain-ai/docs/actions/runs/16992739025/job/48175935263?pr=169#step:11:892

lnhsingh (approved):

PR #184, opened 2025-08-19
chore(lgp): Remove langsmith runs endpoints env var

angus-langchain:


src/langgraph-platform/env-var.mdx
```diff
@@ -54,13 +54,18 @@ For example, if a deployment is scaled up to 10 replicas and `LANGGRAPH_POSTGRES
 
 Defaults to `150` connections.
 
-## `LANGSMITH_RUNS_ENDPOINTS`
```
andrewnguonly, line 57: Are we just removing the documentation or the functionality has been removed from LangSmith SDK?

If this is removed from the SDK, then this might break Betclic (and some other users). Chat LangChain uses this as well.

Chat LangChain: https://smith.langchain.com/o/ebbaf2eb-769b-4505-aca2-d11de10372a4/host/deployments/48211402-480b-44a7-b46f-3e460db85b00
image

src/langgraph-platform/env-var.mdx
```diff
@@ -54,13 +54,18 @@ For example, if a deployment is scaled up to 10 replicas and `LANGGRAPH_POSTGRES
...
-
+## `LANGSMITH_ENDPOINT`
```
andrewnguonly, line 57: Right now, both `LANGSMITH_ENDPOINT` and `LANGSMITH_API_KEY` are reserved and are blocked (API will return 400) from being set in the control plane: https://github.com/langchain-ai/langchainplus/blob/main/host-backend/host/models/env_var.py#L112

If we want to allow this for self-hosted deployments, they should be added here: https://github.com/langchain-ai/langchainplus/blob/main/host-backend/host/models/env_var.py#L158

PR #213, opened 2025-08-22
Update tabs to code groups in LangSmith

katmayb:
This PR updates all of the Tabs to CodeGroups in the LangSmith directory to enable code tab syncing for users.


lnhsingh (approved):
@katmayb need to update /langsmith/trace-openai to use codegroups but otherwise it looks good!

PR #245, opened 2025-08-28
LangSmith docs updates

tanushree-sharma:
Big changes:
- Added images to the home page
- Moved concepts from get started to the feature specific pages
- Renamed evaluation techniques -> evaluation approaches -- **changed the file name so redirect needs a change**

Latest preview URL: https://langchain-5e9cc07a-preview-lschan-1756361888-2e31b40.mintlify.app/oss/python/overview


katmayb (approved):
This looks good to me! I think we should work on the size of this image, as it presents pretty small. However, I'm going to merge because that's a nit and we can drop in a new image.

@lnhsingh Tanushree has moved the conceptual pieces under the respective categories as we were musing on yesterday, so that's great!

PR #258, opened 2025-08-28
docs(oss): add JS example to quickstart page

christian-bromann:
Adding JS compliant example for quick start guide.

### DIFF
--- a/src/oss/langchain-quickstart.mdx
+++ b/src/oss/langchain-quickstart.mdx
@@ -7,13 +7,13 @@ title: Quickstart
 Super quick start
 :::python
 ```python
-from langgraph.prebuilt import create_agent
+from langchain.agents import create_react_agent
 
 def get_weather(city: str) -> str:
     """Get weather for a given city."""
     return f"It's always sunny in {city}!"
 
-agent = create_agent(
+agent = create_react_agent(
     model="anthropic:claude-3-7-sonnet-latest",
     tools=[get_weather],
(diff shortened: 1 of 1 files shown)


src/oss/langchain-quickstart.mdx
```diff
@@ -109,23 +122,102 @@ agent = create_agent(
...
+  "2": "SF",
+} as const;
+
+const getUserInfo = tool(
+  (_, config) => {
+    const { user_id } = config.context as {
+      user_id: keyof typeof USER_LOCATION;
+    };
+    console.log("user_id", config.context);
+    return USER_LOCATION[user_id];
```
hntrl, line 180: nit (aiming for types to be more readable):
```suggestion
const USER_LOCATION = {
  "1": "Florida",
  "2": "SF",
};

const getUserInfo = tool(
  (_, config) => {
    const { user_id } = config.context;
    console.log("user_id", config.context);
    return USER_LOCATION[user_id];
```

hntrl (approved):

src/oss/langchain-quickstart.mdx
christian-bromann: @hntrl this would fail because of:

> Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ "1": string; "2": string; }'.ts(7053)

Unfortunately `config.context` is typed with `Record<string, any>`. The tool doesn't know the context schema. It's on my wish list to improve 😉

PR #290, opened 2025-08-30
chore(oss): fix some js translations in messages

hntrl:


hwchase17:
i think i got all the changes

PR #342, opened 2025-09-02
oss: add markdownlint to `Makefile` and run on repo

mdrxy:


.markdownlint.json
```diff
@@ -0,0 +1,8 @@
+{
```
mdrxy, line 1: - "default": true - Enables all default markdownlint rules
- "MD001": false - Disables "Heading levels should only increment
  by one level at a time"
- "MD013": false - Disables "Line length" (usually 80 char limit)
- "MD025": false - Disables "Multiple top level headings in the
same document"
- "MD033": false - Disables "Inline HTML" (allows HTML tags in
markdown)
- "MD041": false - Disables "First line in file should be a top
level heading"

.vscode/settings.json
```diff
@@ -85,13 +85,7 @@
...
   // Disable format on save to prevent unwanted line breaks
   "editor.formatOnSave": false,
-  "[markdown]": {
-    "editor.formatOnSave": false
-  },
-  "[mdx]": {
-    "editor.formatOnSave": false
-  },
-  
+
```
mdrxy, line 94: (removed and then restored in later commit)

.markdownlint.json
beklapko: For MD001 and MD025, I think this might be OK but could possibly change depending on how we end up writing our style guide. I'm kind of assuming that we'll be setting titles in frontmatter and then all headers in the body of the doc will be H2, H3, H4, etc. Not super sure though.

PR #351, opened 2025-09-03
Clarify that creating the listener in the UI does not actually install the listener in the K8s cluster

andrewnguonly:
### Summary
Feedback from user:
```
I added a listener about 20 minutes ago. i don't see anything new in my K8s deployment related to this and the docs are ver sparse. i don't actually know what this is doing or how it works. i just see the docs here, i assume this is the right docs?
https://docs.langchain.com/langgraph-platform/hybrid#listeners
but i thought there was supposed to be a UI or something?
i also looked for any K8s resource with these IDs anywhere and i don't see anything in my cluster :confused:

i got more info on this, all good! the docs linked in the UI were not correct. the support team gave me the right docs while i was on the call for the other issue :slightly_smiling_face: so we are good here i need to go through the docs and add to helm, i'm doing that now
```

There might be some confusion about the relationship between creating a listener in the LangSmith UI and installing the listener in the K8s cluster. I'm adding an explicit clarification about it.

image


src/langgraph-platform/deploy-hybrid.mdx
```diff
@@ -50,7 +54,7 @@ The Hybrid deployment option requires an [Enterprise](/langgraph-platform/plans)
         langsmithWorkspaceId: "" # Workspace ID
         hostBackendUrl: "https://api.host.langchain.com" # Only override this if on EU
         smithBackendUrl: "https://api.smith.langchain.com" # Only override this if on EU
-        langgraphListenerId: "" # Listener ID from Step 2-6
+        langgraphListenerId: "" # Listener ID from Step 2f
```
andrewnguonly, line 57: The site is now rendering `a-z` for the sub-bullet points (preferred). It was previously `1-9`.

katmayb (approved):

PR #353, opened 2025-09-03
Add script to populate python integrations landing page

ccurme:
Migrated from the langchain monorepo: https://github.com/langchain-ai/langchain/blob/master/docs/scripts/partner_pkg_table.py

Updated for compatibility with new docs repo

Resolves broken links in the existing table

### DIFF
--- a/pipeline/tools/partner_pkg_table.py
+++ b/pipeline/tools/partner_pkg_table.py
@@ -151,7 +151,7 @@ These providers have standalone `langchain-{{provider}}` packages for improved v
 
 {table()}
 
-## All Providers
+## All providers
 
 Click [here](/oss/integrations/providers/all_providers) to see all providers or search
 for a provider using the search field.


pipeline/tools/partner_pkg_table.py
```diff
@@ -0,0 +1,165 @@
...
+title: Integration packages
+---
+
+<Info>
+
+If you'd like to contribute an integration, see [Contributing integrations](/oss/integrations/contributing).
+
+</Info>
+
+## Integration Packages
```
beklapko:
```suggestion
## Integration packages
```

pipeline/tools/partner_pkg_table.py
```diff
@@ -0,0 +1,165 @@
...
+
+These providers have standalone `langchain-{{provider}}` packages for improved versioning, dependency management and testing.
+
+{table()}
+
+## All Providers
```
beklapko:
```suggestion
## All providers
```

pipeline/tools/partner_pkg_table.py
beklapko: @ccurme do we need to remove this header and the "All providers" one from the mdx file? I wasn't sure if the script would dupe them or not.

PR #373, opened 2025-09-04
add uv equivalents to pip install commands

mdrxy:
Added a script to automate & standardized

Manually reviewed all changes

PR #376, opened 2025-09-04
fix broken build

eyurtsev:


mdrxy:
fixed in #373, merging in there

PR #378, opened 2025-09-04
chore(langsmith): add cloudflare tags

hntrl:


katmayb (approved):

src/langsmith/trace-without-env-vars.mdx
```diff
@@ -11,7 +11,7 @@ As mentioned in other guides, the following environment variables allow you to c
 * `LANGSMITH_ENDPOINT`
 * `LANGSMITH_PROJECT`
 
-In some environments, it is not possible to set environment variables. In these cases, you can set the tracing configuration programmatically.
+If you want to trace some runs with different configuration than others, your environment doesn't support typical environment variables (such as Cloudflare Workers), or you'd prefer not to rely on environment variables directly, LangSmith allows you the tracing configuration programmatically.
```
katmayb:
```suggestion
If you want to trace some runs with different configuration than others, your environment doesn't support typical environment variables (such as Cloudflare Workers), or you'd prefer not to rely on environment variables directly, LangSmith allows you to set the tracing configuration programmatically.
```

Looked like some words were missing, not sure if the suggestion was what you were going for?

PR #382, opened 2025-09-05
docs(oss): add langchain/production test docs

casparb:
Add docs on testing `create_agent()`, covering:
* Unit testing
  * Basic mock model
  * Testing stateful logic that doesn't necessitate a network call
* Intergration testing with [AgentEvals](https://github.com/langchain-ai/agentevals)
  * Agent trajectory matching
  * LLM-as-a-judge
* Integration with LangSmith pytest/vitest/jest
* Caching with `vcrpy`

TODO
* @christian-bromann christian-bromann to check js snippets

### DIFF
--- a/src/oss/langchain-test.mdx
+++ b/src/oss/langchain-test.mdx
@@ -1,3 +1,931 @@
 ---
-title: (Coming Soon) Test
+title: Test
 ---
+
+Agentic applications give an LLM freedom over control flow in order to solve problems. While this freedom can be extremely powerful, the black box nature of LLMs can make it difficult to understand how changes in one part of your agent will affect others downstream. This makes testing your agents especially important.
+
+## Unit testing
+
+For logic that doesn't require an API call, you can replace the real LLM with an in-memory stub. LangChain provides fake models for mocking responses.
+
+:::python
+LangChain provides `GenericFakeChatModel` for mocking text responses. It accepts an iterator of responses and returns one per invocation. It supports both regular and streaming usage.
+
+### Basic invocation
+
+We can pre-load three responses into the model and test 
+
+```python
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain.agents import create_agent
+
+model = GenericFakeChatModel(messages=iter(["foo", "bar"]))
+
+agent = create_agent(model, tools=[], prompt="Echo: {input}")
+
+def test_prompt_formatting():
+    result = agent.invoke({"messages": [
+        {"role": "user", "content": "hello"}
+    ]})
+    assert result["messages"][-1].content_blocks == [{'type': 'text', 'text': 'foo'}]
+
+    result = agent.invoke({"messages": [
+        {"role": "user", "content": "hello, again!"}
+    ]})
+    assert result["messages"][-1].content_blocks == [{'type': 'text', 'text': 'bar'}]
+```
+:::
+
+:::js
+LangChain provides `FakeChatModel` for mocking text responses. It accepts an array of responses and returns one per invocation.
+
+### Basic invocation
+
+We can pre-load responses into the model and test:
+
+```typescript
+import { FakeChatModel } from "@langchain/core/utils/testing";
+import { createAgent } from "langchain";
+
+const model = new FakeChatModel({
+  responses: ["foo", "bar"]
+});
+
+const agent = createAgent({
+  llm: model,
+  tools: [],
+  prompt: "Echo: {input}"
+});
+
+async function testPromptFormatting() {
+  const result = await agent.invoke({
+    messages: [{ role: "user", content: "hello" }]
+  });
+  expect(result.messages[result.messages.length - 1].contentBlocks).toBe([{type: "text", text: "foo"}]);
+
+  const result2 = await agent.invoke({
+    messages: [{ role: "user", content: "hello, again!" }]
+  });
+  expect(result2.messages[result2.messages.length - 1].contentBlocks).toBe([{type: "text", text: "bar"}]);
+}
+```
+:::
+
+### Testing stateful logic
+
+You can simulate multiple turns to test state-dependent behavior, such as switching models based on conversation length.
+
+In this example, we configure a `checkpointer` to persist the agent's state between invocations. This allows us to simulate the multiple conversation turns that might occur during a live invocation, where the model can call tools, respond to human interrupts, and so on.
+
+:::python
+```python
+from itertools import repeat
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain.agents import create_agent, AgentState
+from langchain_core.messages import AIMessage
+from langgraph.runtime import Runtime
+from langgraph.checkpoint.memory import InMemorySaver
+
+small_model = GenericFakeChatModel(messages=repeat("small response", 3))
+large_model = GenericFakeChatModel(messages=repeat("large response", 1))
+
+def choose_model(state: AgentState, runtime: Runtime):
+    """Choose model based on conversation complexity."""
+    messages = state["messages"]
+    ai_message_count = sum(isinstance(m, AIMessage) for m in messages)
+
+    print(messages)
+    print(ai_message_count)
+
+    return large_model if ai_message_count > 2 else small_model
+
+agent = create_agent(
+    choose_model, 
+    tools=[],
+    checkpointer=InMemorySaver()
+)
+
+def test_model_switching():
+    # Simulate 4 conversation turns
+    for i in range(4):
+        result = agent.invoke(
+            {"messages": [{"role": "user", "content": f"message {i}"}]}, 
+            {"configurable": {"thread_id": "1"}}
+        )
+    
+    assert result["messages"][-1].content_blocks == [{'type': 'text', 'text': 'large response'}]
+```
+:::
+
+:::js
+```typescript
+import { FakeChatModel } from "@langchain/core/utils/testing";
+import { createAgent } from "langchain"
+import { AIMessage } from "@langchain/core/messages";
+import { MemorySaver } from "@langchain/langgraph";
+
+const smallModel = new FakeChatModel({
+  responses: Array(3).fill("small response")
+});
+
+const largeModel = new FakeChatModel({
+  responses: ["large response"]
+});
+
+function chooseModel(state: any): any {
+  const messages = state.messages;
+  const aiMessageCount = messages.filter((m: any) => m instanceof AIMessage).length;
+
+  console.log(messages);
+  console.log(aiMessageCount);
+
+  return aiMessageCount > 2 ? largeModel : smallModel;
+}
+
+const agent = createAgent({
+  llm: chooseModel,
+  tools: [],
+  checkpointer: new MemorySaver()
+});
+
+async function testModelSwitching() {
+  // Simulate 4 conversation turns
+  for (let i = 0; i < 4; i++) {
+    const result = await agent.invoke(
+      { messages: [{ role: "user", content: `message ${i}` }] },
+      { configurable: { thread_id: "1" } }
+    );
+  }
+  
+  expect(result.messages[result.messages.length - 1].contentBlocks).toBe([{type: "text", text: "large response"}]);
+}
+```
+:::
+
+## Integration testing
+
+Many agent behaviors only emerge when using a real LLM, such as which tool the agent decides to call, how it formats responses, or whether a prompt modification affects the entire execution trajectory. LangChain offers an `agentevals` package that provides evaluators specifically designed for testing agent trajectories with live models.
+
+AgentEvals lets you easily evaluate the trajectory of your agent (i.e. the agent's message history) by matching it against a [reference trajectory](#trajectory-match-evaluator) or by using an [LLM judge](#llm-as-judge-evaluator).
+
+### Installing agentevals
+
+:::python
+```bash
+pip install agentevals
+```
+:::
+
+:::js
+```bash
+npm install agentevals @langchain/core
+```
+:::
+
+Or, clone the [AgentEvals repo](https://github.com/langchain-ai/agentevals) directly.
+
+### Trajectory match evaluator
+
+:::python
+AgentEvals offers the `create_trajectory_match_evaluator` function to match your agent's trajectory against a reference trajectory. You can customize its behavior in a few ways:
+
+- Setting `trajectory_match_mode` to [`strict`](#strict-match), [`unordered`](#unordered-match), [`subset`](#subset-and-superset-match), or [`superset`](#subset-and-superset-match) to provide the general strategy the evaluator will use to compare trajectories
+- Setting [`tool_args_match_mode`](#tool-args-match-modes) and/or [`tool_args_match_overrides`](#tool-args-match-modes) to customize how the evaluator considers equality between tool calls in the actual trajectory vs. the reference. By default, only tool calls with the same arguments to the same tool are considered equal.
+:::
+
+:::js
+AgentEvals offers the `createTrajectoryMatchEvaluator` function to match your agent's trajectory against a reference trajectory. You can customize its behavior in a few ways:
+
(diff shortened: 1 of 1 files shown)


src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
 ---
-title: (Coming Soon) Test
+title: Test
 ---
+
+Agentic applications give an LLM freedom over control flow in order to solve problems. While this freedom can be extremely powerful, the black box nature of LLMs can make it difficult to understand how changes in one part of your agent will affect others downstream. This makes testing your agents especially important.
```
mdrxy: the first two sentences are wordy/don't read super smooth

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
...
+
+## Unit testing
+
+For logic that doesn't require an API call, you can replace the real LLM with an in-memory stub. LangChain provides fake models for mocking responses.
```
mdrxy:
```suggestion
For logic not requiring API calls, you can use an in-memory stub for mocking responses.
```

mdrxy: "LangChain provides fake models for mocking responses" the following snippets showcase this

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
 ---
-title: (Coming Soon) Test
+title: Test
 ---
+
+Agentic applications give an LLM freedom over control flow in order to solve problems. While this freedom can be extremely powerful, the black box nature of LLMs can make it difficult to understand how changes in one part of your agent will affect others downstream. This makes testing your agents especially important.
+
```
eyurtsev: Right here before unit testing I'd include a few more sentences or a comparison to explain what the approaches are. 

The jump straight into unit testing is too fast if folks aren't familiar with testing approaches.

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
...
+We can pre-load three responses into the model and test 
+
+```python
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain.agents import create_agent
+
+model = GenericFakeChatModel(messages=iter(["foo", "bar"]))
+
+agent = create_agent(model, tools=[], prompt="Echo: {input}")
```
eyurtsev: I'd probably show how GenericFakeChatModel works on its own before showing how to use it within an agent. 

Users may want to mock an LLM potentially to test out code that doesn't involve agents.

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
...
+
+small_model = GenericFakeChatModel(messages=repeat("small response", 3))
+large_model = GenericFakeChatModel(messages=repeat("large response", 1))
+
+def choose_model(state: AgentState, runtime: Runtime):
+    """Choose model based on conversation complexity."""
+    messages = state["messages"]
+    ai_message_count = sum(isinstance(m, AIMessage) for m in messages)
+
+    print(messages)
```
eyurtsev: prints usually do not go into any production code or even test code. So only use them if they're really necessary for demonstrating an output (i don't believe this is the case here)

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
...
+Results will be automatically logged to LangSmith.
+
+<Tip>
+To learn more about evaluating your agent, see the [LangSmith docs](/langsmith/vitest-jest).
+</Tip>
+:::
+
+## Caching
+
+Integration tests that call real LLM APIs can be slow and expensive, especially when run frequently in CI/CD pipelines. We recommend using a caching library to recording HTTP requests and responses during the first test run, then replaying them in subsequent runs without making actual network calls.
```
eyurtsev:
```suggestion
Integration tests that call real LLM APIs can be slow and expensive, especially when run frequently in CI/CD pipelines. We recommend using a library for recording HTTP requests and responses, then replaying them in subsequent runs without making actual network calls.
```

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,931 @@
 ---
```
eyurtsev: The content is overwhelming as presented currently.

We need an overview that explains the different approaches possible.

For example, is llm as a judge for integration testing or for benchmarking?

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1002 @@
...
+  expect(result.messages[result.messages.length - 1].contentBlocks).toBe([{type: "text", text: "large response"}]);
+}
+```
+:::
+
+## Integration testing
+
+Many agent behaviors only emerge when using a real LLM, such as which tool the agent decides to call, how it formats responses, or whether a prompt modification affects the entire execution trajectory. LangChain offers an `agentevals` package that provides evaluators specifically designed for testing agent trajectories with live models.
+
+AgentEvals lets you easily evaluate the trajectory of your agent (i.e. the agent's message history) by performing a **trajectory match** or by using an **LLM judge**:
```
lnhsingh: Link to additional AgentEvals docs/repo?

casparb: good catch

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1002 @@
...
+title: Test
 ---
+
+Agentic applications let an LLM decide its own next steps to solve a problem. That flexibility is powerful, but the model's black-box nature makes it hard to predict how a tweak in one part of your agent will affect the rest. To build production-ready agents, thorough testing is essential.
+
+There are a few approaches we will cover to test your agents:
+- **Unit tests** exercise small, deterministic pieces of your agent in isolation using in-memory fakes so you can assert exact behavior quickly and cheaply.
+- **Integration tests** validate how components interact under real conditions, testing the agent using real network calls (live LLM endpoints, retrievers, and external APIs) to confirm that components work together, credentials and schemas line up, and latency is acceptable.
+
+Agentic applications tend to lean more on integration because they chain multiple components together and must deal with the nondeterministic nature of LLMs, where outputs can vary across runs. To keep these tests cost-effective, we will also cover how to record and replay responses (see [Caching](#caching)).
```
eyurtsev: 1. Highlighting of Caching here makes the caching section look more important than the other sections (visually).
2. Can also condense the text (e.g., probably not necessary to say that it keeps tests cost-effective. cost might not even be an issue, the issue might be time it takes to run the tests or flakiness due to non deterministic behavior)

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1002 @@
...
+
+## Unit testing
+
+For logic not requiring API calls, you can use an in-memory stub for mocking responses.
+
+:::python
+LangChain provides `GenericFakeChatModel` for mocking text responses. It accepts an iterator of responses and returns one per invocation. It supports both regular and streaming usage.
+
+### Basic invocation
```
eyurtsev: The header title doesn't work fit for the information architecture.

Information architecture looks like:

unit testing -> basic invocation

and it's not clear that this is referring to an example with a chat model.

One of the ways I find useful to think about the information architecture for a given page is by looking at the TOC (the table on the right side) and considering whether a person would be able to guess what the section is about or jump to the correct section without having read any content. 

---

I'd probably instead make a header called "Fake Chat Model" or "Mocking Chat Model" and then briefly explain the interface for this implementation or provide a usage example.

We should also cross link to the API reference for the implementation

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+```
+
+If we invoke the model again, it will return the next item in the iterator:
+
+```python
+model.invoke("hello, again!")
+# AIMessage(content='bar', additional_kwargs={}, response_metadata={}, id='lc_run--...')
+```
+
+### Agent with fake model
```
eyurtsev: This example uses a fake chat model and an agent, but it's not clear what the test is attempting is testing or whether it needs to be tested?

Is the goal to test that the prompt is being formatted properly? If so, I don't think that this unit test verifies that.

Instead we'd need to test that the inputs into the chat model from the agent where what we expected.

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+
+  const result2 = await agent.invoke({
+    messages: [{ role: "user", content: "hello, again!" }]
+  });
+  expect(result2.messages[result2.messages.length - 1].contentBlocks).toBe([{type: "text", text: "bar"}]);
+}
+```
+:::
+
+### Testing stateful logic
```
eyurtsev: `Testing stateful logic` => `perhaps testing with persistence` (or something like that to match langgraph terminology).

You could even call this section `InMemorySaver Checkpointer` or something along those lines. With this sort of organization we will be highlighting the relevant components as part of the information architecture.


I don't think that we need that much for this section except for telling users to use `InMemorySaver` as the checkpointer to enable persistence during testing.

The current example is confusing -- since it's not helping to test for a common use case, and is introducing an idea of dynamic models (which isn't that common of a pattern).

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+  }
+  
+  expect(result.messages[result.messages.length - 1].contentBlocks).toBe([{type: "text", text: "large response"}]);
+}
+```
+:::
+
+## Integration testing
+
+Many agent behaviors only emerge when using a real LLM, such as which tool the agent decides to call, how it formats responses, or whether a prompt modification affects the entire execution trajectory. LangChain offers an [`agentevals`](https://github.com/langchain-ai/agentevals) package that provides evaluators specifically designed for testing agent trajectories with live models.
```
eyurtsev: nit: 

LangChain offers an [`agentevals`]

=> 

LangChain's [`agentevals`] package

eyurtsev: It could be helpful to define an agent trajectory a bit more explicitly.

For example,

> An agent trajectory is a sequence of decisions that the agent takes (e.g., which tool to call, how to format a response) etc.

The content is already in the paragraph, I think a re-arrangement to make it more explicit what a trajectory is could be helpful

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+
+AgentEvals lets you easily evaluate the trajectory of your agent (i.e. the agent's message history) by performing a **trajectory match** or by using an **LLM judge**:
+
+<Card title="Trajectory match" icon="equals" arrow="true" href="#trajectory-match-evaluator">
+Ideal for testing well-defined workflows where you know the expected behavior. Use when you have specific expectations about which tools should be called and in what order. This approach is deterministic, fast, and cost-effective since it doesn't require additional LLM calls.
+</Card>
+<Card title="LLM-as-judge" icon="gavel" arrow="true" href="#llm-as-judge-evaluator">
+More flexible and can assess nuanced aspects like efficiency and appropriateness, but requires an LLM API call and is less deterministic. Use when you want to evaluate the overall quality and reasonableness of the agent's trajectory without strict tool call or ordering requirements.
```
eyurtsev: Many users will not know what llm-as-judge is, so useful for the first sentence to define what the technique is and only then explain when / why to use

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+:::
+
+## Integration testing
+
+Many agent behaviors only emerge when using a real LLM, such as which tool the agent decides to call, how it formats responses, or whether a prompt modification affects the entire execution trajectory. LangChain offers an [`agentevals`](https://github.com/langchain-ai/agentevals) package that provides evaluators specifically designed for testing agent trajectories with live models.
+
+AgentEvals lets you easily evaluate the trajectory of your agent (i.e. the agent's message history) by performing a **trajectory match** or by using an **LLM judge**:
+
+<Card title="Trajectory match" icon="equals" arrow="true" href="#trajectory-match-evaluator">
+Ideal for testing well-defined workflows where you know the expected behavior. Use when you have specific expectations about which tools should be called and in what order. This approach is deterministic, fast, and cost-effective since it doesn't require additional LLM calls.
```
eyurtsev: Would be very useful for the first sentence to explain what a trajectory match is / does / how it works (i.e., the developer hard-codes an expected sequence of decisions that the agent performs for a given input.)

And only then proceed with the rest of the content (i.e., when to use / what are the advantages etc.)

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+:::
+
+| Mode | Description | Use Case |
+|------|-------------|----------|
+| `strict` | Exact match of messages and tool calls in the same order | Testing specific sequences (e.g., policy lookup before authorization) |
+| `unordered` | Same tool calls allowed in any order | Verifying information retrieval when order doesn't matter |
+| `subset` | Agent calls only tools from reference (no extras) | Ensuring agent doesn't exceed expected scope |
+| `superset` | Agent calls at least the reference tools (extras allowed) | Verifying minimum required actions are taken |
+
+<Accordion title="Strict match">
```
eyurtsev: The accordions are nice here

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+  expect(evaluation.score).toBe(true);
+}
+```
+:::
+
+</Accordion>
+
+<Info>
+:::python
+You can also set the `tool_args_match_mode` property and/or `tool_args_match_overrides` to customize how the evaluator considers equality between tool calls in the actual trajectory vs. the reference. By default, only tool calls with the same arguments to the same tool are considered equal. Visit the [repo](https://github.com/langchain-ai/agentevals?tab=readme-ov-file#tool-args-match-modes) for more details.
```
eyurtsev: `repo` -> `repository`

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,1057 @@
...
+```
+
+Results will be automatically logged to LangSmith.
+
+<Tip>
+To learn more about evaluating your agent with LangSmith, see the [LangSmith docs](/langsmith/vitest-jest).
+</Tip>
+:::
+
+## Caching
```
eyurtsev: What about naming it more explicitly: "## Recording & Replaying HTTP Calls"

Caching feels ambiguous to me, especially since we have other forms of caching (e.g., caching for embeddings models)

eyurtsev (commented):
In the unit testing section, how about something like:

## Using a fake chat model

Overview of GenericFakeChatModel

Minimal example of using the fake chat model

Collapsable example of using it with create_agent.

## In memory persitence

And then highlighting InMemorySaver

src/oss/langchain-test.mdx
casparb: I agree that it is a bit of a non-example. A more useful example here would instead be mocking a tool call and doing something with the result, but even then, I am not sure how useful this would be at all - because at that point, why not test the tool directly? We use `FakeToolCallingModel` to test `create_agent` internally, but it's not part of any package.

I also want to avoid suggesting that users should need to test `create_agent` logic like making sure it provides the correct inputs to the chat model. What do you suggest?

src/oss/langchain-test.mdx
casparb: Unless we use `FakeToolCallingModel`, the agent's execution will immediately end after the `GenericFakeChatModel` response. I use a checkpointer here to sort of simulate a multi turn agent execution. Dynamic models are introduced in the new LangChain Agents doc, so I don't think its completely unusual to mention them here, but happy to remove this example if you think so.

src/oss/langchain-test.mdx
```diff
@@ -1,3 +1,946 @@
...
+  },
+  {
+    name: "get_detailed_forecast",
+    description: "Get detailed weather forecast for a city.",
+    schema: z.object({ city: z.string() }),
+  }
+);
+
+const agent = createAgent({
+  llm: "openai:gpt-4o",
```
jacoblee93: Should use `gpt-5-mini` or some variant

PR #388, opened 2025-09-05
docs(oss): bring consistency to tool names in langchain quickstart

markgoho:
the tool names in the system prompt don't currently match the names of the tools being provided to the agent

whether or not this has a material impact on the agent's ability to choose them, I'm not sure, but it certainly leads to confusion for new users who are left wondering why the tool names are different between system prompt and tool instantiation


lnhsingh:
Thank you!

src/oss/langchain-quickstart.mdx
```diff
@@ -159,7 +159,7 @@ Let's walk through each step:
                 return USER_LOCATION[user_id];
             },
             {
-                name: "get_user_info",
+                name: "get_user_location",
```
lnhsingh, line 162: This name should be updated in the description above too (replace `getUserInfo`)

markgoho: great callout -- more consistency 💪🏻

eyurtsev:
nice

PR #391, opened 2025-09-05
fix: unhide the agent auth page. add it under LGP access + control se…

samecrowder:
…ction

I think it's been around long enough now that we can/should do this

### DIFF
--- a/src/langgraph-platform/agent-auth.mdx
+++ b/src/langgraph-platform/agent-auth.mdx
@@ -1,7 +1,7 @@
 ---
-title: Agent Auth
+title: Setting up Agent Auth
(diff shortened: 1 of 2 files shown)


src/langgraph-platform/agent-auth.mdx
```diff
@@ -1,7 +1,7 @@
 ---
-title: Agent Auth
+title: Setting up Agent Auth
+description: Enable secure access from agents to any system using OAuth 2.0 credentials with Agent Auth
```
katmayb:
```suggestion
title: Set up Agent Auth
description: Enable secure access from agents to any system using OAuth 2.0 credentials with Agent Auth.
```

katmayb (approved):

PR #410, opened 2025-09-08
docs(oss): simplify chunk aggregation in messages doc

markgoho:
slightly opinionated here, but also fixing a broken reduce function

for...of is typically seen as a simpler way to loop over an array, but sticking with `concat()` since it has some additional behavior

```typescript
const chunks = [];

const stream = await model.stream('Write a poem');
for await (const chunk of stream) {
  chunks.push(chunk);
}

let finalChunk: AIMessageChunk | undefined;
for (const chunk of chunks) {
  finalChunk = finalChunk ? finalChunk.concat(chunk) : chunk;
}

console.log(`Full response: ${finalChunk?.text}`);

// output
Full response: I am a large language model,
Trained by Google, I can tell.
Stories, poems, code, and more,
All at your command, I'll explore.

What can I create for you today?
Just give me a prompt, and I'll obey.
From a haiku to an epic grand,
I'll do my best to understand.
```

### DIFF
--- a/src/oss/langchain/messages.mdx
+++ b/src/oss/langchain/messages.mdx
@@ -816,86 +816,116 @@ it as output. Below we show short examples of input messages featuring multimoda
 <CodeGroup>
     ```python Image input
     # From URL
-    response = model.invoke([
-        {"type": "text", "text": "Describe the content of this image."},
-        {"type": "image", "url": "https://example.com/path/to/image.jpg"},
-    ])
+    message = {
+        "role": "user",
+        "content": [
+            {"type": "text", "text": "Describe the content of this image."},
+            {"type": "image", "url": "https://example.com/path/to/image.jpg"},
+        ]
+    }
 
     # From base64 data
-    response = model.invoke([
-        {"type": "text", "text": "Describe the content of this image."},
-        {
-            "type": "image",
-            "base64": "AAAAIGZ0eXBtcDQyAAAAAGlzb21tcDQyAAACAGlzb2...",
-            "mime_type": "image/jpeg",
-        },
-    ])
+    message = {
+        "role": "user",
+        "content": [
+            {"type": "text", "text": "Describe the content of this image."},
+            {
+                "type": "image",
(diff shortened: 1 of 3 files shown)


src/oss/langchain/messages.mdx
```diff
@@ -458,10 +458,11 @@ During streaming, you'll receive `AIMessageChunk` objects that can be combined:
...
 
-    const finalChunk = chunks.reduce(
-        (acc, chunk) => acc.concat(chunk.text),
-        new AIMessageChunk("") // Start with an empty chunk
-    );
+    let fullText = '';
+    for (const chunk of chunks) {
+        fullText += chunk.text;
+    }
+    const finalChunk = new AIMessageChunk(fullText);
```
hntrl: I think we want to point people towards using the `.concat` function on AIMessageChunk directly since it does more than just aggregate text (if we're observing a stream directly, a chunk can also contain tool call chunks, content parts, etc.)

what about something like this instead?

```suggestion
    let finalChunk: AIMessageChunk;
    for (const chunk of chunks) {
        finalChunk = finalChunk ? finalChunk.concat(chunk) : chunk;
    }
```

hntrl (commented):
Thanks for the PR @markgoho! What's your thoughts on this?

src/oss/langchain/messages.mdx
markgoho: oof, I totally missed that this was a method on a class! partially because my linter (we use unicorn-eslint) flagged the rule violation that prefers the spread syntax over `concat`

all that is to say, for sure we should use the native method if it does extra stuff -- although _maybe_ there's an opportunity to rename it to `concatChunks` so as to avoid the potential misunderstanding that I made? not an easy decision, and not something we need to resolve here

I'll push an update with a slightly more typesafe version of what you've suggested

hntrl (approved):
love it! thanks for the fix!

beklapko (approved):

PR #411, opened 2025-09-08
fix mint broken links

lnhsingh:
Problem: Running mint broken-links from project root caused Mintlify to parse Python virtual environment files in .venv/, triggering MDX parsing errors (CI tests).

Fix: Updated Makefile so make mint-broken-links runs the command from the src/ directory (where docs.json lives) instead of project root, avoiding the .venv/ directory entirely.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: dev build format lint test install clean format_diff lint_diff unsafe_format lint_md lint_md_fix mint-broken-links mint-preview
+.PHONY: dev build format lint test install clean format_diff lint_diff unsafe_format lint_md lint_md_fix mint-broken-links
 
 dev:
 	@echo "Starting development mode..."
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -61,15 +61,26 @@ clean:
 	@find . -name "*.pyd" -delete
 	@find . -name "__pycache__" -type d -exec rm -rf {} +
 
+# Mintlify commands (run from src directory where docs.json is located)
```
eyurtsev: We don't want to check for broken links from src. Only from `build`

src isn't guaranteed to be mintlify code (it could be latex code that compiles to mintlify)

Makefile
```diff
@@ -61,15 +61,26 @@ clean:
...
+mint-broken-links:
+	@echo "Checking for broken links..."
+	@cd src && mint broken-links
+
+mint-preview:
+	@echo "Starting mint preview..."
+	@cd src && mint preview
```
eyurtsev: what is mint preview?

.github/workflows/test-build-docs.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           node-version: '18'
 
       - name: Install Mintlify CLI
-        run: npm i -g mint
+        run: npm i -g mint@4.1.0
```
eyurtsev, line 61: We probably want to stay on the most recent version?

lnhsingh: most recent version was throwing errors, hence this PR

PR #381, opened 2025-09-04
Brody/fix install blocks

beklapko:
- This fixes/adds missing install code tabs for some JS integration pages.
- Removes a callout from several docs (the callout linked to a page that's in the old docs but not the new docs, and didn't seem critical to me to have right now).

PR #415, opened 2025-09-08
Brody/port callout install tab changes

beklapko:
## Summary

There are three main changes in this PR but they're spread across a lot of docs.

1. This tooltip was removed (the callout linked to a page that's in the old docs but not the new docs, and didn't seem critical to me to have right now).

```
import IntegrationInstallTooltip from '/snippets/javascript-integrations/integration-install-tooltip.mdx';

<IntegrationInstallTooltip/>
```

2. Installation code blocks like this didn't migrate so well.

```
<Npm2Yarn>
  @arcjet/redact
</Npm2Yarn>
```

For now, I've replaced these with code groups (below is just an example, each code group is different based on the library getting installed)

````
<CodeGroup>
```bash npm
npm install @arcjet/redact
```
```bash yarn
yarn add @arcjet/redact
```
```bash pnpm
pnpm add @arcjet/redact
```
</CodeGroup>
````

There is probably a smarter/better long term way to do this with snippets but this was doable in the short term to fix the missing example/broken rendering.

3. Some [tooltips](https://github.com/langchain-ai/docs/pull/415/files#diff-cedeee983cdeb7a6d86f7e209be9d3719baa1107f1f7bd21c5756c903333f0daL369) and snippets got put in these `{=mdx}` blocks:

````
```{=mdx}
<Warning>
**Compatibility**

This feature is currently in beta.
</Warning>
```
````

So I removed the opening and closing "tags" which fixes the rendering issues.

## Unresolved issue

This snippet shows up in several places:

```
<ChatModelTabs customVarName="llm" />
```

It was enclosed with `{=mdx}`  and the triple backticks which caused it to be rendered like this in prod:

Screenshot 2025-09-08 at 1 32 06 PM

I removed the enclosing `{=mdx}` so it doesn't get rendered at all, but this doesn't fix the issue entirely. That snippet, from what I can tell, should render out something like this section in the old JS docs: https://js.langchain.com/docs/integrations/retrievers/exa/#use-within-a-chain.

It'd take more time and research to see if we can actually recreate that functionality here or not, which this PR does not address.

## Testing

I spot checked a few instances of each of these changes locally but I did not verify every change in every file.


beklapko:
Related but closed PR just for reference: https://github.com/langchain-ai/docs/pull/381.

I created this PR to fix additional stuff but also to start fresh on `main`.

lnhsingh:
Closing in favor of https://github.com/langchain-ai/docs/pull/1075

PR #418, opened 2025-09-08
multi agent

hwchase17:
Adding first version of multi agent document that describes tools and handoffs.

After this PR, we still need to:

1. Update the inputs (passing the full history needs to be fixed, we should probably adjust to just incorporating another key from the state)
2. Need to fill in the sections about handoffs

### DIFF
--- a/src/oss/langchain/multi-agent.mdx
+++ b/src/oss/langchain/multi-agent.mdx
@@ -2,10 +2,10 @@
 title: Multi-agent
 ---
 
+
 import AlphaCallout from '/snippets/alpha-lc-callout.mdx';
 
 <AlphaCallout />
-
 
 **Multi-agent systems** break a complex application into multiple specialized agents that work together to solve problems.  
 Instead of relying on a single agent to handle every step, **multi-agent architectures** allow you to compose smaller, focused agents into a coordinated workflow.
@@ -16,20 +16,12 @@ Multi-agent systems are useful when:
 * Context or memory grows too large for one agent to track effectively.
 * Tasks require **specialization** (e.g., a planner, researcher, math expert).
 
-Benefits include:
+## Multi-agent patterns
 
-| Benefit            | Description                                                       |
-|--------------------|-------------------------------------------------------------------|
-| **Modularity**     | Easier to develop, test, and maintain smaller, focused agents.    |
-| **Specialization** | Each agent can be optimized for a particular domain or task type. |
-| **Control**        | Explicitly define communication and control flow between agents.  |
-
-## Two common patterns
-
-| Pattern          | How it Works                                                                                                                                                     | Control Flow                                                | Example Use Case                                 |
-|------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------|--------------------------------------------------|
-| **Tool Calling** | A central agent calls other agents as *tools*. The “tool” agents don’t talk to the user directly — they just run their task and return results.                  | Centralized — all routing passes through the calling agent. | Task orchestration, structured workflows.        |
-| **Handoffs**     | The current agent decides to **transfer control** to another agent. The active agent changes, and the user may continue interacting directly with the new agent. | Decentralized — agents can change who is active.            | Multi-domain conversations, specialist takeover. |
+| Pattern          | How it works                                                                                                                                                     | Control flow                                               | Example use case                                 |
+|------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------|--------------------------------------------------|
+| **Tool Calling** | A central agent calls other agents as *tools*. The “tool” agents don’t talk to the user directly — they just run their task and return results.                  | Centralized: all routing passes through the calling agent. | Task orchestration, structured workflows.        |
+| **Handoffs**     | The current agent decides to **transfer control** to another agent. The active agent changes, and the user may continue interacting directly with the new agent. | Decentralized: agents can change who is active.            | Multi-domain conversations, specialist takeover. |
 
 ### Tool calling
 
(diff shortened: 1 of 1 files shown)


src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,224 @@
...
+````
+
+### Handoffs
+
+In **handoffs**, agents can directly pass control to each other. The “active” agent changes, and the user interacts with whichever agent currently has control.
+
+Flow:
+
+1. The **current agent** decides it needs help from another agent.
+2. It passes control (and state) to the **next agent**.
```
eyurtsev: "next" is probably the wrong adjective here

src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,224 @@
...
+:::python
+```python
+from langchain.tools import tool
+from langchain.agents import create_agent
+
+subagent1 = create_agent(..)
+
+@tool(
+    name="subagent1_name",
+    description="subagent1_description"
```
eyurtsev, line 90: note: we'll need to figure out separately (not related to multi-agent at all)
---

using `description` in the tool makes explanation easy, but provides no way to document `query`.

We'd need the `parse_docstring=True` for this

eyurtsev:
Looks pretty good. I can take over to refine visually, we'll need a TODO stub for handoffs

src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,308 @@
...
+    return result.messages.at(-1)?.text;
+  },
+  {
+    name: "subagent1_name",
+    description: "subagent1_description",
+    schema: z.object({
+      query: z.string().describe("The query to send to subagent1")
+    })
+  }
+);
```
hntrl:
```suggestion
const callSubagent1 = tool(
  async (query) => {
    const result = await subagent1.invoke({
      messages: [{ role: "user", content: input.query }]
    });
    return result.messages.at(-1)?.text;
  },
  {
    name: "subagent1_name",
    description: "subagent1_description",
    schema: z.string().describe("The query to send to subagent1")
  }
);
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,308 @@
...
+  },
+  {
+    name: "subagent1_name",
+    description: "subagent1_description", 
+    schema: z.object({
+      query: z.string().describe("The query to send to subagent1")
+    })
+  }
+);
+```
```
hntrl: Think this is duplicated from the snippet below. If I'm mirroring the same snippet above it should look something like:
```suggestion
```typescript
import { createAgent, tool, AgentState, ToolMessage } from "langchain";
import { Command } from "@langchain/langgraph";
import { z } from "zod";

// Example of passing the full conversation history to the sub agent via the state.
// When passing the full conversation, you often do NOT want to pass in a query.
const callSubagent1 = tool(
  async () => {
    const state = getCurrentTaskInput<AgentState>();
    const result = await subagent1.invoke({
      messages: state.messages,
    });
    return result.messages.at(-1)?.content;
  },
  {
    name: "subagent1_name",
    description: "subagent1_description",
  }
);
```
```

File: src/oss/langchain/multi-agent.mdx [hunk 5]
```diff
@@ -1,13 +1,308 @@
...
+    name="subagent1_name",
+    description="subagent1_description"
+)
+# Example of passing the full conversation history to the sub agent via the state.
+# When passing the full conversation, you often do NOT want to pass in a query.
+def call_subagent1(state: Annotated[CustomState, InjectedState]):
+    result = subagent1.invoke({
+        "messages": state["messages"]
+    })
+    return result["messages"][-1].text
```
hntrl: does this work? If we pass the whole message history in the invocation, the last message in the list will be a message with an unanswered tool call. I get an error like this if I'm using the snippet I recommended below in JS

```
Error: 400 An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_mPo9wbxDu6ioF5jE5OlIFVyl
```

We don't do any magic in python specifically when parsing messages for multi agent do we?

eyurtsev: I'll include in system message -- was trying to avoid heavy processing but that's where it typically needs to go

eyurtsev: I mean human message (or combination of system and human)

src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,292 @@
...
+```
+:::
+:::js
+```typescript
+import { tool, ToolMessage } from "langchain";
+import { Command } from "@langchain/langgraph";
+import { z } from "zod";
+
+const callSubagent1 = tool(
+  async (input: { query: string }, config) => {
```
hntrl:
```suggestion
  async (input, config) => {
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -1,13 +1,292 @@
...
+      messages: [{ role: "user", content: input.query }]
+    });
+    return result.messages.at(-1)?.text;
+  },
+  {
+    name: "subagent1_name",
+    description: "subagent1_description",
+    schema: z.string().describe("The query to send to subagent1")
+  }
+);
```
hntrl:
```suggestion
const callSubagent1 = tool(
  async (input) => {
    const result = await subagent1.invoke({
      messages: [{ role: "user", content: input.query }]
    });
    return result.messages.at(-1)?.text;
  },
  {
    name: "subagent1_name",
    description: "subagent1_description",
    schema: z.object({
      z.string().describe("The query to send to subagent1"),
    }),
  }
);
```

PR #481, opened 2025-09-10
fix(js/middleware): have users use re-exported Command from LangGraph

christian-bromann:
Small example fix to avoid type issues with incompatible LangGraph versions.

Blocked on release of https://github.com/langchain-ai/langchainjs/pull/8907

### DIFF
--- a/src/oss/langchain/middleware.mdx
+++ b/src/oss/langchain/middleware.mdx
@@ -271,9 +271,8 @@ agent = create_agent(
 
 :::js
 ```typescript
-import { createAgent } from "langchain";
+import { createAgent, Command } from "langchain";
 import { humanInTheLoopMiddleware } from "langchain/middleware";
-import { Command } from "@langchain/langgraph";
 
 const agent = createAgent({
     model: "openai:gpt-4o",


src/oss/langchain/middleware.mdx
```diff
@@ -271,9 +271,8 @@ agent = create_agent(
 
 :::js
 ```typescript
-import { createAgent } from "langchain";
+import { createAgent, Command } from "langchain";
```
eyurtsev, line 274: Are we re-exporting `Command` from langchain?

sydney-runkle: Hmm I don't think we should do this

eyurtsev: `Command` isn't used in this example

sydney-runkle: `Command` is very much a langgraph structure

sydney-runkle: At least for Python

christian-bromann: Agreed.

PR #575, opened 2025-09-18
fix code cell for streaming

ccurme:
Currently we split the streaming loop and combination into separate code blocks.

There are a few problems with the code block for combining chunks:

1. It doesn't work (we sum an AIMessage, which results in a ChatPromptTemplate):
```python
from langchain_core.messages import AIMessage, AIMessageChunk

chunks = [AIMessageChunk("foo"), AIMessageChunk(" bar")]

final_message = sum(chunks, AIMessage(""))  # ChatPromptTemplate
```

2. More nitpicky, but I think aggregating chunks is important and also concise enough that we can display it in the first cell with just two additional lines.

Here we demonstrate aggregation in the first cell (and also remove an import).

### DIFF
--- a/src/oss/langchain/messages.mdx
+++ b/src/oss/langchain/messages.mdx
@@ -422,25 +422,14 @@ for chunk in model.stream("Hi"):
 
 :::js
 <CodeGroup>
-    ```typescript Stream and collect chunks
-    const chunks = [];
-    const stream = await model.stream("Write a poem");
-    for await (const chunk of stream) {
-        chunks.push(chunk);
-        console.log(chunk.text);
-    }
-    ```
+```typescript
+import { AIMessageChunk } from "langchain";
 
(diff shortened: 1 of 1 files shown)


src/oss/langchain/messages.mdx
```diff
@@ -407,25 +407,17 @@ for (const toolCall of response.tool_calls) {
...
-
-    ```python Combine chunks
-    from langchain_core.messages import AIMessage
-
-    # Combine chunks into final message
-    final_message = sum(chunks, AIMessage(""))
-    print(f"Full response: {final_message.text}")
-    ```
-</CodeGroup>
+```python
```
lnhsingh, line 413: Should this also be updated on the JS side?

ccurme: @lnhsingh done (thanks @hntrl)

PR #577, opened 2025-09-18
Updated screenshots/gifs on analyze LS page

katmayb:
Removed some unnecessary screenshots and gifs. Updated text instructions on some of the pages.


katmayb:
@lnhsingh This is some easy win removals and doc updates. Eventually I'll need to come back to the Analyze page because it needs to have some updated screenshots / remove gifs, but it's a bit more involved.

src/langsmith/analyze-an-experiment.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: Analyze an experiment
+sidebarTitle: Analyze an experiment
```
lnhsingh, line 3: it's not a big deal, but we don't need the sidebarTitle if it's the same as the title

katmayb: Oh yeah, I feel like I added for it to be complete, which totally isn't necessary.

PR #586, opened 2025-09-18
add snippet js translation missing callout

eyurtsev:
Added a snippet for js translations. No need to merge until we need to use it. 

Idea is to make sure we have an easy way to identify which code blocks need to be translated!


src/snippets/js-snippet-missing.mdx
```diff
@@ -0,0 +1,4 @@
+<Callout icon="hand-holding-hand" color="#FFC107" iconType="regular">
+    We don’t have this code snippet in JavaScript yet. Want to help?
```
lnhsingh:
```suggestion
    We don’t have this code example in JavaScript yet. Want to help?
```

PR #608, opened 2025-09-19
Advertise LG dev through NPM package

christian-bromann:
As a JavaScript developer I may not have Python installed and prefer a snippet using NPM.

cc @dqbd  

fwiw: I don't think [these docs](https://docs.langchain.com/langgraph-platform/local-server#node-server) are helpful for JS folks to onboard, due to following reasons:

- `npx @langchain/langgraph-cli` doesn't install langgraph CLI (it just downloads the CLI and executes it)
- `langgraph new path/to/your/app --template new-langgraph-project-js` doesn't work because the JS version of LG CLI doesn't have a `new` command, it has `dev`, `dockerfile`, `build`, `up`, `sysinfo` and `help` but no `new`

### DIFF
--- a/src/langgraph-platform/quick-start-studio.mdx
+++ b/src/langgraph-platform/quick-start-studio.mdx
@@ -30,18 +30,18 @@ Next, install the [LangGraph CLI](/langgraph-platform/langgraph-cli):
 <CodeGroup>
 ```bash pip
 pip install -U "langgraph-cli[inmem]"
+langgraph dev
 ```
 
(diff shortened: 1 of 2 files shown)


dqbd (approved):

dqbd:
@christian-bromann for JS we had the `npm create langgraph` command, but might as well add it in CLI 
https://github.com/langchain-ai/langgraphjs/pull/1685

katmayb (approved):

PR #639, opened 2025-09-23
nits: lgp prose

mdrxy:
no offense taken if disagree with these

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -573,7 +573,7 @@
                     ]
                   },
                   {
-                    "group": "Configure your application for deployment",
+                    "group": "Configure app for deployment",
                     "pages": [
                       "langgraph-platform/setup-app-requirements-txt",
                       "langgraph-platform/setup-pyproject",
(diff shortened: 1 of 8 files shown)


src/docs.json
```diff
@@ -573,7 +573,7 @@
                     ]
```
katmayb, line 573: I'm ok with these changes — maybe change this group title to "Deployment guides"

src/langgraph-platform/deploy-hybrid.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: How to deploy hybrid
-sidebarTitle: Deploy hybrid
+sidebarTitle: Hybrid deployment
```
katmayb, line 3: If we're going with minimal in the sidebar, maybe Cloud and Hybrid can just drop "deployment" from their sidebarTitle

src/langgraph-platform/use-remote-graph.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: How to interact with a deployment using RemoteGraph
-sidebarTitle: Interact with a deployment using RemoteGraph
+sidebarTitle: Interact using RemoteGraph
```
katmayb: I actually think we should keep the previous version for this one. We had some comments that users and internal folks didn't feel it was immediately obvious what RemoteGraph was. So, "with a deployment" seems to provide a bit of extra context that may help.

src/docs.json
eyurtsev: I like the shorter variant if it helps :)

PR #640, opened 2025-09-24
feat: reference docs

hntrl:
Adds a new `reference/` path that centralizes the logic for creating reference documentation

## For Python

We can leverage the existing reference docs pipeline thats used within the `langchain` repo, and adapting it to include docs from more sources (see https://github.com/langchain-ai/langchain/pull/33066). More details about how this works are in `reference/python/README.md`.

## For TypeScript

We'll be keeping the existing [typedoc](https://typedoc.org/) library to render references, but we're onshoring the work of consolidating those into one reference site within this repo.

The basic process of how that happens is:
* by reading a statically defined list of packages and their locations (defined in reference/javascript/build.ts`)
* doing a shallow clone of each unique repo + branch
* doing the appropriate install steps so that types can be inferred properly
* extracting package entrypoints by reading the `exports` key in `package.json`
* pointing typedoc to add those as entrypoints with the [packages strategy](https://typedoc.org/documents/Options.Input.html#packages)
* building the typedoc outputs to reference/dist/javascript/

## New `make` targets

- `make build-references` will run scripts to format python + js reference documentation concurrently (which will end up in reference/dist/)
- `make preview-references` will run aforementioned build scripts and start a dev server (this won't work until a vercel project is configured)

## Vercel Setup

In order for these changes to be reflected on `reference.langchain.com`, a new Vercel project needs to be created (which can happen once this is merged).

This was tested against a "test" vercel setup and works as expected, with both reference targets accessible from their respective subpaths


reference/python/build.py
```diff
@@ -0,0 +1,104 @@
...
+VERSION_TAGS: list[str] = []
+
+def _extract_html_dir(tar: tarfile.TarFile, path: Path) -> None:
+    """Extract only `api_reference_build/html` members from the tar into `path`.
+
+    Guards against path traversal by verifying each member remains under `path`.
+    """
+    for member in tar.getmembers():
+        member_path = path / member.name
+        if not member_path.is_relative_to(path):
```
eyurtsev, line 32: Any attacks possible via symbolic links? I think the `is_relative_to` eliminates them effectively?

reference/python/__main__.py
```diff
@@ -0,0 +1,14 @@
+"""Main entry point for the Python reference docs build."""
```
eyurtsev, line 1: Lol definitely claude generated.

I've only ever seen claude use `__main__`

eyurtsev (approved):

reference/python/build.py
hntrl: yeah pretty sure. Am less concerned about this since this only becomes an issue if we get MIM'd (github gets hacked) or the ref docs build starts including symlinks (one of us merges a REALLY bad change)

reference/python/__main__.py
hntrl: if you give a hunter an infinite amount of tokens and time he will eventually rewrite all of python






..
https://en.wikipedia.org/wiki/Infinite_monkey_theorem

PR #641, opened 2025-09-24
fix: update import path for message classes

leonyangdev:


hntrl:
Hey @hefeng6500!

This is intended actually -- we're thinking it's better for the end user if you're importing everything you need from one package. The only time you would need to import these from core is when you're writing integrations to avoid the dependency bloat.

In light of that, going to close this.

leonyangdev:
> Hey @hefeng6500!
> 
> This is intended actually -- we're thinking it's better for the end user if you're importing everything you need from one package. The only time you would need to import these from core is when you're writing integrations to avoid the dependency bloat.
> 
> In light of that, going to close this.

Hi @hntrl, thanks for the review and for keeping imports minimal for end users.
However, when I run this code on langchain v1.0.0a9, I get the following error.

```python
from langchain.messages import HumanMessage, AIMessage, SystemMessage
```


I get the following error:

```python
ModuleNotFoundError: No module named 'langchain.messages'
```


It only works when I use:

```python
from langchain_core.messages import HumanMessage, AIMessage, SystemMessage
```


Could you confirm the correct import path for the docs examples? Otherwise, users following the documentation will run into the same error.

hntrl:
apologies, this was a miscomm on my part! We're currently working on adding these to `langchain`

cc @sydney-runkle @eyurtsev

mcavdar:
Hi @hntrl , 
I was about to open a PR for the same issue but noticed this one already exists.
Are contributions still welcome? Would it be possible to consider reopening this PR and merging it?

hntrl:
@mcavdar @hefeng6500 We can reopen this and merge, then when `langchain.messages` becomes available we'll swap it over. Thanks!

hntrl (approved):

PR #644, opened 2025-09-24
feat(langchain): add docs on dynamic models

christian-bromann:
This patch adds a docs page on "Dynamic Models" explaining concepts how users can dynamically switch models within `create_agent` based on context (simple) or complexity derived from state (advanced).

Happy to enhance with Python docs once we are ok with the structure and examples.

### DIFF
--- a/src/oss/langchain/dynamic-models.mdx
+++ b/src/oss/langchain/dynamic-models.mdx
@@ -1,0 +1,132 @@
+---
+title: Dynamic Models
+---
+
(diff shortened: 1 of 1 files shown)


eyurtsev:
Let's keep in draft mode. We don't want this to be a standalone page, and we want to remove most of the text I think in this case. We show how to change the model, without recommendations about the situations in which to do it. (This is one of those tasks that will be obvious to a user if they need to do it.)

PR #646, opened 2025-09-24
fix: reverse proxy reference docs

hntrl:
Instead of sourcing from a `dist` folder, `rewrites` will modify web assets to point to the appropriate subpath. Fixes web assets not loading on reference docs


lnhsingh (approved):
Approving but might want an eng review - up to you!

PR #652, opened 2025-09-25
docs: add some emphasis for setting up blob and restrict role

langchain-infra:
Restrict role and emphasize blob setup.

### DIFF
--- a/src/langsmith/kubernetes.mdx
+++ b/src/langsmith/kubernetes.mdx
@@ -234,5 +234,6 @@ As a next step, it is strongly recommended you work with your infrastructure adm
 * Configure SSL to ensure in-transit encryption of traces submitted to LangSmith
 * Configure LangSmith with [Single Sign-On](/langsmith/self-host-sso) to secure your LangSmith instance
 * Connect LangSmith to external Postgres and Redis instances
+* Setup [Blob Storage](/langsmith/self-host-blob-storage) for storing large files
 
 Review our [configuration section](/langsmith/architectural-overview) for more information on how to configure these options.
--- a/src/langsmith/kubernetes.mdx
+++ b/src/langsmith/kubernetes.mdx
@@ -234,6 +234,6 @@ As a next step, it is strongly recommended you work with your infrastructure adm
 * Configure SSL to ensure in-transit encryption of traces submitted to LangSmith
 * Configure LangSmith with [Single Sign-On](/langsmith/self-host-sso) to secure your LangSmith instance
 * Connect LangSmith to external Postgres and Redis instances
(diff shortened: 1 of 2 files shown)


src/langsmith/self-host-blob-storage.mdx
```diff
@@ -44,7 +44,12 @@ To authenticate to [Amazon S3](https://aws.amazon.com/s3/), you will need to cre
   "Statement": [
     {
       "Effect": "Allow",
-      "Action": "s3:*",
+      "Action": [
```
katmayb, line 47: In the description before this JSON, we say "you will need to create an IAM policy granting admin permissions on your bucket"... should we remove `admin`?

... you will need to create an IAM policy granting the following permissions on your bucket."

src/langsmith/self-host-blob-storage.mdx
```diff
@@ -3,7 +3,7 @@ title: Enable blob storage
 sidebarTitle: Enable blob storage
 ---
 
-By default, LangSmith stores run inputs, outputs, errors, manifests, extras, and events in ClickHouse. If you so choose, you can instead store this information in blob storage, which has a couple of notable benefits:
+By default, LangSmith stores run inputs, outputs, errors, manifests, extras, and events in ClickHouse. If you so choose, you can instead store this information in blob storage, which has a couple of notable benefits. We **highly recommend** using blob storage for production deployments due to the benefits outlined below:
```
katmayb:
```suggestion
By default, LangSmith stores run inputs, outputs, errors, manifests, extras, and events in ClickHouse. If you so choose, you can instead store this information in blob storage, which has a couple of notable benefits. For the best results in production deployments, we **strongly** recommend using blob storage, which offers the following benefits:
```
What do you think?

src/langsmith/kubernetes.mdx
```diff
@@ -234,5 +234,6 @@ As a next step, it is strongly recommended you work with your infrastructure adm
 * Configure SSL to ensure in-transit encryption of traces submitted to LangSmith
 * Configure LangSmith with [Single Sign-On](/langsmith/self-host-sso) to secure your LangSmith instance
 * Connect LangSmith to external Postgres and Redis instances
+* Setup [Blob Storage](/langsmith/self-host-blob-storage) for storing large files
```
katmayb:
```suggestion
* Set up [Blob Storage](/langsmith/self-host-blob-storage) for storing large files
```

katmayb (approved):
LGTM, couple of non-blocking suggestions

src/langsmith/self-host-blob-storage.mdx
langchain-infra: updated thx

PR #733, opened 2025-10-02
fix(docs): fix tools import

casparb:
Fix `InjectedState` import (moved from langchain.agents.tool_node to langchain.tools)


lnhsingh (approved):
ty!

PR #738, opened 2025-10-03
chore(lgp): document middleware_order setting

pjrule:
## Overview
This documents the `middleware_order` feature shipped with `langgraph-api` v0.4.35.

## Type of change

### DIFF
--- a/src/langgraph-platform/custom-middleware.mdx
+++ b/src/langgraph-platform/custom-middleware.mdx
@@ -58,6 +58,26 @@ Add the following to your `langgraph.json` configuration file. Make sure the pat
 }
 ```
 
+### Customize middleware ordering
+By default, custom middleware runs before authentication and authorization logic. To run custom middleware _after_ authentication and authorization, set `middleware_order` to `auth_first` in your `http` configuration.
+
+```json
+{
+  "dependencies": ["."],
+  "graphs": {
+    "agent": "./src/agent/graph.py:graph"
+  },
+  "env": ".env",
+  "http": {
+    "app": "./src/agent/webapp.py:app",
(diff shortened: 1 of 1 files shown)


src/langgraph-platform/custom-middleware.mdx
```diff
@@ -58,6 +58,26 @@ Add the following to your `langgraph.json` configuration file. Make sure the pat
 }
 ```
 
+### Customize middleware ordering
+By default, custom middleware runs before authentication and authorization logic. To run custom middleware _after_ authentication and authorization, set `middleware_order` to `auth_first` in your `http` configuration.
```
hinthornw: Maybe say min server version for this?

pjrule: Was thinking about that, though I'm not aware of a uniform convention in the docs for doing this (it would be really great if we had a directive like `@since` in Javadoc).

@lnhsingh / @katmayb – do you have a preferred wording here? could just add a parenthetical, something like
> (This customization is supported beginning in version 0.4.35 of the API server.)

katmayb: Maybe:

> This customization is supported starting with API server v0.4.35 and later.

hinthornw: ah also 'authorization' still runs after middleware sorry. Let's just keep "authentication"

eyurtsev:
```suggestion

By default, custom middleware runs before authentication and authorization logic. To run custom middleware _after_ authentication and authorization, set `middleware_order` to `auth_first` in your `http` configuration.
```

PR #756, opened 2025-10-06
feat: update `package.yml` download counts on workflow cron

mdrxy:


.github/workflows/update-package-downloads.yml
```diff
@@ -0,0 +1,48 @@
...
+    - cron: "59 23 * * 0"
+  workflow_dispatch:  # Allow manual triggering
+
+jobs:
+  update-downloads:
+    name: Update package download counts
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: write  # Required to push changes
```
eyurtsev: The safe way to do this is the same as the release workflow.

Generate an artifact with the new yaml file. Then load the artifact from another step that has write permissions

PR #792, opened 2025-10-08
feat(docs): add thinking in langgraph

eyurtsev:
Adding thinking in langgraph

### DIFF
--- a/src/oss/langgraph/thinking-in-langgraph.mdx
+++ b/src/oss/langgraph/thinking-in-langgraph.mdx
@@ -36,18 +36,18 @@ Start by identifying the distinct steps in your process. Each step will become a
 flowchart TD
     A[START] --> B[Read Email]
     B --> C[Classify Intent]
-    
+
     C -.-> D[Doc Search]
     C -.-> E[Bug Track]
     C -.-> F[Human Review]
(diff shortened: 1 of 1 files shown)


src/docs.json
```diff
@@ -195,6 +195,7 @@
                       "oss/python/langgraph/install",
                       "oss/python/langgraph/quickstart",
                       "oss/python/langgraph/local-server",
+                      "oss/langgraph/thinking-in-langgraph",
```
eyurtsev:
```suggestion
                      "oss/python/langgraph/thinking-in-langgraph",
```

src/oss/langgraph/thinking-in-langgraph.mdx
```diff
@@ -0,0 +1,568 @@
+---
+title: Thinking in LangGraph
```
lnhsingh, line 2:
```suggestion
title: LangGraph design patterns 
```

What about something like this?

eyurtsev: I like thinking in langgraph more (I think it's clearer that it's more of a conceptual walk through than design patterns). We can poll more folks?

PR #843, opened 2025-10-11
Update agents.mdx

gsmini:
I found that the `wrap_model_call` function no longer exists, so i use func `modify_model_request ` and the parameters passed to the target function that calls the `modify_model_request` are different from `wrap_model_call `. Therefore, I made modifications to this part of the example based on the comments in the source code

### DIFF
--- a/src/oss/langchain/agents.mdx
+++ b/src/oss/langchain/agents.mdx
@@ -144,33 +144,36 @@ Dynamic models are selected at <Tooltip tip="The execution environment of your a
 
 :::python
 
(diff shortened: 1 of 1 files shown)


sydney-runkle:
Ah I think you're on an old alpha version, the newest uses wrap!

gsmini:
> Ah I think you're on an old alpha version, the newest uses wrap!

Yes, after I re-executed `pip install -- pre-U langchain`, the code took effect and `wrap_model_call` could be imported normally. However, at the same time, the function `ModelResponse` does not exist. Please kindly check it

PR #863, opened 2025-10-13
ref envrionment switcher

mdrxy:


reference/python/README.md
```diff
@@ -51,36 +51,108 @@ For packages that live in the `langchain-ai/langchain` monorepo, the path to the
...
+
+### Development Workflow
+
+For local development with live source code:
+
+```bash
+# 1. Ensure repos are cloned in the expected structure (see below)
+
+# 2. Switch to development mode and install
+make dev-install
```
eyurtsev, line 69: IMO OK for now, but we'll probably want to do this differently long-term; i.e., we'd want to kill all the pyproject.toml files

We don't want the references to depend on "installing" the packages -- instead ideally we make their source code available, but not install them or their sub-dependencies. This is possible since griffe is based on ast scanning

src/oss/concepts/context.mdx
eyurtsev, line 40: Perhaps break this into a separate PR?

mdrxy: these got commits absorbed due to branch mismatch I think

PR #865, opened 2025-10-13
fix ref: make path case insensitive

mdrxy:
URLs should be case insensitive unless there is a good reason otherwise. Currently, I don't see any situations where we rely on it (or should be).

I'm not married to this; if adding middleware introduces unnecessary excess overhead, we can close this unmerged.

This is currently only scoped to the reference docs; unclear whether supported in our Mintlify deployment.


eyurtsev:
Why is this an issue? (i.e., why is it important if the URL is case sensitive?)

mdrxy:
@eyurtsev if a user types in `reference.langchain.com/Python` or it autocorrects to the proper noun (which happened to me on mobile and inspired this), it returns 404.

Not a super big issue

lnhsingh (approved):

PR #886, opened 2025-10-14
fix: homepage sidebar, dropdowns, typescript rename

hntrl:
* hides the homepage sidebar using the custom page mode
* makes the version picker a dropdown instead of a locale switcher
  * from what I gather this is more performant
  * should reduce the amount of info in the top bar
* renames "javascript" -> "typescript" in the language dropdown
  * group consensus was this is better positioning with JS OSS devs


lnhsingh:
does the icon in the language dropdown need to be TS instead of JS or does it not matter?

PR #905, opened 2025-10-15
Update short-term-memory.mdx

gsmini:
package need to be installed:
 ```
pip install langgraph-checkpoint-postgres
```
and `checkpointer.setup()` should be called( auto create tables in postgresql )

### DIFF
--- a/src/oss/langchain/short-term-memory.mdx
+++ b/src/oss/langchain/short-term-memory.mdx
@@ -77,11 +77,12 @@ await agent.invoke(
 
 In production, use a checkpointer backed by a database:
 
-```python
+
+:::python
+```shell
 pip install langgraph-checkpoint-postgres
 ```
 
-:::python
 ```python
 from langchain.agents import create_agent
 

### CODE REVIEW

File: src/oss/langchain/short-term-memory.mdx [hunk 1]
```diff
@@ -77,13 +77,19 @@ await agent.invoke(
 
 In production, use a checkpointer backed by a database:
 
+```python
+    pip install langgraph-checkpoint-postgres
```
eyurtsev:
```suggestion
pip install langgraph-checkpoint-postgres
```

src/oss/langchain/short-term-memory.mdx
```diff
@@ -77,13 +77,19 @@ await agent.invoke(
 
 In production, use a checkpointer backed by a database:
 
+```python
```
eyurtsev: This needs to be inside the `:::python` block. Would you mind moving it there?

PR #940, opened 2025-10-16
fix: `javascript` -> `typescript`

mdrxy:
for consistency w/ new lang dropdown

also add feedback for tabs

new icon for advanced agent


src/index.mdx
```diff
@@ -54,18 +54,18 @@ mode: "custom"
                     </Card>
                 </CardGroup>
             </Tab>
-            <Tab title="JavaScript" icon="js">
+            <Tab title="TypeScript" icon="js">
```
eyurtsev, line 57: javascript might be better here. I think developers will know that it supports typescript as well, but execs might not know the difference between javascript or typescript (and may be more familiar with javascript)

PR #947, opened 2025-10-16
chore(langchain): add custom state schema back

sydney-runkle:


src/oss/python/migrate/langchain-v1.mdx
```diff
@@ -74,7 +74,7 @@ The table below outlines what functionality has changed from `create_react_agent
 | [Prompts](#prompts) | Parameter renamed to `system_prompt`, dynamic prompts use middleware |
 | [Pre-model hook](#pre-model-hook) | Replaced by middleware with `before_model` method |
 | [Post-model hook](#post-model-hook) | Replaced by middleware with `after_model` method |
-| [Custom state](#custom-state) | Defined in middleware, `TypedDict` only |
+| [Custom state](#custom-state) | Parameter name unchanged (`state_schema`), `TypedDict` only |
```
sydney-runkle:
```suggestion
| [Custom state](#custom-state) | `TypedDict` only, can be defined on middleware |
```

src/oss/langchain/agents.mdx
```diff
@@ -669,26 +669,21 @@ Information stored in the state can be thought of as the [short-term memory](/os
 
 :::python
 
-Custom state schemas must extend @[`AgentState`] as a `TypedDict`. Define them in middleware using the `state_schema` attribute:
+Custom state schemas must extend @[`AgentState`] as a `TypedDict`. Pass them to @[`create_agent`] using the `state_schema` parameter:
```
eyurtsev: Could we add a note here after the code block saying that this is shorthand for convenience. If users need tools that require state, we recommend shipping everything as part of middleware.

PR #966, opened 2025-10-17
clarify behavior around prereleases in langsmith deployments

isahers1:


phvash (approved):

katmayb:
@isahers1 I edited your content slightly and added it to a snippet so it could be included on multiple pages. There may be more places that this tip could be useful, but for now just added to the pyproject and requirements.txt pages.

PR #981, opened 2025-10-17
Translate supervisor tutorial to js

pokey:
Just translates the Python version of the supervisor tutorial to js. 

- Depends on https://github.com/langchain-ai/langchainjs/pull/9228
- See also https://github.com/langchain-ai/langchainjs/pull/9235


src/oss/langchain/supervisor.mdx
```diff
@@ -790,9 +1465,47 @@ def schedule_event(request: str) -> str:
...
+      request: z.string().describe("Natural language scheduling request"),
+    }),
+  }
+);
+```
+:::
 
 **Important:** Make sure sub-agent prompts emphasize that their final message should contain all relevant information. A common failure mode is sub-agents that perform tool calls but don't include the results in their final response.
 
+:::js
```
pokey, line 1492: I don't think we have a Python example so I've made this callout js-only for now

src/docs.json
```diff
@@ -308,7 +308,7 @@
                           "oss/python/langchain/knowledge-base",
                           "oss/python/langchain/rag",
                           "oss/python/langchain/sql-agent",
-                          "oss/python/langchain/supervisor"
+                          "oss/langchain/supervisor"
```
hntrl: it's not very intuitive how these doc files get named, but having the `oss/python` prefix is actually correct. Mintlify doesn't handle the code fencing syntax themselves, so what happens is when we detect that there is code fences in a doc page, two files get outputted in a pre process step (one under `oss/python` and one under `oss/javascript`). We don't apply that logic to `docs.json` since there's some edge cases we want to be explicit about.

src/oss/langchain/supervisor.mdx
```diff
@@ -52,11 +54,31 @@ conda install langchain -c conda-forge
...
+```
+```bash yarn
+yarn add langchain
+```
+```bash pnpm
+pnpm add langchain
+```
+</CodeGroup>
+
+For more details, see our [Installation guide](/oss/javascript/langchain/install).
```
hntrl: similar story to `docs.json`. This link should be `/oss/langchain/install`

src/oss/langchain/supervisor.mdx
```diff
@@ -160,6 +286,27 @@ for step in calendar_agent.stream(
...
+
+for await (const step of stream) {
+    for (const update of Object.values(step)) {
+        if (update && typeof update === "object" && "messages" in update) {
+            for (const message of update.messages) {
+                console.log(message.prettyPrint());
+            }
+        }
+    }
+}
```
hntrl, line 294: weird formatting -- should use two spaces

src/oss/langchain/supervisor.mdx
```diff
@@ -225,6 +373,45 @@ for step in email_agent.stream(
...
+});
+
+for await (const step of stream) {
+  for (const update of Object.values(step)) {
+    if (update && typeof update === "object" && "messages" in update) {
+      for (const message of update.messages) {
+        console.log(message.prettyPrint());
+      }
+    }
+  }
```
hntrl, line 398: this looked odd until I realized that stream steps are just cast as `any` types 🫠

src/oss/langchain/supervisor.mdx
```diff
@@ -704,6 +1300,42 @@ for step in supervisor_agent.stream(
...
 ```
+:::
+
+:::js
+```typescript
+import { Command } from "@langchain/langgraph"; // [!code highlight]
+
+const resume: Record<string, any> = {};
+for (const interrupt of interrupts) {
+  if (interrupt.id === "2b56f299be313ad8bc689eff02973f16") {
```
hntrl: hmm don't think this is desired?

src/oss/langchain/supervisor.mdx
```diff
@@ -759,6 +1392,47 @@ def schedule_event(
...
+:::
+
+:::js
+```typescript
+const scheduleEvent = tool(
+  async ({ request }, config) => {
+    // Customize context received by sub-agent
+    // Access full thread messages from the config
+    const currentMessages =
+      config?.configurable?.pregel_scratchpad?.currentTaskInput?.messages || [];
```
hntrl: The preferred way to introspect state (currently):
```suggestion
    // Customize context received by sub-agent
    // Access full thread messages from the config
    const currentMessages = getCurrentTaskInput<InternalAgentState>(config).messages;
```
importing InternalAgentState from langchain and `getCurrentTaskInput` from langgraph

src/oss/langchain/supervisor.mdx
```diff
@@ -759,6 +1392,47 @@ def schedule_event(
...
+
+    const originalUserMessage = currentMessages.find(
+      (message: any) => message._getType() === "human"
+    );
```
hntrl:
```suggestion
    const originalUserMessage = currentMessages.find(HumanMessage.isInstance);
```

src/oss/langchain/supervisor.mdx
pokey: yeah this was the cleanest I could come up with, but does actually result in a type error. Should I add an explicit `any` cast? Would be good to revisit at some point to see if we can come up with something clean that keeps strong types

src/oss/langchain/supervisor.mdx
pokey: nice catch. I've fixed the ts but left the python

src/oss/langchain/supervisor.mdx
pokey: I made this fix, but it from what I can tell, `InternalAgentState` is not actually exported from `langchain`? Did we want to export that?

pokey:
Ok I addressed your comments and left responses to a couple

PR #1023, opened 2025-10-20
Copy trajectory oss content to LS docs

katmayb:
Fixes DOC-379

## Preview

https://langchain-5e9cc07a-preview-agente-1760972097-cd727ef.mintlify.app/langsmith/trajectory-evals

### DIFF
--- a/src/langsmith/trajectory-evals.mdx
+++ b/src/langsmith/trajectory-evals.mdx
@@ -1,6 +1,6 @@
 ---
 title: How to evaluate your agent with trajectory evaluations
-sidebarTitle: Trajectory evaluation
+sidebarTitle: Evaluate agent trajectories
 ---
 
 Many agent behaviors only emerge when using a real LLM, such as which tool the agent decides to call, how it formats responses, or whether a prompt modification affects the entire execution trajectory. LangChain's [`agentevals`](https://github.com/langchain-ai/agentevals) package provides evaluators specifically designed for testing agent trajectories with live models.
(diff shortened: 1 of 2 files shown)


src/langsmith/trajectory-evals.mdx
```diff
@@ -0,0 +1,589 @@
...
+  // }
+  expect(evaluation.score).toBe(true);
+}
+```
+
+</CodeGroup>
+
+### Unordered match
+
+The `unordered` mode allows the same tool calls in any order, which is helpful when you want to verify that specific information was retrieved but don't care about the sequence. For example, an agent might need to check both weather and events for a city, but the order doesn't matter.
```
catherine-langchain: instead of "that specific information", maybe can say "the correct set of tools are being invoked but don't care about the sequence"

src/langsmith/trajectory-evals.mdx
```diff
@@ -0,0 +1,589 @@
...
+  // }
+  expect(evaluation.score).toBe(true);
+}
+```
+
+</CodeGroup>
+
+### Subset and superset match
+
+The `superset` and `subset` modes match partial trajectories. The `superset` mode verifies that the agent called at least the tools in the reference trajectory, allowing additional tool calls. The `subset` mode ensures the agent did not call any tools beyond those in the reference.
```
catherine-langchain: I think we could be a bit more prescriptive about why / the scenarios where they might be used here! 
For the subset mode - it could be helpful to checking a few key tools are called in the execution, but you are okay with the agent calling additional tools.
For the superset mode - this is helpful to ensure agent efficiency, where they did not call any irrelevant tools that could be unnecessary

katmayb: Done!

src/langsmith/trajectory-evals.mdx
katmayb: Done!

katmayb:
TFTR @catherine-langchain ! I've implemented your feedback, PTAL!

PR #1108, opened 2025-10-23
Update config options for ls deployment

katmayb:


src/langsmith/deploy-self-hosted-full-platform.mdx
```diff
@@ -46,11 +46,16 @@ This guide builds on top of the [Kubernetes installation guide](/langsmith/kuber
 
 ## Setup
 
-1. As part of configuring your self-hosted LangSmith instance, you enable the `langgraphPlatform` option. This will provision a few key resources.
+1. As part of configuring your self-hosted LangSmith instance, you enable the `deployment` option. This will provision a few key resources.
     1. `listener`: This is a service that listens to the [control plane](/langsmith/control-plane) for changes to your deployments and creates/updates downstream CRDs.
     2. `LangGraphPlatform CRD`: A CRD for LangSmith Deployment. This contains the spec for managing an instance of a LangSmith deployment.
```
katmayb, line 51: @langchain-ai/infra what should the CRD be updated to?

langchain-infra: that one we cant actually update super easily as it will break existing deployments. I think we just keep this for now.

katmayb: Got it! Thanks.

PR #1168, opened 2025-10-28
fix references to chat completions

ccurme:
1. We refer to Chat Completions as a legacy API— as far as I know this is not the case (let me know if OpenAI documents that somewhere)
2. We link to the legacy Completions API, which is not the same as Chat Completions


src/oss/python/integrations/chat/openai.mdx
```diff
@@ -15,7 +15,7 @@ You can find information about OpenAI's latest models, their costs, context wind
 <Note>
     **Chat Completions API compatiblity**
```
mdrxy:
```suggestion
    **Chat Completions API compatibility**
```

PR #1266, opened 2025-11-04
Fix environment variable name for Google Scholar API

rajputshivam2324:
Update API key environment variable reference , 

In credentials  , environmental variable is defined like this :
process.env.GOOGLE_SCHOLAR_API_KEY="your-serp-api-key"
 to  : SERPAPI_API_KEY="your-serp-api-key"

### DIFF
--- a/src/oss/javascript/integrations/tools/google_scholar.mdx
+++ b/src/oss/javascript/integrations/tools/google_scholar.mdx
@@ -31,7 +31,7 @@ npm install @langchain/community
 Ensure you have the appropriate API key to access Google Scholar. Set it in your environment variables:
 
 ```typescript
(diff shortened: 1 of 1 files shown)


mdrxy:
Can you provide attribution?

rajputshivam2324:
In the credentials section, it mentioned using the below code snippets in the .env file.
However, in the .env file, we don’t write process.env.API_KEY_NAME; we only define the key name and its value.

Also, since there’s no official Google Scholar API key, the SERPAPI_API_KEY is actually used for this purpose and the same is referenced later in the docs.
This change helps avoid confusion for beginners who might misunderstand the snippet.

mdrxy (approved):

PR #1281, opened 2025-11-04
fix run delete docs

angus-langchain:
## Overview
Updates run delete docs and fixes some api request issues

### DIFF
--- a/src/langsmith/data-purging-compliance.mdx
+++ b/src/langsmith/data-purging-compliance.mdx
@@ -26,7 +26,7 @@ All trace deletions will delete related entities like feedbacks, aggregations, a
 
 ### Deletion timeline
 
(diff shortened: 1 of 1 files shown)


src/langsmith/data-purging-compliance.mdx
```diff
@@ -26,18 +26,18 @@ All trace deletions will delete related entities like feedbacks, aggregations, a
 
 ### Deletion timeline
 
-Trace deletions are processed during non-peak usage times and are not instant, usually within a few hours. There is no confirmation of deletion - you'll need to query the data again to verify it has been removed.
+Trace deletions are processed during non-peak usage times and are not instant. We run the delete job on the weekend. There is no confirmation of deletion - you'll need to query the data again to verify it has been removed.
```
katmayb:
```suggestion
Trace deletions are processed during non-peak usage times and are not instant. LangChain runs the delete job on the weekend. There is no confirmation of deletion - you'll need to query the data again to verify it has been removed.
```

PR #1364, opened 2025-11-10
Add feedback to initial example for run evals api only

katmayb:
Fixes DOC-374

Adds another stage to the first example to add feedback using the API

## Preview

https://langchain-5e9cc07a-preview-feedba-1762805584-28a22f6.mintlify.app/langsmith/run-evals-api-only


katmayb:
TFTR, and yes tested!

PR #1368, opened 2025-11-10
fix(langchain): update docs on summarization middleware

christian-bromann:
After recent changes made in:

- Python: https://github.com/langchain-ai/langchainjs/pull/9373
- JS: https://github.com/langchain-ai/langchainjs/pull/9373


sydney-runkle:
I don't think these are the right recommendations

PR #1371, opened 2025-11-10
feat(langchain): document interrupts in middleware

christian-bromann:
We received the following feedback from @casparb 

> There should be mention in the middleware docs of which middleware you can and cannot raise interrupts in. E.g. I've seen a few cases where people reading the docs think it makes sense to raise an interrupt in @wrap_tool_call  to validate a tool message before it's sent to the model. But you can't raise an interrupt here, because on resume you'd re-execute ToolNode.
Maybe more generally HITL middleware should be expanded to have an interrupt_after param. I don't really know what the alternative to this looks like? look at the last message in the messages list and inspect the tool call name in a before_model middleware?

This patch adds a section speaking about how to use interrupts in middleware.

### DIFF
--- a/src/oss/langchain/middleware.mdx
+++ b/src/oss/langchain/middleware.mdx
@@ -299,6 +299,8 @@ const agent = createAgent({
 <Note>
     **Important:** Human-in-the-loop middleware requires a [checkpointer](/oss/langgraph/persistence#checkpoints) to maintain state across interruptions.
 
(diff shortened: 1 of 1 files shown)


sydney-runkle:
I don't think we should merge this, not fully correct as per our slack disc today

lnhsingh:
closing due to other middleware updates

PR #1393, opened 2025-11-13
docs: add guidance on langsmith deployment scaling

jdrogers940:
## Type of change

**Type:** [Replace with: New documentation page / Update existing documentation / Fix typo/bug/link/formatting / Remove outdated content / Other]

## Related issues/PRs

- GitHub issue:
- Feature PR:

- Linear issue:
- Slack thread:

## Checklist

- I have gotten approval from the relevant reviewers
- (Internal team members only / optional) I have created a preview deployment using the [Create Preview Branch workflow](https://github.com/langchain-ai/docs/actions/workflows/create-preview-branch.yml)

### DIFF
--- a/src/langsmith/agent-server-scale.mdx
+++ b/src/langsmith/agent-server-scale.mdx
@@ -160,7 +160,7 @@ These settings are only required for [self-hosted](/langsmith/self-hosted) deplo
 
 ##### Configure autoscaling for bursty workloads
 
-Autoscaling is disabled by default, but should be configured for bursty workloads. You can determine the maximum number of api servers you should allow the autoscaler to scale to based on maximum expected throughput.
+Autoscaling is disabled by default, but should be configured for bursty workloads. You can determine the maximum number of api servers you should allow the autoscaler to scale to based on maximum expected throughput. The default for [cloud](/langsmith/cloud) deployments is a maximum of 10 API servers.
 
 ## Example self-hosted Agent Server configurations
 
@@ -347,11 +347,6 @@ api:
     limits:
       cpu: "2"
       memory: "4Gi"
-  autoscaling:
-    enabled: true
-    minReplicas: 15
-    maxReplicas: 25
-    targetCPUUtilizationPercentage: 70
 
 queue:
   replicas: 10
@@ -362,11 +357,6 @@ queue:
     limits:
       cpu: "2"
       memory: "4Gi"
-  autoscaling:
-    enabled: true
-    minReplicas: 10
-    maxReplicas: 20
-    targetCPUUtilizationPercentage: 70
 
 config:
   numberOfJobsPerWorker: 50
@@ -386,6 +376,26 @@ postgres:
     limits:
       cpu: "16"
       memory: "64Gi"
+```
+
+### Autoscaling
+
+If your deployment experiences bursty traffic, you can enable autoscaling to scale the number of API servers and queue workers to handle the load.
+
+Here is a sample configuration for autoscaling for high reads and high writes:
+
+```yaml
+api:
+  autoscaling:
+    enabled: true
+    minReplicas: 15
+    maxReplicas: 25
+
+queue:
+  autoscaling:
+    enabled: true
+    minReplicas: 10
+    maxReplicas: 20
 ```
 
 <Note>


src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+## Scaling for write load
+
+Write load is primarily driven by the following factors:
+
+- Creation of new runs
+- Creation of new checkpoints during run execution
+- Writing to long term memory
+- Creation of new threads
+- Creation of new assistants
+- Execution of periodic cron jobs
```
hinthornw: The cron jobs folds into the new runs stuff so i'd probably drop

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
+---
+title: Configure LangSmith Deployment for scale
+sidebarTitle: Deployment
+---
+
+The default configuration for self-hosted LangSmith Deployment can handle substantial read and write load, and you can configure your instance to achieve even higher scale as necessary. This page describes scaling considerations and provides examples to help configure your deployment.
```
hinthornw: I think this reads a bit too much like "we tuned it partway and you can tune it all the way", but the message of this doc from what I can tell is more that "we tuned it to scale well across many types of workloads. You may get better performance characteristics by further tuning based on your code and usage patterns"

hinthornw: It would also probably be helpful to organize based on what you can configure in our cloud offering vs. self-hosted, since we don't let them tune cpu, etc. today

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+- Creation of new checkpoints during run execution
+- Writing to long term memory
+- Creation of new threads
+- Creation of new assistants
+- Execution of periodic cron jobs
+
+The following systems are primarily responsible for handling write load:
+
+- API server: Handles initial request and persistence of data to the database.
+- Queue: Handles the execution of runs.
```
hinthornw: I'd say "Queue worker: ..."

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+- Redis: Handles the storage of ephemeral data about on-going runs.
+- Postgres: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory
+
+### Best practices for scaling the write path
+
+#### Enable the use of queue replicas
+
+By default, the api replica manages the queue and does not use queue replicas. You can enable the use of queue replicas by setting the `queue.enabled` configuration to `true`.
```
hinthornw: This depends on deployment scenario - the queue yaml configuration is only if they're self-hosting the helm chart. We should make it obvious who should read (or skip) a given section

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+This will allow the api replica to offload the queue management to the queue replicas, significantly reducing the load on the api replica and allowing it to focus on handling requests.
+
+#### Change N_JOBS_PER_WORKER based on assistant characteristics
+
+The default value of N_JOBS_PER_WORKER is 10. You can change this value to scale the maximum number of runs that can be executed at a time by a single queue replica based on the characteristics of your assistant.
+
+Some general guidelines for changing N_JOBS_PER_WORKER:
+
+- If your assistant is computationally intensive, lower N_JOBS_PER_WORKER to handle fewer concurrent runs per replica or keep it at the default.
```
hinthornw: I'd only lower it if you know it's compute intensive AND you see that the execution of one run slows down another probably (?)

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+- If your assistant is computationally lightweight, increase N_JOBS_PER_WORKER to handle more concurrent runs per replica.
```
hinthornw: it's more about cpu-bound vs. io-bound i think. io-bound, async work can be scaled out a lot.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+```
+
+So the minimum number of workers you should provision to support your expected steady state throughput is:
+```
+number_of_queue_replicas = throughput_per_second * average_run_execution_time_seconds / N_JOBS_PER_WORKER
+```
+
+#### Configure autoscaling for bursty workloads
+
+Autoscaling is disabled by default, but should be configured for bursty workloads. Using the same calculations as above, you can determine the maximum number of workers you should allow the autoscaler to scale to based on maximum expected throughput.
```
hinthornw: this isn't true in cloud, right? Unsure about in helm

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+#### Avoid synchronous blocking operations
+
+Avoid synchronous blocking operations in your code. If required, set BG_JOB_ISOLATED_LOOPS to True to run them in a separate event loop and avoid blocking the main event loop.
```
hinthornw: `loops` should be plural

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+#### Minimize redundant checkpointing
+
+Minimize redundant checkpointing by setting `checkpoint_during` to the minimum value necessary to ensure your data is durable.
```
hinthornw: This was deprecated in favor of `durability` arg, (`durability="exit"`)

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+- Getting the results of a run
+- Getting the state of a thread
+- Searching for runs, threads, cron jobs and assistants
+- Retrieving checkpoints and long term memory
+
+The following systems are primarily responsible for handling read load:
+
+- API server: Handles the request and direct retrieval of data from the database.
+- Database: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory.
```
hinthornw: We call it `Postgres` above. should be consistent

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+- Redis: Handles the storage of ephemeral data about on-going runs, including streaming messages from queue replicas to api replicas.
+
+### Best practices for scaling the read path
+
+#### Use pagination and filtering to reduce the number of resources returned per request
```
hinthornw: this is done by default, right?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+- Searching for runs, threads, cron jobs and assistants
+- Retrieving checkpoints and long term memory
+
+The following systems are primarily responsible for handling read load:
+
+- API server: Handles the request and direct retrieval of data from the database.
+- Database: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory.
+- Redis: Handles the storage of ephemeral data about on-going runs, including streaming messages from queue replicas to api replicas.
+
+### Best practices for scaling the read path
```
hinthornw, line 139: I think another one we should mention is to use the `/threads/join` route rather than polling for thread status (and to avoid polilng in general) to reduce the load on the database.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+If you are searching for resources, use pagination and filtering to reduce the number of resources returned per request.
+
+#### Set a TTL on threads and checkpoints to automatically delete old data
```
hinthornw: this just applies to threads, the checkpoints are implicitly included

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+Set a TTL on threads and checkpoints to automatically delete old data.
+
+#### Use resumable streams for extra messaging resilience
```
hinthornw: this feels out of place

src/docs.json
```diff
@@ -1362,7 +1362,13 @@
                     {
                       "group": "Configuration",
                       "pages": [
-                        "langsmith/self-host-scale",
+                        {
+                          "group": "Configuration for scale",
+                          "pages": [
+                            "langsmith/self-host-scale",
+                            "langsmith/deployments-scale"
```
andrewnguonly: Suggestion: I think we should revert this change and move the `deployments-scale` page under the `Deployment` tab.

After some confusion about "one time platform" setup vs "ongoing deployment" setup, we decided to move the LSD deployment docs out of the `Platform setup` tab.

I think this content is more appropriate for `Deployment` > `Deployment guides`, for example. Thoughts @katmayb?

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
+---
+title: Configure LangSmith Deployment for scale
```
andrewnguonly: Nit: `LangSmith Deployment` refers to the platform feature (e.g. control plane) that deploys `Agent Server` instances. This content is technically for scaling `Agent Server` instances.

We may want to change the naming in other places in this PR as well.

We've had issues with naming before. @katmayb can you confirm what the latest recommendation/preference is?

src/langsmith/self-host-scale.mdx
```diff
@@ -1,6 +1,6 @@
 ---
-title: Configure LangSmith for scale
-sidebarTitle: Configure for scale
+title: Configure LangSmith Observability and Evaluation for scale
+sidebarTitle: Observability and Evaluation
```
andrewnguonly: Maybe revert this based on my other comment.

src/langsmith/deployments-scale.mdx
andrewnguonly: Preference: Remove the first sentence.

Why? Language like "substantial" and "higher scale" is imprecise, relative, and embellishes the product a bit. My preference for technical documentation is that the content is plain and un-opinionated. Just state what can be done and what the features are.

The alternative is to additionally publish the raw benchmark numbers (which we should do at some point).

Defer to @katmayb for language and style preferences. Not blocking.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+sidebarTitle: Deployment
+---
+
+The default configuration for self-hosted LangSmith Deployment can handle substantial read and write load, and you can configure your instance to achieve even higher scale as necessary. This page describes scaling considerations and provides examples to help configure your deployment.
+
+For example configurations, refer to [Example Deployment configurations for scale](#example-deployment-configurations-for-scale).
+
+## Summary
+
+The table below provides an overview comparing different LangSmith Deployment configurations for various load patterns (read requests per second / write requests per second) and standard assistant characteristics (average run execution time of 1 seconds, moderate CPU and memory usage):
```
andrewnguonly: `1 seconds` --> `1 second`

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+Write load is primarily driven by the following factors:
+
+- Creation of new runs
+- Creation of new checkpoints during run execution
+- Writing to long term memory
+- Creation of new threads
+- Creation of new assistants
+- Execution of periodic cron jobs
+
+The following systems are primarily responsible for handling write load:
```
andrewnguonly: Suggestion: `systems` --> `components`?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+| **Queue replicas**<br />(1 CPU, 2Gi per replica) | 1 (default) | 10 | 1 (default) | 5 | 10 |
+| **N_JOBS_PER_WORKER** | 10 (default) | 50 | 10 | 10 | 50 |
+| **Redis resources** | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) |
+| **Postgres resources** | 2 CPU<br />8 Gi (default) | 4 CPU<br />16 Gi memory | 4 CPU<br />16 Gi per replica | 4 CPU<br />16 Gi memory | 8 CPU<br />32 Gi memory |
+
+Below we go into more details about the request handling and scaling patterns as well as provide configuration examples for your LangSmith Deployment.
+
+## Scaling for write load
+
+Write load is primarily driven by the following factors:
```
andrewnguonly, line 12: Even "deletes" are considered "write" operations. Should we mention it? For example, frequently deleting threads or something like that?

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+- API server: Handles initial request and persistence of data to the database.
+- Queue: Handles the execution of runs.
+- Redis: Handles the storage of ephemeral data about on-going runs.
+- Postgres: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory
```
andrewnguonly: `memory` --> `memory.`

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+### Best practices for scaling the write path
+
+#### Enable the use of queue replicas
+
+By default, the api replica manages the queue and does not use queue replicas. You can enable the use of queue replicas by setting the `queue.enabled` configuration to `true`.
+
+```yaml
```
andrewnguonly: This code block feels out of context. For example, it doesn't apply to SaaS deployments. Can we add a note that says this is for self-hosted deployments?

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+queue:
+  enabled: true
+```
+
+This will allow the api replica to offload the queue management to the queue replicas, significantly reducing the load on the api replica and allowing it to focus on handling requests.
+
+#### Change N_JOBS_PER_WORKER based on assistant characteristics
```
andrewnguonly: Suggestion: Add a link to https://docs.langchain.com/langsmith/env-var#n-jobs-per-worker somewhere in this section.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+The default value of N_JOBS_PER_WORKER is 10. You can change this value to scale the maximum number of runs that can be executed at a time by a single queue replica based on the characteristics of your assistant.
```
andrewnguonly: Double checking: `assistant` or `agent`?

@katmayb thoughts?

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+Some general guidelines for changing N_JOBS_PER_WORKER:
+
+- If your assistant is computationally intensive, lower N_JOBS_PER_WORKER to handle fewer concurrent runs per replica or keep it at the default.
+- If your assistant is computationally lightweight, increase N_JOBS_PER_WORKER to handle more concurrent runs per replica.
+
+There is no upper limit to N_JOBS_PER_WORKER. However, the polling algorithm queues use to pick up new runs is greedy, meaning they will try to pick up as many runs as possible and begin executing them right away. If N_JOBS_PER_WORKER is too high and traffic is burstly, you may see uneven replica utilization and potentially longer run execution times.
```
andrewnguonly: Nit: `queues use` --> this reads slightly weird. How about just `the polling algorithm to pick up new runs is greedy`?

src/langsmith/deployments-scale.mdx
andrewnguonly: Commenting again. We need to distinguish which sections are applicable to SaaS, self-hosted, or both.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+So the minimum number of workers you should provision to support your expected steady state throughput is:
+```
+number_of_queue_replicas = throughput_per_second * average_run_execution_time_seconds / N_JOBS_PER_WORKER
+```
+
+#### Configure autoscaling for bursty workloads
+
+Autoscaling is disabled by default, but should be configured for bursty workloads. Using the same calculations as above, you can determine the maximum number of workers you should allow the autoscaler to scale to based on maximum expected throughput.
+
+#### Avoid synchronous blocking operations
```
andrewnguonly: Suggestions for this section:
1. Show a code example of sync vs async code.
2. Explain some of the symptoms that may appear if implementing sync code.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+Avoid synchronous blocking operations in your code. If required, set BG_JOB_ISOLATED_LOOPS to True to run them in a separate event loop and avoid blocking the main event loop.
+
+#### Minimize redundant checkpointing
```
andrewnguonly: Suggestion: Show client code example for how to set `durability`.

src/langsmith/deployments-scale.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+### Best practices for scaling the read path
+
+#### Use pagination and filtering to reduce the number of resources returned per request
+
+If you are searching for resources, use pagination and filtering to reduce the number of resources returned per request.
+
+#### Set a TTL on threads and checkpoints to automatically delete old data
+
+Set a TTL on threads and checkpoints to automatically delete old data.
```
andrewnguonly: Suggestion: Link to docs https://docs.langchain.com/langsmith/configure-ttl

andrewnguonly (approved):
Do you feel like the section headers are generic enough that we can continue to add to this page? For example, if we want to add scaling docs for Postgres (e.g. increasing `max_connections` to handle scale out).

src/langsmith/deployments-scale.mdx
jdrogers940: Yeah great point, I definitely need to reword.

Can definitely break the page up a bit more by deployment type too. I think I'll add subsections to the various recommendations sections that are self-hosted only and make those drop downs with a note.

src/docs.json
jdrogers940: Sounds good to me - I do think that'll make this a little less confusing with some of the terminology and I can break the page down more clearly by what deployment type each setup has to clarify a bit more.

src/langsmith/deployments-scale.mdx
jdrogers940: Ah, good to know, I'll update that.

src/langsmith/deployments-scale.mdx
jdrogers940: Refactored to break out guidance in these sections to highlight the specific parameters that self-hosted have to configure, with a banner highlighting this is already done for cloud.

src/langsmith/deployments-scale.mdx
jdrogers940: I guess it depends on how compute intensive. But this value is pretty conservative so I do think it would be rare to lower. Will update.

src/langsmith/deployments-scale.mdx
jdrogers940: No, in cloud this is enabled by default. Based on the helm chart, this is off by default there though so moved to the self-hosted section.

src/langsmith/deployments-scale.mdx
jdrogers940: Good call. Keeping as Postgres for now.

src/langsmith/deployments-scale.mdx
jdrogers940: Pagination is, filtering is not. Will update accordingly with some examples.

src/langsmith/deployments-scale.mdx
jdrogers940: Yeah removed. I was trying to callout that if you use resumable streams, you can end up with high redis memory usage. But I think the better place to call that out is in `RESUMABLE_STREAM_TTL_SECONDS` env var section. Updated accordingly.

src/langsmith/deployments-scale.mdx
jdrogers940: Yeah I generally agree. Was matching style of the Langsmith scaling page. I've changed per Will's call out above to explain why the page exists. Do agree we should publish benchmarks soon too.

src/langsmith/agent-server-scale.mdx
jdrogers940: Yeah good call! Will add to the list in case folks are doing really aggressive manual deletes. I don't think we have any particular recommendations on that route now though.

src/langsmith/deployments-scale.mdx
jdrogers940: Moved to a subsection to clarify this.

src/langsmith/deployments-scale.mdx
jdrogers940: Will look for guidance from @katmayb here. I like assistant cause it matches the term we use in Agent Server, but can see that being confusing to folks potentially.

src/langsmith/deployments-scale.mdx
katmayb: Yeah, I think Agent Server sis preferable here.

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
+---
+title: Configure LangSmith Agent Server for scale
+sidebarTitle: Configure Agent Server for scale
+---
+
+The default configuration for LangSmith Agent Server is designed to handle substantial read and write load across a variety of different workloads. By following the best practices outlined below, you can tune your Agent Server to perform optimally for your specific workload. This page describes scaling considerations for the Agent Server and provides examples to help configure your deployment.
+
+For example self-hosted configurations, refer to [Example Agent Server configurations for scale](#example-agent-server-configurations-for-scale).
```
katmayb:
```suggestion
For some example self-hosted configurations, refer to the [Example Agent Server configurations for scale](#example-agent-server-configurations-for-scale) section.
```

Adding the "some" just because it's easy to read this as "For example, ...."

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+## Scaling for write load
+
+Write load is primarily driven by the following factors:
+
+- Creation of new runs
+- Creation of new checkpoints during run execution
+- Writing to long term memory
+- Creation of new threads
+- Creation of new assistants
+- Deletion of runs, checkpoints, threads, assistants and cron jobs
```
katmayb:
```suggestion
- Creation of new [runs](/langsmith/observability-concepts#runs)
- Creation of new [checkpoints](/oss/python/langgraph/persistence#checkpoints) during run execution
- Writing to long term memory
- Creation of new [threads](/langsmith/use-threads)
- Creation of new [assistants](/langsmith/assistants)
- Deletion of runs, checkpoints, threads, assistants and cron jobs
```

It'd be great to add some links here, these look relevant?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+#### Change N_JOBS_PER_WORKER based on assistant characteristics
+
+The default value of [N_JOBS_PER_WORKER](/langsmith/env-var#n-jobs-per-worker) is 10. You can change this value to scale the maximum number of runs that can be executed at a time by a single queue worker based on the characteristics of your assistant.
+
+Some general guidelines for changing N_JOBS_PER_WORKER:
+
+- If your assistant is CPU bounded, the default value of 10 is likely sufficient. You might lower N_JOBS_PER_WORKER if you notice excessive CPU usage on queue workers or delays in run execution.
+- If your assistant is IO bounded, increase N_JOBS_PER_WORKER to handle more concurrent runs per worker.
+
+There is no upper limit to N_JOBS_PER_WORKER. However, Queue workers are greedy when fetching new runs, meaning they will try to pick up as many runs they have available jobs and begin executing them right away. If N_JOBS_PER_WORKER is too high and traffic is burstly, you may see uneven worker utilization and potentially longer run execution times.
```
katmayb:
```suggestion
There is no upper limit to `N_JOBS_PER_WORKER`. However, queue workers are greedy when fetching new runs, which means they will try to pick up as many runs as they have available jobs and begin executing them immediately. Setting `N_JOBS_PER_WORKER` too high in environments with bursty traffic can lead to uneven worker utilization and increased run execution times.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+- API server: Handles initial request and persistence of data to the database.
+- Queue worker: Handles the execution of runs.
+- Redis: Handles the storage of ephemeral data about on-going runs.
+- Postgres: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory.
+
+### Best practices for scaling the write path
+
+#### Change N_JOBS_PER_WORKER based on assistant characteristics
+
+The default value of [N_JOBS_PER_WORKER](/langsmith/env-var#n-jobs-per-worker) is 10. You can change this value to scale the maximum number of runs that can be executed at a time by a single queue worker based on the characteristics of your assistant.
```
katmayb: What are your thoughts on code highlighting `N_JOBS_PER_WORKER` in the body text since it is an env variable?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+Some general guidelines for changing N_JOBS_PER_WORKER:
+
+- If your assistant is CPU bounded, the default value of 10 is likely sufficient. You might lower N_JOBS_PER_WORKER if you notice excessive CPU usage on queue workers or delays in run execution.
+- If your assistant is IO bounded, increase N_JOBS_PER_WORKER to handle more concurrent runs per worker.
+
+There is no upper limit to N_JOBS_PER_WORKER. However, Queue workers are greedy when fetching new runs, meaning they will try to pick up as many runs they have available jobs and begin executing them right away. If N_JOBS_PER_WORKER is too high and traffic is burstly, you may see uneven worker utilization and potentially longer run execution times.
+
+#### Avoid synchronous blocking operations
+
+Avoid synchronous blocking operations in your code. If required, set [BG_JOB_ISOLATED_LOOPS](/langsmith/env-var#bg-job-isolated-loops) to True to execute each run in a separate event loop and avoid blocking the main event loop.
```
katmayb:
```suggestion
Avoid synchronous blocking operations in your code. If required, set [`BG_JOB_ISOLATED_LOOPS`](/langsmith/env-var#bg-job-isolated-loops) to `True` to execute each run in a separate event loop and avoid blocking the main event loop.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+#### Minimize redundant checkpointing
+
+Minimize redundant checkpointing by setting [`durability`](/oss/langgraph/durable-execution#durability-modes) to the minimum value necessary to ensure your data is durable.
+
+#### Self-hosted
+
+<Note>
+These settings are only required for self-hosted deployments. By default, Cloud deployments already have these best practices enabled.
```
katmayb:
```suggestion
These settings are only required for [self-hosted](/langsmith/self-hosted) deployments. By default, [cloud](/langsmith/cloud) deployments already have these best practices enabled.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+</Note>
+
+###### Enable the use of queue workers
```
katmayb:
```suggestion
##### Enable the use of queue workers
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+By default, the api server manages the queue and does not use queue workers. You can enable the use of queue workers by setting the `queue.enabled` configuration to `true`.
```
katmayb:
```suggestion
By default, the API server manages the queue and does not use queue workers. You can enable the use of queue workers by setting the `queue.enabled` configuration to `true`.
```
When we say API server, are we talking about the Agent Server? (This is more a question for me, btw!)

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+```yaml
+queue:
+  enabled: true
+```
+
+This will allow the api server to offload the queue management to the queue workers, significantly reducing the load on the api server and allowing it to focus on handling requests.
```
katmayb:
```suggestion
This will allow the API server to offload the queue management to the queue workers, significantly reducing the load on the API server and allowing it to focus on handling requests.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+##### Support a number of jobs equal to expected throughput
+
+The more runs you execute in parallel, the more jobs you will need to handle the load. There are two main parameters to scale the available jobs:
+
+- number_of_queue_workers: The number of queue workers provisioned.
+- N_JOBS_PER_WORKER: The number of runs that can be executed at a time by a single queue worker. Defaults to 10.
```
katmayb:
```suggestion
- `number_of_queue_workers`: The number of queue workers provisioned.
- `N_JOBS_PER_WORKER`: The number of runs that a single queue work can execute at a time. Defaults to 10.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+```
+available_jobs = number_of_queue_workers * N_JOBS_PER_WORKER
+```
+
+Throughput is then the number of runs that can be executed per second by the available jobs:
+```
+throughput_per_second = available_jobs / average_run_execution_time_seconds
+```
+
+So the minimum number of queue workers you should provision to support your expected steady state throughput is:
```
katmayb:
```suggestion
Therefore, the minimum number of queue workers you should provision to support your expected steady state throughput is:
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+```
+number_of_queue_workers = throughput_per_second * average_run_execution_time_seconds / N_JOBS_PER_WORKER
+```
+
+##### Configure autoscaling for bursty workloads
+
+Autoscaling is disabled by default, but should be configured for bursty workloads. Using the same calculations as above, you can determine the maximum number of queue workers you should allow the autoscaler to scale to based on maximum expected throughput.
```
katmayb:
```suggestion
Autoscaling is disabled by default, but should be configured for bursty workloads. Using the same calculations as the [previous section](#support-a-number-of-jobs-equal-to-expected-throughput), you can determine the maximum number of queue workers you should allow the autoscaler to scale to based on maximum expected throughput.
```

katmayb: (Easier for people using screen readers to just link to reference rather than using directional language like "above")

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+## Scaling for read load
+
+Read load is primarily driven by the following factors:
+
+- Getting the results of a run
+- Getting the state of a thread
+- Searching for runs, threads, cron jobs and assistants
+- Retrieving checkpoints and long term memory
```
katmayb: Could you add the same set of links here, in case a user lands in this section rather than going from the top of the page to the bottom?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+The following components are primarily responsible for handling read load:
+
+- API server: Handles the request and direct retrieval of data from the database.
```
katmayb, line 135: I didn't mention this in the previous list of components in the write load section, but do you think it's worthwhile linking to these components in some of the architecture docs we have?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+- Postgres: Handles the storage of all data, including run, thread, assistant, cron job, checkpointing and long term memory.
+- Redis: Handles the storage of ephemeral data about on-going runs, including streaming messages from queue workers to api servers.
+
+### Best practices for scaling the read path
+
+#### Use filtering to reduce the number of resources returned per request
+
+Agent Server provides a search API for each resource type. These APIs implement pagination by default and offer many filtering options. Use filtering to reduce the number of resources returned per request and improve performance.
```
katmayb:
```suggestion
[Agent Server](/langsmith/agent-server) provides a search API for each resource type. These APIs implement pagination by default and offer many filtering options. Use filtering to reduce the number of resources returned per request and improve performance.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+#### Set a TTL on threads to automatically delete old data
+
+Set a TTL on threads to automatically clean up old data. Runs and checkpoints are automatically deleted when the associated thread is deleted.
```
katmayb: Would /langsmith/configure-ttl be a good link fit here?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+#### Avoid polling and use /join to monitor the state of a run
+
+Avoid polling the state of a run by using the `/join` API. This API allows you to monitor the state of a run and return a response with the final state once the run is complete.
```
katmayb: Is this referring to: https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html?_gl=1*1e82sx3*_gcl_au*MTM4NTQ5NjQwNy4xNzYxNzYwODY2*_ga*MTEzMTAyNjkxOC4xNzUzOTgyNTUw*_ga_47WX3HKKY2*czE3NjM0NzM2NzgkbzI5OSRnMSR0MTc2MzQ3NjI1NiRqMiRsMCRoMA..#tag/thread-runs/get/threads/{thread_id}/runs/{run_id}/join?

maybe we can say API endpoints if so?

If you can confirm that is the correct API ref, I'll add all the shorthand linking to the API refs once you've merged your PR

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+If you need to monitor the state of a run in real-time, you can use the `/stream` API instead.
+
+#### Self-hosted
+
+<Note>
+These settings are only required for self-hosted deployments. By default, Cloud deployments already have these best practices enabled.
```
katmayb: Same links as previously here.

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+</Note>
+
+##### Configure autoscaling for bursty workloads
+
+Autoscaling is disabled by default, but should be configured for bursty workloads. You can determine the maximum number of api servers you should allow the autoscaler to scale to based on maximum expected throughput.
+
+## Example self-hosted Agent Server configurations
+
+<Note>
+The exact optimal configuration depends on your application complexity, request patterns, and data requirements. Use the examples below in combination with the information above and your specific usage to update your deployment configuration as needed. If you have any questions, please reach out to the LangChain team.
```
katmayb:
```suggestion
The exact optimal configuration depends on your application complexity, request patterns, and data requirements. Use the following examples in combination with the information in the previous sections and your specific usage to update your deployment configuration as needed. If you have any questions, reach out to the LangChain team at [support@langchain.dev](mailto:support@langchain.dev).
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+</Note>
+
+The table below provides an overview comparing different LangSmith Agent Server configurations for various load patterns (read requests per second / write requests per second) and standard assistant characteristics (average run execution time of 1 second, moderate CPU and memory usage):
```
katmayb:
```suggestion
The following table provides an overview comparing different LangSmith Agent Server configurations for various load patterns (read requests per second / write requests per second) and standard assistant characteristics (average run execution time of 1 second, moderate CPU and memory usage):
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+|  | **[Low / low](#low-reads-low-writes)** | **[Low / high](#low-reads-high-writes)** | **[High / low](#high-reads-low-writes)** | [Medium / medium](#medium-reads-medium-writes) | [High / high](#high-reads-high-writes) |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| <Tooltip tip="Number of write requests being processed by the Deployment per second">Write requests per second</Tooltip> | 5 | 5 | 500 | 50 | 500 |
```
katmayb:
```suggestion
| <Tooltip tip="Number of write requests being processed by the deployment per second">Write requests per second</Tooltip> | 5 | 5 | 500 | 50 | 500 |
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+| <Tooltip tip="Number of read requests being processed by the Deployment per second">Read requests per second</Tooltip> | 5 | 500 | 5 | 50 | 500 |
```
katmayb:
```suggestion
| <Tooltip tip="Number of read requests being processed by the deployment per second">Read requests per second</Tooltip> | 5 | 500 | 5 | 50 | 500 |
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+| **API servers**<br />(1 CPU, 2Gi per server) | 1 (default) | 6 | 10 | 3 | 15 |
+| **Queue workers**<br />(1 CPU, 2Gi per worker) | 1 (default) | 10 | 1 (default) | 5 | 10 |
+| **N_JOBS_PER_WORKER** | 10 (default) | 50 | 10 | 10 | 50 |
+| **Redis resources** | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) | 2 Gi (default) |
+| **Postgres resources** | 2 CPU<br />8 Gi (default) | 4 CPU<br />16 Gi memory | 4 CPU<br />16 Gi | 4 CPU<br />16 Gi memory | 8 CPU<br />32 Gi memory |
+
+Below we provide sample configurations to enable each of these setups. We define each load level as follows:
```
katmayb:
```suggestion
The following sample configurations enable each of these setups. Load levels are defined as:
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+- Low means roughly 5 requests per second
```
katmayb: Maybe roughly --> approximately?

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+- Medium means roughly 50 requests per second
+- High means roughly 500 requests per second
+
+### Low reads, low writes <a name="low-reads-low-writes"></a>
+
+The default LangSmith Deployment configuration will handle this load. No custom resource configuration is needed here.
```
katmayb:
```suggestion
The default [LangSmith Deployment](/langsmith/deployments) configuration will handle this load. No custom resource configuration is needed here.
```

src/langsmith/env-var.mdx
```diff
@@ -175,3 +175,7 @@ When a run is created and the output is streamed, the stream can be configured t
 See the @[Python][RunsClient.stream] and [JS/TS](https://langchain-ai.github.io/langgraphjs/reference/classes/sdk_client.RunsClient.html#stream) SDKs for more details on how to implement resumable streams.
 
 Defaults to `120` seconds.
+
+<Note>
+Setting a very high value for `RESUMABLE_STREAM_TTL_SECONDS` can result in substantial Redis memory usage when there are many concurrent runs with large or frequent streaming output. Set this value to the minimum value enable recovery during network interruptions and prefer checkpointing for long term durability and execution snapshotting.
```
katmayb:
```suggestion
Setting a very high value for `RESUMABLE_STREAM_TTL_SECONDS` can result in substantial Redis memory usage when there are many concurrent runs with large or frequent streaming output. Set this value to the minimum value to enable recovery during network interruptions and prefer checkpointing for long term durability and execution snapshotting.
```

Was that the missing word there?

katmayb (approved):
Added some very minor nits, mostly potential links!
Otherwise LGTM!

src/langsmith/agent-server-scale.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+#### Self-hosted
+
+<Note>
+These settings are only required for self-hosted deployments. By default, Cloud deployments already have these best practices enabled.
+</Note>
+
+##### Configure autoscaling for bursty workloads
+
+Autoscaling is disabled by default, but should be configured for bursty workloads. You can determine the maximum number of api servers you should allow the autoscaler to scale to based on maximum expected throughput.
```
joaquin-borggio-lc: If they are using managed, they still need guidance on what the default values are so they can translate it to a managed instance right?

src/langsmith/agent-server-scale.mdx
jdrogers940: Yeah links are a good call here. Will update these!

jdrogers940: Although going to leave off checkpoints and memory for now as we don't have good docs on those specific to LSD. Will take that as a follow up.

src/langsmith/agent-server-scale.mdx
jdrogers940: Very much in favor! Will make that change for all the environment variables throughout this page.

src/langsmith/agent-server-scale.mdx
jdrogers940: Ah talking about the LangGraph API server in this diagram: https://docs.langchain.com/langsmith/cloud

So the component of an Agent Server that manages serving user API requests, writing runs to the queue and streaming data back to users.

src/langsmith/agent-server-scale.mdx
jdrogers940: Yes was just thinking about this. I can do this as a follow up on the agent server page: https://docs.langchain.com/langsmith/agent-server I think we need a better architecture diagram there and we can layout each of these with their responsibilities in more detail.

src/langsmith/agent-server-scale.mdx
jdrogers940: Yes! Added the link and updated to call out setting on memory too.

src/langsmith/agent-server-scale.mdx
jdrogers940: That's correct! Updated to API endpoint.

That's the right ref for /join. Would be great to link to the API docs. Could you add for /stream below as well? https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html?_gl=1*1e82sx3*_gcl_au*MTM4NTQ5NjQwNy4xNzYxNzYwODY2*_ga*MTEzMTAyNjkxOC4xNzUzOTgyNTUw*_ga_47WX3HKKY2*czE3NjM0NzM2NzgkbzI5OSRnMSR0MTc2MzQ3NjI1NiRqMiRsMCRoMA..#tag/thread-runs/get/threads/{thread_id}/runs/{run_id}/stream

src/langsmith/deployments-scale.mdx
jdrogers940: Added a simple code snippet and some additional details.

src/langsmith/deployments-scale.mdx
jdrogers940: Added an example and quick blurb on durability defaults vs. more optimized setting.

src/langsmith/agent-server-scale.mdx
jdrogers940: That's fair - will add what the defaults are for Cloud to these if folks want. I add some examples in the section below on configuring auto-scaling, but will also break those out into a separate spot.

PR #1442, opened 2025-11-15
oss(ps): import from `langchain` instead of `langchain_core` where possible

mdrxy:
These are re-exports

In attempt to keep it consistent & following community feedback

https://www.reddit.com/r/LangChain/comments/1oxtc9o/comment/nozrimm/


eyurtsev:
Nice! Do we have anything else for messages that we need to update as well?

mdrxy:
@eyurtsev using `.content_blocks` where possible as opposed to `.content` I think would be a good next step

PR #1547, opened 2025-11-20
Iterative improvements to OSS <> LangSmith

katmayb:
This PR adds some clarification to the OSS docs for deployment and obs in LangSmith. There is likely more work to be done here, but this PR does the following:

- Updates the nav to make the workflow of going from dev to production with LangSmith.
- Updates some of the intros to that content to contextualize LangChain or LangGraph --> LangSmith.
- More links!

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -194,12 +194,17 @@
                       ]
                     },
                     {
-                      "group": "Use in production",
+                      "group": "Agent development",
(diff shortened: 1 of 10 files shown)


lnhsingh (approved):
Nice!

katmayb:
TFTR!

PR #1553, opened 2025-11-20
Add reserved arg names to tools

lnhsingh:
Fixes DOC-475

### DIFF
--- a/src/oss/langchain/tools.mdx
+++ b/src/oss/langchain/tools.mdx
@@ -139,25 +139,17 @@ Define complex inputs with Pydantic models or JSON schemas:
     ```
 </CodeGroup>
 
+:::python
 ### Reserved argument names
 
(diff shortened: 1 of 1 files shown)


src/oss/langchain/tools.mdx
```diff
@@ -138,6 +138,26 @@ Define complex inputs with Pydantic models or JSON schemas:
...
+
+To access runtime information, use the @[`ToolRuntime`] parameter instead of naming your own arguments `config` or `runtime`.
+:::
+
+:::js
+| Parameter name | Purpose |
+|----------------|---------|
+| `config` | Reserved for passing runtime context and configuration to tools |
+
+To access runtime configuration, use the `config` parameter as the second argument to your tool function (not in the schema).
```
hntrl: think the term "reserved argument names" is much more applicable to python then it is to JS. Think it's fine if this is a python only section

PR #1570, opened 2025-11-21
Add reference for metadata parameters and relationships

katmayb:
Fixes DOC-433

## Preview

https://langchain-5e9cc07a-preview-lspara-1763761024-e80127d.mintlify.app/langsmith/ls-metadata-parameters

### DIFF
--- a/src/langsmith/ls-metadata-parameters.mdx
+++ b/src/langsmith/ls-metadata-parameters.mdx
@@ -65,7 +65,6 @@ For more comprehensive tracking, you can include additional configuration parame
     metadata={
         "ls_provider": "openai",
         "ls_model_name": "gpt-4o",
-        "ls_model_type": "chat",
         "ls_temperature": 0.7,
         "ls_max_tokens": 4096,
         "ls_stop": ["END"],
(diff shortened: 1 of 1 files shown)


src/langsmith/ls-metadata-parameters.mdx
```diff
@@ -0,0 +1,584 @@
...
+<CodeGroup>
+
+```python Python
+from langsmith import traceable
+
+@traceable(
+    run_type="llm",
+    metadata={
+        "ls_provider": "my_provider",
+        "ls_model_name": "my_custom_model"
```
jacoblee93, line 30: I *think* this should work without model type but we should double check

src/langsmith/ls-metadata-parameters.mdx
```diff
@@ -0,0 +1,584 @@
...
+
+<CodeGroup>
+
+```python Python
+@traceable(
+    run_type="llm",
+    metadata={
+        "ls_provider": "openai",
+        "ls_model_name": "gpt-4o",
+        "ls_model_type": "chat",
```
jacoblee93: Sorry to flip flip - if the above works, let's actually remove this from examples

src/langsmith/ls-metadata-parameters.mdx
katmayb: Yes, I ran this and it seemed to work fine — trace in project containing metadata. So, I've taken out mentions of `ls_model_type` from the page. Let me know if there is something else I should test with regards to that.

lnhsingh (approved):

PR #1581, opened 2025-11-22
1.1 changelog

sydney-runkle:
Ended up pivoting to a simple changelog approach for the 1.1 release notes. This is more sustainable and motivates us to write up robust docs that we can link out to for new features and changes.

Post actual release, we can link out to the changelog on github w/ nitty gritty details.

Also fixing some broken links :)


pipeline/preprocessors/link_map.py
sydney-runkle, line 1: removing dupes

PR #1590, opened 2025-11-22
fix: agent jump example, enhanced state docs for JS and more

christian-bromann:
Some further improvements on:

- custom state schema: more details on how this is useful
- custom context schema: new section (was not documented before)
- adding references to published middleware videos:
  Screenshot 2025-11-22 at 3 22 10 PM
- (js only): add docs on public/private state
- (js only): fix example for jump to section
- (js only): improvements to openAIModerationMiddleware

### DIFF
--- a/src/oss/langchain/middleware/built-in.mdx
+++ b/src/oss/langchain/middleware/built-in.mdx
@@ -407,6 +407,18 @@ const agent = createAgent({
     For complete examples, configuration options, and integration patterns, see the [Human-in-the-loop documentation](/oss/langchain/human-in-the-loop).
 </Tip>
 
+:::python
+<Callout icon="circle-play" iconType="solid">
+    Watch this [video guide](https://www.youtube.com/watch?v=SpfT6-YAVPk) demonstrating Human-in-the-loop middleware behavior.
+</Callout>
+:::
+
+:::js
+<Callout icon="circle-play" iconType="solid">
+    Watch this [video guide](https://www.youtube.com/watch?v=tdOeUVERukA) demonstrating Human-in-the-loop middleware behavior.
+</Callout>
+:::
+
 ### Model call limit
 
 Limit the number of model calls to prevent infinite loops or excessive costs. Model call limit is useful for the following:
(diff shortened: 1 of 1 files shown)


src/oss/langchain/middleware/custom.mdx
```diff
@@ -623,23 +841,39 @@ class BlockedContentMiddleware(AgentMiddleware):
...
+  model: "gpt-4o",
+  middleware: [
+    createMiddleware({
+      name: "BlockedContentMiddleware",
+      beforeModel: {
+        canJumpTo: ["end"],
+        hook: (state) => {
+          if (state.messages.at(-1)?.content.includes("BLOCKED")) {
+            return {
+              messages: [new AIMessage("I cannot respond to that request.")],
```
mcavdar, line 856:
```suggestion
      afterModel: {
        canJumpTo: ["end"],
        hook: (state) => {
          if (state.messages.at(-1)?.content.includes("BLOCKED")) {
            return {
              messages: [
                ...state.messages.slice(0, -1),
                {
                  ...state.messages.at(-1),
                  content: "I cannot respond to that request."
                }
              ],
```

mcavdar: It might be better if we had an afterModel example that updates the agent message, either here or in the guardrails docs. What do you think?

Edit: Actually, we already have this in guardrails, but using afterAgent. Having another example with afterModel still makes sense.

src/oss/langchain/middleware/custom.mdx
```diff
@@ -623,23 +841,39 @@ class BlockedContentMiddleware(AgentMiddleware):
...
+          return;
+        },
+      },
+    }),
+  ],
 });
-```
 
+const result = await agent.invoke({
+    messages: "Hello, world! BLOCKED"
```
mcavdar, line 868:
```suggestion
    messages: "Hello, world! Return a message that contains BLOCKED"
```

christian-bromann:
Closing in favor of #1608

PR #1608, opened 2025-11-24
fix: agent jump example, enhanced state docs for JS and more

christian-bromann:
Some further improvements on:

- custom state schema: more details on how this is useful
- custom context schema: new section (was not documented before)
- adding references to published middleware videos:
  Screenshot 2025-11-22 at 3 22 10 PM
- (js only): add docs on public/private state
- (js only): fix example for jump to section
- (js only): improvements to openAIModerationMiddleware

### DIFF
--- a/src/oss/langchain/middleware/custom.mdx
+++ b/src/oss/langchain/middleware/custom.mdx
@@ -552,6 +552,7 @@ console.log(result._internalFlag); // undefined
 
 :::
 
+:::js
 ## Custom context
 
 Middleware can define a custom context schema to access per-invocation metadata. Unlike state, context is read-only and not persisted between invocations. This makes it ideal for:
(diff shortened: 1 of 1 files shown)


src/oss/langchain/middleware/built-in.mdx
```diff
@@ -2876,6 +2930,28 @@ agent = create_agent(
 )
 ```
 :::
+:::js
+**API reference:** @[`openAIModerationMiddleware`]
```
hntrl: wouldn't hurt to call out that:
1. this only works with openai models
2. this won't work with model strings

src/oss/langchain/middleware/custom.mdx
```diff
@@ -365,7 +365,12 @@ const loggingMiddleware = createMiddleware({
 
 ## Custom state schema
 
-Middleware can extend the agent's state with custom properties.
+Middleware can extend the agent's state with custom properties. This enables middleware to:
+
+- **Track state across execution**: Maintain counters, flags, or other values that persist throughout the agent's execution lifecycle
+- **Share data between hooks**: Pass information from `before_model` to `after_model` or between different middleware instances
```
hntrl, line 382: would flag this as something we should make language specific: `before_model` in python is defined as `beforeModel` in js

src/oss/langchain/middleware/built-in.mdx
christian-bromann: > 1. this only works with openai models

The section starts with _Middleware specifically designed for OpenAI models._

> 2\. this won't work with model strings

It does. The middleware `model` option can be a string too.

src/oss/langchain/middleware/custom.mdx
```diff
@@ -365,7 +365,17 @@ const loggingMiddleware = createMiddleware({
 
 ## Custom state schema
 
-Middleware can extend the agent's state with custom properties.
+Middleware can extend the agent's state with custom properties. This enables middleware to:
```
eyurtsev, line 378: Will readers know what agent state is at this point? (is there a cross-link that could be added if not)

src/oss/langchain/middleware/custom.mdx
```diff
@@ -498,6 +508,219 @@ const result = await agent.invoke({
...
+// result only contains publicCounter, not _internalFlag
+console.log(result.publicCounter); // 1
+console.log(result._internalFlag); // undefined
+```
+
+:::
+
+## Custom context
+
+Middleware can define a custom context schema to access per-invocation metadata. Unlike state, context is read-only and not persisted between invocations. This makes it ideal for:
```
eyurtsev, line 558: In python middleware cannot define custom context yet. it can only access existing context (but can't require that certain keys exist in the context).

If the intention is context extension, then we should only do this for JS (entire section heading should be wrapped in that)

src/oss/langchain/middleware/custom.mdx
christian-bromann: We don't have any section describing what state and context mean in for React agent. I suggest adding that seperately?

eyurtsev:
Thank you looks great!!

PR #1622, opened 2025-11-25
fix(langchain): add missing section on handling tool errors

christian-bromann:
The section was missing examples.

### DIFF
--- a/src/oss/python/migrate/langchain-v1.mdx
+++ b/src/oss/python/migrate/langchain-v1.mdx
@@ -592,19 +592,52 @@ You can now configure the handling of tool errors with middleware implementing t
 
 <CodeGroup>
 ```python v1 (new)
+from langchain.agents import create_agent
+from langchain.agents.middleware import wrap_tool_call
+from langchain.messages import ToolMessage
+
 
 @wrap_tool_call
-def retry_on_error(request, handler):
-    max_retries = 3
-    for attempt in range(max_retries):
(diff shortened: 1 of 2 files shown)


src/oss/python/migrate/langchain-v1.mdx
```diff
@@ -592,10 +592,45 @@ You can now configure the handling of tool errors with middleware implementing t
...
+from langchain.agents.middleware import wrap_tool_call
+from langchain.messages import ToolMessage
+
+
+@wrap_tool_call
+def handle_tool_errors(request, handler):
+    """Handle tool execution errors with custom messages."""
+    try:
+        return handler(request)
+    except Exception as e:
```
eyurtsev, line 605: This should have a comment saying something conceptually about what kind of errors need to be retried by an LLM.


There are at least 4 types of errors:
1. network failures (should be handeled using tool retry middleware)
2. incorrect tool implementation (should bubble up)
3. llm invocation errors that arise from a schema mismatch -- already auto-handeled!
4. llm invocation errors that happen somewhere inside the function (e.g., `def execute_sql(sql: str) -> str` was invoked with sql="hello world". `sql` is a string (so won't raise an exception from schema mismatch, but will raise an error later on due to the sql being invalid sql! <-- This is the only type of error that should be handled this way

eyurtsev:
Thank you @christian-bromann

PR #1628, opened 2025-11-25
Agent server openapi spec

hinthornw:
Right now we redirect to another git site and use `scalar` to host the api refs. 

Scalar's docs are nicer, but Mintlify doesn't seem to support hosting static html (lol). 
They do support some meh api reference doc generation from an openapi spec though. 
This PR uses that.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -58,3 +58,6 @@ jobs:
 
       - name: Validate documentation links
         run: make broken-links
+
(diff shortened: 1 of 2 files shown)


hinthornw:
Good call-outs. I'll also do a final pass on this before marking as ready to review

lnhsingh (approved):
We'll follow up to de-dupe reference docs from the Deployment section and fix any broken links (if any upon merge)

PR #1639, opened 2025-11-27
updates to changelog/releases structure

mdrxy:
## Overview
Use unified changelog across LangChain/LangGraph

## Type of change

Update existing documentation

## Additional notes

A slight problem with this approach is that clicking "changelog" from the LangGraph tab brings you to the LangChain tab

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -145,7 +145,6 @@
                   "tab": "LangChain",
                   "pages": [
                     "oss/python/langchain/overview",
-                    "oss/python/releases/changelog",
                     {
                       "group": "Get started",
                       "pages": [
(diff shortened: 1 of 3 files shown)


src/docs.json
```diff
@@ -208,13 +208,7 @@
...
                   "pages": [
                     "oss/python/langgraph/overview",
-                    {
-                      "group": "LangGraph v1.0",
-                      "pages": [
-                        "oss/python/releases/langgraph-v1",
-                        "oss/python/migrate/langgraph-v1"
-                      ]
-                    },
+                    "oss/python/releases/changelog",
```
lnhsingh: IMO, the changelog should not be on the individual LangChain vs LangGraph tabs unless they are completely different pages.

We should use the [filter feature](https://www.mintlify.com/docs/create/changelogs#tag-filters) and from each of the release pages, just link to the filtered version of the general changelog.

PR #1715, opened 2025-12-02
feat(langgraph+lsd): custom encryption at rest

cwlbraa:
## Overview

This PR documents new SDK annotations for injecting custom encryption at rest handlers in langsmith deployments.

## Type of change

**Type:** New documentation page
## Related issues/PRs
https://github.com/langchain-ai/langgraph/pull/6482
https://github.com/langchain-ai/langgraph-api/pull/1773

- Linear issue: [LSD-172](https://linear.app/langchain/issue/LSD-172/add-api-for-user-specific-run-and-storeitem-encryption)
- Slack thread:

## Checklist

  - AWS example was tested against localstack
  - both were tested without a real langgraph-server but are similar enough to the custom_encryption.py test graph that we should be safe.  

### DIFF
--- a/src/langsmith/encryption.mdx
+++ b/src/langsmith/encryption.mdx
@@ -33,11 +33,11 @@ Basic encryption only encrypts checkpoint blobs. Metadata fields remain unencryp
 ## Custom encryption
 
 <Note>
-Requires Agent Server version 0.6.9+ and Python SDK version `langgraph-sdk>=0.3.1`.
+Requires Agent Server version 0.6.22+ and Python SDK version `langgraph-sdk>=0.3.1`.
 </Note>
 
 <Warning>
-Agent Server versions 0.5.34–0.6.8 included a pre-release version of custom encryption. Data encrypted with these versions will be corrupted when upgrading to 0.6.9+. Do not use custom encryption on these versions.
+Agent Server versions 0.5.34–0.6.21 included a pre-release version of custom encryption. Data encrypted with these versions will be corrupted when upgrading to 0.6.22+. Do not use custom encryption on these versions.
 </Warning>
 
 Use custom encryption when you need:


src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+# Create client with encryption context header
+client = get_client(
+    url="http://localhost:2024",
+    headers={"X-Encryption-Context": encoded_context}
+)
+
+# Create a thread with encrypted metadata
+thread = await client.threads.create(
+    metadata={
+        "owner": "user-456",  # Unencrypted (for search)
```
hinthornw: I think `owner` would actually typically be provided via auth middleware so it isn't spoofed by a client fwiw

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+    return response['Plaintext']
+```
+
+#### JSON encryption (metadata fields)
+
+Use JSON handlers to selectively encrypt metadata fields. Unlike blob encryption, JSON encryption lets you choose which fields to encrypt—keeping some fields unencrypted for search and filtering while protecting sensitive data like conversation history:
+
+```python
+# Encrypt by field name or prefix
+SENSITIVE_FIELDS = {"messages"}
```
hinthornw: WDYT of flipping this and having "PLAINTEXT_FIELDS" or something and encrypting everything else (?)

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+1. **Define handlers** - Create encryption and decryption functions decorated with `@encryption.encrypt.blob`, `@encryption.decrypt.blob`, `@encryption.encrypt.json`, and `@encryption.decrypt.json`
+2. **Configure** - Add the path to your encryption module in `langgraph.json`
+3. **Pass context** - Send encryption context (like tenant ID and key ID) via the `X-Encryption-Context` header
+4. **Automatic encryption** - LangGraph automatically encrypts data before storing and decrypts on retrieval
+
+## Add custom encryption to your deployment
+
+### 1. Create an encryption module
+
+Create a Python file (e.g., `encrypt.py`) with your encryption handlers. You can implement **blob handlers**, **JSON handlers**, or both depending on your needs:
```
hinthornw: Maybe pedantic, but thoughts on including a line saying "Below is an example using AWS's `kms` service, but you can modify to use any similar key management service"

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+# Encode as base64
+encoded_context = base64.b64encode(
+    json.dumps(encryption_context).encode()
+).decode()
+
+# Create remote graph with encryption context
+remote_graph = RemoteGraph(
+    "agent",
+    url="http://localhost:2024",
+    headers={"X-Encryption-Context": encoded_context}
```
hinthornw: I think we'd put this in the `ainvoke` call rather than statically adding to the graph

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+Instead of passing the `X-Encryption-Context` header, you can derive encryption context from the authenticated user using `@encryption.context`:
+
+```python
+@encryption.context
+async def get_encryption_context(user: BaseUser, ctx: EncryptionContext) -> dict:
+    """Called once per request after auth. Returns the new ctx.metadata."""
+    return {
+        **ctx.metadata,  # Preserve any header context
+        "tenant_id": getattr(user, "tenant_id", "default"),
```
hinthornw: Could you do `user["tenant_id"]`? or na

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,407 @@
...
+            ).decode()
+        else:
+            # Keep fields like "owner" or "status" unencrypted for search/filtering
+            encrypted_data[key] = value
+
+    return encrypted_data
+
+
+@encryption.decrypt.json
+async def decrypt_json_data(ctx: EncryptionContext, data: dict) -> dict:
```
hinthornw: For a given object (say, a `Thread`), what would `data` be? Below is a thread object

```
class Thread(TypedDict):
    thread_id: UUID
    """The ID of the thread."""
    created_at: datetime
    """The time the thread was created."""
    updated_at: datetime
    """The last time the thread was updated."""
    metadata: Fragment
    """The thread metadata."""
    config: Fragment
    """The thread config."""
    context: Fragment
    """The thread context."""
    status: ThreadStatus
    """The status of the thread. One of 'idle', 'busy', 'interrupted', "error"."""
    values: Mapping
    """The current state of the thread."""
    interrupts: dict[str, list[Interrupt]]
    """The current interrupts of the thread, a map of task_id to list of interrupts."""
```

Would the method be called once for each of values, config, context? Or would it be once for the dict? 

I have two concerns:
a) `dict` has no typing information so I'm not sure the shape fo the object I'm expecting based on this doc and
b) reliability/perf:  if the kms client is making an API call for each encrypt/decrypt request, it feels fairly easy to  hit their rate limits of 10k requests per second. And you'd at least  get degraded perf from the sequential roundtrips.

It seems like we could instead recommend fetching a key for a given encryption context and then using that to encrypt/decrypt locally. Could also use an in-mem cache  if a given context is being used repeatedly (?)

Feel free to push back here

cwlbraa: a. data will be `config | context | metadata` here. it's annoying to wrangle. i would not be surprised if folks need the ability to customize based on column, like `@encryption.encrypt.thread.context`, but I think that's best added later at user request - there's likely complexity around the interrelations between columns that we'd have to work out. plus i think this is already possible using EncryptionContext.field, just a little awkward.

b. yes. 

> It seems like we could instead recommend fetching a key for a given encryption context and then using that to encrypt/decrypt locally. Could also use an in-mem cache if a given context is being used repeatedly (?)

let's discuss this sync, i think you're presenting valid trees but we're gonna miss the forest if i keep trying to address them one by one

src/langsmith/custom-encryption.mdx
cwlbraa: reminds me i should probably test this... i had confidence in earlier examples because it was directly cribbed from the test app, this is pure unvalidated AI slop

src/langsmith/custom-encryption.mdx
cwlbraa: hesitant, let's discuss sync

src/langsmith/custom-encryption.mdx
cwlbraa: true, i kinda wanna remove all the owner mentions - i just threw em in there because there's already an index on this field.

src/langsmith/custom-encryption.mdx
cwlbraa: basically says that on line 21, don't wanna repeat too much

src/langsmith/custom-encryption.mdx
cwlbraa: slimmed down this section massively, basically just deleted this

src/langsmith/custom-encryption.mdx
cwlbraa: fixed

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
+---
+title: Add custom at-rest encryption
+sidebarTitle: Custom encryption
+---
+
+Custom encryption lets you control how sensitive data is encrypted before it's stored in the database.
+
+<Note>
+Custom encryption is available for **self-hosted** LangGraph Platform deployments only (Python graphs only).
```
jdrogers940: Is the self-hosted piece here true? Why wouldn't this work for Cloud deployments as well?

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
+---
+title: Add custom at-rest encryption
```
jdrogers940: A bit of a scope creep, but can we change name:

```suggestion
title: Add encryption at-rest
```

We're missing docs on setting the `LANGGRAPH_AES_KEY` today (which has a couple steps, i.e. 1) set environment variable, 2) add pycryptodome to dependencies). Would be good to call out explicitly what this actually encrypts with this as well.

Then sections below become:
1. How to choose which encryption method
2. Setting basic encryption
3. Setting custom encryption
With step by step instructions and caveats of what you can do with each.

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
...
+  },
+  "encryption": {
+    "path": "./encryption.py:encryption"
+  }
+}
+```
+
+## Blob encryption (checkpoints)
+
+Blob handlers encrypt checkpoint data—the serialized state from graph execution. Here's a simplified example using per-tenant keys:
```
jdrogers940: Thoughts on including a brief explanation of what Fernet is and maybe a link for the example below? Don't want it to focus on that too much, but also kind of confusing to see this library imported without the context that it's just a default python crypto thing for a local example.

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
...
+- `run.metadata`, `run.kwargs`
+- `cron.metadata`, `cron.payload`
+
+**Blob handlers** (`@encryption.encrypt.blob` / `@encryption.decrypt.blob`):
+
+- Checkpoint blobs (graph execution state)
+
+## Passing encryption context
+
+Pass encryption context via the `X-Encryption-Context` header. The context is available in your handlers as `ctx.metadata` and is stored alongside encrypted data for use during decryption.
```
jdrogers940: Maybe just clarify even more that this is arbitrary data. The user defines the schema and has it available during encryption and decryption.

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
...
+
+
+@encryption.decrypt.blob
+async def decrypt_blob(ctx: EncryptionContext, data: bytes) -> bytes:
+    return _get_fernet(ctx).decrypt(data)
+```
+
+The `ctx.metadata` dict comes from the `X-Encryption-Context` header and is stored alongside encrypted data, so the correct key is used on decryption.
+
+For production deployments with key rotation and audit logging, see [Envelope encryption with AWS Encryption SDK](#envelope-encryption-with-aws-encryption-sdk).
```
jdrogers940: I feel like I'd remove these two comments and in the above `How it works` section just link to all these more detailed sections. Maybe also glob them together into categories based on the overarching theme? Like a section Defining Encryption Module that then has these subsections for blob encryption, json encryption, etc.

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
...
+async def get_encryption_context(user: BaseUser, ctx: EncryptionContext) -> dict:
+    return {
+        **ctx.metadata,
+        "tenant_id": user["tenant_id"],
+    }
+```
+
+This handler runs once per request after authentication. The returned dict becomes `ctx.metadata` for all encryption operations in that request.
+
+## Model-specific handlers
```
jdrogers940: I think I'd move this up just below the other sections on configuring the encryption module.

src/langsmith/custom-encryption.mdx
```diff
@@ -0,0 +1,352 @@
...
+    return decrypt_encrypted_fields(data)
+
+@encryption.decrypt.json.assistant
+async def decrypt_assistant(ctx: EncryptionContext, data: dict) -> dict:
+    return decrypt_encrypted_fields(data)
+```
+
+Supported model types: `thread`, `assistant`, `run`, `cron`, `checkpoint`.
+
+## Envelope encryption with AWS Encryption SDK
```
jdrogers940: Not blocking - wonder if we can add some more examples here for Azure, GCP, etc.

jdrogers940 (commented):
Nice! These docs look really good. Proposed some high level restructuring but content and examples themselves look awesome.

src/langsmith/custom-encryption.mdx
cwlbraa: good call, i suppose it fits well here!

src/langsmith/custom-encryption.mdx
cwlbraa: yeah true, i don't see why it wouldn't

src/langsmith/custom-encryption.mdx
cwlbraa: yeah sure

src/langsmith/custom-encryption.mdx
cwlbraa: yeah, agreed

src/langsmith/custom-encryption.mdx
cwlbraa: i haven't used their SDK emulators, localstack made testing this relatively easy.

i also kinda expect the examples on other IaaSes will be larger and more touchy - i doubt azure and gcp have equivalents to aws-encryption-sdk.

cwlbraa: anyways this is a classic "wait for somebody to ask or PR" situation

src/langsmith/encryption.mdx
```diff
@@ -0,0 +1,377 @@
+---
+title: Add encryption at-rest
+sidebarTitle: Encryption at-rest
+---
+
+LangGraph Platform supports encryption at-rest for checkpoint data and metadata. You can choose between basic encryption with a single key or custom encryption for advanced use cases.
```
jdrogers940:
```suggestion
Agent Server supports encryption at-rest for checkpoint data and metadata. You can choose between basic encryption with a single key or custom encryption for advanced use cases.
```

src/langsmith/encryption.mdx
```diff
@@ -0,0 +1,377 @@
...
+
+<Note>
+Encryption at-rest is available for LangGraph Platform deployments (Python graphs only).
```
jdrogers940: nit: This is in the Deployments section so I would just call out the python piece. Also the API version needed similar to here: https://docs.langchain.com/langsmith/custom-lifespan#how-to-add-custom-lifespan-events

src/langsmith/encryption.mdx
```diff
@@ -0,0 +1,377 @@
...
+</Note>
+
+## Choosing an encryption method
+
+| Method | What's encrypted | Use case |
+|--------|------------------|----------|
+| **Basic encryption** | Checkpoint blobs only | Single static key, automatic AES encryption |
+| **Custom encryption** | Checkpoints + JSON fields | Per-tenant keys, KMS integration, selective field encryption |
```
jdrogers940:
```suggestion
| **Custom encryption** | Checkpoints, threads, runs, assistants and crons | Per-tenant keys, KMS integration, selective field encryption |
```
Or something else to show these resources. JSON fields feels too vague.

src/langsmith/encryption.mdx
```diff
@@ -0,0 +1,381 @@
+---
+title: Add encryption at-rest
+sidebarTitle: Encryption at-rest
+---
+
+LangGraph Platform supports encryption at-rest for checkpoint data and metadata. You can choose between basic encryption with a single key or custom encryption for advanced use cases.
+
+<Note>
+Encryption at-rest is available for LangGraph Platform deployments (Python graphs only).
```
katmayb:
```suggestion
Encryption at-rest is available for Agent Server deployments (Python graphs only).
```

Or LangSmith deployments?

cwlbraa: willfix

src/langsmith/encryption.mdx
cwlbraa: the "metadata" ambiguity is brutal. i think naming the models, as you've suggested, instead of naming the fields is probably best here, but it's also annoyingly vague.

PR #1723, opened 2025-12-03
refs(py): add langchain-azure-postgresql

mdrxy:


mdrxy:
FYI @marlenezw the docstrings in the pkg will need to eventually undergo a similar treatment to https://github.com/langchain-ai/langchain-azure/issues/171

PR #1759, opened 2025-12-05
Add information about setup for persistence

rahul-langchain:
## Overview
We currently say you have to call `setup` for a saver or store but don't specify what that does or best practices for when to call it.

## Type of change

## Checklist

### DIFF
--- a/src/oss/langgraph/add-memory.mdx
+++ b/src/oss/langgraph/add-memory.mdx
@@ -2146,7 +2146,7 @@ await checkpointer.deleteThread(threadId);
 
 If you are using any database-backed persistence implementation (such as Postgres or Redis) to store short and/or long-term memory, you will need to run migrations to set up the required schema before you can use it with your database.
 
-The convention is to define a `setup()` method on the checkpointer or store instance to run the required migrations. However, you should check with your specific implementation of @[`BaseCheckpointSaver`] or @[`BaseStore`] to confirm the exact method name and usage.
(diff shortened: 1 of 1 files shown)


src/oss/langgraph/persistence.mdx
```diff
@@ -9,8 +9,8 @@ LangGraph has a built-in persistence layer, implemented through checkpointers. W
 ![Checkpoints](/oss/images/checkpoints.jpg)
 
 <Info>
-**LangGraph API handles checkpointing automatically**
-When using the LangGraph API, you don't need to implement or configure checkpointers manually. The API handles all persistence infrastructure for you behind the scenes.
+**Agent Server handles checkpointing automatically**
```
rahul-langchain, line 12: Unrelated but noticed while I was doing this so I decided to fix as well!

src/oss/langgraph/add-memory.mdx
```diff
@@ -2141,3 +2141,12 @@ await checkpointer.deleteThread(threadId);
 
 **LangMem** is a LangChain-maintained library that offers tools for managing long-term memories in your agent. See the [LangMem documentation](https://langchain-ai.github.io/langmem/) for usage examples.
 :::
+
+## Database management
+
+If you are using Postgres or Redis to store short and/or long-term memory, you will need to call the LangGraph-provided `saver` or `store`'s `setup` method before you can use it with your database.
```
hinthornw: This applies to any checkpointer or store really - any DB-backed persistence implementation would have migrations.

The convention is to define a setup() method to run the required migrations for that instance, but the user should check with their specific implementation of the BaseCheckpointSaver or BaseStore to confirm.

We'd recommend running migration as a dedicated deployment step. Or you can ensure they're run as a part of server startup.

PR #1781, opened 2025-12-08
Add link to SCIM from Self-host SSO page

katmayb:
Fixes DOC-518


katmayb:
@bvs-langchain Is this where you were thinking for the link?

src/langsmith/self-host-sso.mdx
bvs-langchain, line 160: I would link here as well

PR #1811, opened 2025-12-09
feat: add example of parsing structured content in an interceptor

sydney-runkle:
adding structured content parsing example

### DIFF
--- a/src/oss/langchain/mcp.mdx
+++ b/src/oss/langchain/mcp.mdx
@@ -592,7 +592,7 @@ for message in result["messages"]:
 
 **Appending structured content via interceptor**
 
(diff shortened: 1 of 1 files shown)


src/oss/langchain/mcp.mdx
```diff
@@ -590,6 +590,29 @@ for message in result["messages"]:
         structured_content = message.artifact["structured_content"]
 ```
 
+**Appending structured content via interceptor**
+
+You can use an [interceptor](#tool-interceptors) to automatically append structured content to the tool message, making it visible to the model:
```
sydney-runkle:
```suggestion
If you want structured content to be visible in the conversation history (visible to the model), you can use an [interceptor](#tool-interceptors) to automatically append structured content to the tool result:
```

lnhsingh (approved):

PR #1855, opened 2025-12-11
feat: multi agent docs

sydney-runkle:
* adding progressive disclosure via skills (from @eyurtsev)
* adding graph based handoffs (sequential or router based)
* elevating context eng section to the top
* consolidating some verbose "when to use" sections
* using some mintlify features for more interactive / engaging page :)

still todo
* adding "how to choose" section
* decide on wording for handoffs options


src/oss/langchain/multi-agent.mdx
```diff
@@ -298,13 +299,21 @@ const callSubagent1 = tool(
...
 ## Handoffs
 
-In **handoffs**, agents can directly pass control to each other. The “active” agent changes, and the user interacts with whichever agent currently has control.
+In **handoffs**, agents pass control to each other. There are two main approaches:
 
-Flow:
+| Approach | How it works | Best for |
+|----------|--------------|----------|
+| [**Tool handoffs**](#tool-handoffs) | Agent calls a tool that updates state to transfer control | Dynamic routing, user-driven flows |
+| [**Graph handoffs**](#graph-handoffs) | Graph edges define transitions between agents | Fixed pipelines, deterministic routing |
```
eyurtsev, line 307: langgraph is great for things that are dynamic and include cycles and bespoke logic, so "fixed pipelines, determinsitic routing" isn't really correct?

src/oss/langchain/multi-agent.mdx
```diff
@@ -313,5 +322,472 @@ graph LR
...
+
+    A content pipeline where each agent processes and passes to the next:
+
+    :::python
+    ```python
+    from langgraph.graph import StateGraph, MessagesState, START
+    from langchain.agents import create_agent
+
+    # Create specialized agents for each stage
+    research_agent = create_agent(
```
eyurtsev, line 445: This example won't work I think? it'll result in AIMessage being fed into the editor agent?

---

A fixed pipeline could probably be achieved with middleware + force tool choice?

src/oss/langchain/multi-agent.mdx
eyurtsev, line 307: One of the points of handoffs was that it allows users to maintain conversation with the active agent. Graph handoffs definitely do not though that, so we should find a different way of presenting it and likely not grouping together.

----

We probably want to introduce an idea of a **router** that routes the user's input into the agent that's carrying the conversation (if there are multiple agents) or suggest an architecture where there's a single agent that's doing all the hard work (via skills)

src/oss/langchain/multi-agent.mdx
```diff
@@ -7,85 +7,85 @@ Instead of relying on a single agent to handle every step, **multi-agent archite
...
-|-------------------------------------------------------|--------------|----------|
-| Need centralized control over workflow?               | ✅ Yes        | ❌ No     |
-| Want agents to interact directly with the user?       | ❌ No         | ✅ Yes    |
-| Complex, human-like conversation between specialists? | ❌ Limited    | ✅ Strong |
+There are three main multi-agent patterns:
+
+| Pattern | How it works | Example use case |
+|---------|--------------|------------------|
+| [**Tool calling**](#tool-calling) | A supervisor agent calls other agents as tools. **Centralized** control — all routing passes through the supervisor. | Task orchestration, parallel execution |
+| [**Handoffs**](#handoffs) | Agents transfer control to each other via tool calls or graph edges. **Decentralized** — the active agent changes dynamically. | Customer support, multi-domain conversations |
```
eyurtsev, line 32: The classifier pattern that runs on every user input is very common and not captured in this classification

src/oss/langchain/multi-agent.mdx
```diff
@@ -7,85 +7,85 @@ Instead of relying on a single agent to handle every step, **multi-agent archite
...
+| [**Skills**](#skills) | Specialized prompts loaded **on-demand**. The main agent stays in control and gains additional context as needed. | Progressive disclosure, large knowledge bases |
```
eyurtsev, line 33: Progressive disclosure is not a use case -- it's a technique for context management.

src/oss/langchain/multi-agent.mdx
```diff
@@ -313,5 +322,472 @@ graph LR
...
+    B --> C[Python skill]
+    B --> D[SQL skill]
+    B --> E[Docs skill]
+    C --> B
+    D --> B
+    E --> B
+    B --> A
+```
+
+Use skills when you have a **single agent with many possible specializations** and need to manage large amounts of domain knowledge (e.g., database schemas, API docs) without overwhelming context. Skills are lighter than full sub-agents and can be developed independently by different teams.
```
eyurtsev, line 703: **single agent with many possible specializations** -> this inverts the logic?

Users need to decide what they need to build? Should they build a single agent and use skills or should they use different agents? Why should they use different agents?

src/oss/langchain/multi-agent.mdx
```diff
@@ -313,5 +322,472 @@ graph LR
...
+    Available skills: python, sql, docs
+    """
+    if skill_name not in SKILLS:
+        return f"Unknown skill: {skill_name}. Available: {list(SKILLS.keys())}"
+    return f"Skill loaded. Additional context:\n\n{SKILLS[skill_name]}"
+
+agent = create_agent(
+    model="openai:gpt-4o",
+    tools=[load_skill, ...],  # Other tools the agent needs
+    prompt="You are a helpful assistant. Use load_skill to gain expertise when needed."
```
eyurtsev: This doesn't work well since the skills aren't included in the system prompt? i.e., the agent doesn't know which skills it can load

Perhaps we link into the tutorial for now?

PR #1866, opened 2025-12-11
feat: multi agent part 2

sydney-runkle:
New and revamped multi agent docs!! Weeks in the making!

Added pages for:
* Multi agent overview
  * Subagents
  * Router
  * Handoffs
  * Skills
  * Custom workflow

Added tutorials for:
* Router
* Handoffs
* Skills

### DIFF
--- a/src/oss/langchain/multi-agent/index.mdx
+++ b/src/oss/langchain/multi-agent/index.mdx
@@ -236,79 +236,118 @@ Each language agent/skill contains ~2000 tokens of documentation. All patterns c
   <Tab title="Subagents">
     **5 calls, ~9K tokens**
 
-    ```
-    Call 1: Main agent (1K tokens)
-      ├─ Calls 3 subagents in parallel
-    Call 2: Python subagent (2K tokens) ─┐
-    Call 3: JavaScript subagent (2K tokens) ├─ Parallel
-    Call 4: Rust subagent (2K tokens) ─────┘
-    Call 5: Main synthesizes (2K tokens)
+    ```mermaid
+    sequenceDiagram
+        participant User
+        participant Main Agent
+        participant Python Subagent
+        participant JS Subagent
+        participant Rust Subagent
 
-    Total: 1K + 2K + 2K + 2K + 2K = 9K tokens
+        User->>Main Agent: "Compare Python, JS, Rust"
+        Note over Main Agent: Call 1 (1K)
+        par Parallel execution
+            Main Agent->>Python Subagent: python_subagent()
+            Note over Python Subagent: Call 2 (2K)
+            Python Subagent-->>Main Agent: Python analysis
+        and
+            Main Agent->>JS Subagent: js_subagent()
+            Note over JS Subagent: Call 3 (2K)
+            JS Subagent-->>Main Agent: JS analysis
+        and
+            Main Agent->>Rust Subagent: rust_subagent()
+            Note over Rust Subagent: Call 4 (2K)
+            Rust Subagent-->>Main Agent: Rust analysis
+        end
+        Note over Main Agent: Call 5 (2K)
+        Main Agent-->>User: Synthesized comparison
     ```
 
-    Each subagent works in **isolation** with only its relevant context.
+    Each subagent works in **isolation** with only its relevant context. Total: **9K tokens**.
   </Tab>
 
   <Tab title="Handoffs">
     **7+ calls, ~14K+ tokens**
 
-    ```
-    Call 1: Main agent handoff to Python (1K)
-    Call 2-3: Python agent researches (2-3 calls, ~2K each)
-    Call 4: Handoff to JavaScript agent (included in Python's response)
-    Call 5-6: JavaScript agent researches (2-3 calls, ~2K each)
-    Call 7: Handoff to Rust agent (included in JS's response)
-    Call 8-9: Rust agent researches (2-3 calls, ~2K each)
+    ```mermaid
+    sequenceDiagram
+        participant User
+        participant Main Agent
+        participant Python Agent
+        participant JS Agent
+        participant Rust Agent
 
-    Total: ~14K+ tokens across sequential handoffs
+        User->>Main Agent: "Compare Python, JS, Rust"
+        Note over Main Agent: Call 1 (1K)
+        Main Agent->>Python Agent: transfer_to_python()
+        Note over Python Agent: Calls 2-3 (2K each)
+        Python Agent->>JS Agent: transfer_to_js()
+        Note over JS Agent: Calls 4-5 (2K each)
+        JS Agent->>Rust Agent: transfer_to_rust()
+        Note over Rust Agent: Calls 6-7 (2K each)
+        Rust Agent-->>User: Combined comparison
     ```
 
-    Handoffs executes **sequentially**—can't research all three languages in parallel. Growing conversation history adds overhead.
+    Handoffs executes **sequentially**—can't research all three languages in parallel. Growing conversation history adds overhead. Total: **~14K+ tokens**.
   </Tab>
 
   <Tab title="Skills">
     **3 calls, ~15K tokens**
 
+    ```mermaid
+    sequenceDiagram
+        participant User
+        participant Agent
+        participant load_skill tool
+
+        User->>Agent: "Compare Python, JS, Rust"
+        Note over Agent: Call 1 (1K)
+        Agent->>load_skill tool: load_skill("python", "js", "rust")
+        load_skill tool-->>Agent: +6K context (2K each)
+        Note over Agent: Call 2 (7K: 1K base + 6K skills)
+        Note over Agent: Call 3 (7K: 1K base + 6K skills)
+        Agent-->>User: Synthesized comparison
     ```
-    Call 1: Load 3 skills (1K tokens)
-      └─ Adds Python (2K) + JavaScript (2K) + Rust (2K) = 6K to context
 
-    Call 2: Research (7K tokens)
-      └─ Base (1K) + ALL skill contexts (6K) = 7K total
-
-    Call 3: Synthesize (7K tokens)
-      └─ Base (1K) + ALL skill contexts (6K) = 7K total
-
-    Total: 1K + 7K + 7K = 15K tokens
-    ```
-
-    After loading, **every subsequent call processes all 6K tokens of skill documentation**. Subagents processes 67% fewer tokens overall due to context isolation.
+    After loading, **every subsequent call processes all 6K tokens of skill documentation**. Subagents processes 67% fewer tokens overall due to context isolation. Total: **15K tokens**.
   </Tab>
 
   <Tab title="Router">
     **5 calls, ~9K tokens**
 
-    ```
-    Call 1: Router LLM analyzes query (1K tokens)
-      ├─ Routes to Python, JavaScript, Rust agents
-    Call 2: Python agent (2K tokens) ─┐
-    Call 3: JavaScript agent (2K tokens) ├─ Parallel
-    Call 4: Rust agent (2K tokens) ─────┘
-    Call 5: Synthesis LLM combines results (2K tokens)
+    ```mermaid
+    sequenceDiagram
+        participant User
+        participant Router LLM
+        participant Python Agent
+        participant JS Agent
+        participant Rust Agent
+        participant Synthesis LLM
 
-    Total: 1K + 2K + 2K + 2K + 2K = 9K tokens
+        User->>Router LLM: "Compare Python, JS, Rust"
+        Note over Router LLM: Call 1 (1K)
+        par Parallel execution
+            Router LLM->>Python Agent: Route to Python
+            Note over Python Agent: Call 2 (2K)
+            Python Agent-->>Synthesis LLM: Python analysis
+        and
+            Router LLM->>JS Agent: Route to JS
+            Note over JS Agent: Call 3 (2K)
+            JS Agent-->>Synthesis LLM: JS analysis
+        and
+            Router LLM->>Rust Agent: Route to Rust
+            Note over Rust Agent: Call 4 (2K)
+            Rust Agent-->>Synthesis LLM: Rust analysis
+        end
+        Note over Synthesis LLM: Call 5 (2K)
+        Synthesis LLM-->>User: Combined comparison
     ```
 
-    Router uses an **LLM for routing**, then invokes agents in parallel. Similar to Subagents but with explicit routing step.
+    Router uses an **LLM for routing**, then invokes agents in parallel. Similar to Subagents but with explicit routing step. Total: **9K tokens**.
   </Tab>
 </Tabs>
 
 **Key insight:** For multi-domain tasks, patterns with parallel execution (Subagents, Router) are most efficient. Skills has fewer calls but high token usage due to context accumulation. Handoffs is inefficient here—it must execute sequentially and can't leverage parallel tool calling for consulting multiple domains simultaneously.
-
-<Warning>
-**When to avoid Skills**: The Skills pattern is ideal for a small number of lightweight skills. When you need many skills with extensive documentation (API references, detailed examples, comprehensive guidelines), use **Subagents** or **Router** instead. Context isolation prevents repeatedly processing accumulated documentation.
-</Warning>
 
 ### Summary
 


src/oss/langchain/multi-agent.mdx
```diff
@@ -2,90 +2,89 @@
...
-    * Adding human-in-the-loop review for sensitive actions
-</Card>
+When developers say they need "multi-agent," they're usually looking for one or more of these capabilities:
 
+* **Context management**: Provide specialized knowledge without overwhelming the model's context window. If context were infinite and latency zero, you could dump all knowledge into a single prompt — but since it's not, you need patterns to selectively surface relevant information.
+* **Distributed development**: Allow different teams to develop and maintain capabilities independently, composing them into a larger system with clear boundaries and independent testing.
+* **Parallelization**: Spawn specialized workers for subtasks and execute them concurrently.
 
-## Choosing a pattern
+## When to use
```
eyurtsev: Maybe we can kill `#when to use` and add the sequential constraints to the multiagent above?

and maybe incorporate this part if we feel it's not obvious?: ``` A single agent has too many [tools](/oss/langchain/tools) and makes poor decisions about which to use.
* [Context](/oss/concepts/context) or [memory](/oss/langchain/short-term-memory) grows too large for one agent to track effectively.```

src/oss/langchain/multi-agent.mdx
```diff
@@ -296,22 +296,482 @@ const callSubagent1 = tool(
 :::
 
 
-## Handoffs
+## Dynamic agent
+
+In the **dynamic agent** architecture, you have a single agent whose behavior (prompt, tools, and capabilities) changes dynamically based on state. Under the hood, this is a state machine implemented via [middleware](/oss/langchain/middleware/overview) that performs handoffs through tool calls.
```
eyurtsev: We need to choose terminology here but perhaps we can avoid:

* state machine
* handoffs

---

We could just say that it changes its behaviors based on previous actions (tool calls) it has taken?

---

So instead of having to represent an IT triage issue using 3 different agents, you can just use 1. (Do you really need 3 different people in real life to debug a 2-step issue?)

src/oss/langchain/multi-agent.mdx
```diff
@@ -296,22 +296,482 @@ const callSubagent1 = tool(
...
+
+    const result = await workflow.invoke({ question: "I was charged twice for my subscription last month" });
+    console.log(result.answer);
+    ```
+    :::
+  </Tab>
+</Tabs>
+
+
+## Skills
```
eyurtsev: 1. Could we say that we do not offer a prebuilt middleware for skills yet, but users can implement custom logic and can reference the tutorial?
2. Could we steal a section from the tutorial about enhancing them? i.e., i think it's worth while explaining that with custom implementations one could have loading a skill also load tools

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
+Use this table to match your requirements to the right pattern:
+
+| Pattern | Distributed development | Parallelization | Multi-hop | Direct user interaction |
+|---------|:-----------------------:|:---------------:|:----------:|:-----------------------:|
+| [**Subagents**](#subagents) | ⭐⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐ | — |
+| [**Handoffs**](#handoffs) | — | — | ⭐⭐⭐ | ⭐⭐⭐ |
+| [**Skills**](#skills) | ⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐ |
+| [**Router**](#router) | ⭐⭐ | ⭐⭐⭐ | — | ⭐⭐ |
+
+- **Distributed development**: Can different teams maintain components independently?
```
lnhsingh: it would be nice to make these `<Tooltip>`s so you can inline them into the table, but I'm not sure if it'll render

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
 >
-    Learn how to build a personal assistant using the supervisor pattern, where a central supervisor agent coordinates specialized worker agents.
-    This tutorial demonstrates:
-    * Creating specialized sub-agents for different domains (calendar and email)
-    * Wrapping sub-agents as tools for centralized orchestration
-    * Adding human-in-the-loop review for sensitive actions
+    Learn how to build a personal assistant using the subagents pattern, where a central main agent (supervisor) coordinates specialized worker agents.
 </Card>
 
+### Sync vs async
```
lnhsingh:
```suggestion
### Sync vs. async
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
-|-------------------------------------------------------|--------------|----------|
-| Need centralized control over workflow?               | ✅ Yes        | ❌ No     |
-| Want agents to interact directly with the user?       | ❌ No         | ✅ Yes    |
-| Complex, human-like conversation between specialists? | ❌ Limited    | ✅ Strong |
+For **long-running tasks** (reviewing contracts, conducting research, auditing code), use **asynchronous execution**. The main agent kicks off a background job and continues conversing with the user while the work completes.
 
-<Tip>
-    You can mix both patterns — use **handoffs** for agent switching, and have each agent **call subagents as tools** for specialized tasks.
-</Tip>
+<Accordion title="Background job pattern">
```
lnhsingh: It seems weird to hide this when it's not hidden elsewhere

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
+```
+
+**Key characteristics:**
+
+* Single task tool: One parameterized tool that can invoke any registered sub-agent by name
+* Convention-based invocation: Agent selected by name, task passed as human message, final message returned as tool result
+* Team distribution: Different teams can develop and deploy agents independently
+* Agent discovery: Sub-agents can be discovered via system prompt (listing available agents) or through [progressive disclosure](/oss/langchain/progressive-disclosure-skills) (loading agent information on-demand via tools)
+
+Use this approach when you want to distribute agent development across multiple teams, need to isolate complex tasks into separate context windows, need a scalable way to add new agents without modifying the coordinator, or prefer convention over customization. This approach trades flexibility in context engineering for simplicity in agent composition and strong context isolation.
```
lnhsingh: I think this should be towards the top of the section. I think users will be more interested in matching their use case first before reading more into the technical details

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
 :::
 
-In this pattern:
-1. The main agent invokes `call_subagent1` when it decides the task matches the subagent’s description.
-2. The subagent runs independently and returns its result.
-3. The main agent receives the result and continues orchestration.
+</Accordion>
 
-### Where to customize
+### Context engineering
```
lnhsingh: We should add a link on `/langchain/context-engineering` page that links here if you want to read about context eng in multi-agents. maybe in Related resources?

src/oss/langchain/multi-agent.mdx
```diff
@@ -213,16 +432,16 @@ const callSubagent1 = tool(
...
-  * Useful when outputs are incomplete, too verbose, or missing key details.
-  * A common failure mode is that the subagent performs tool calls or reasoning but does **not include the results** in its final message. Remind it that the controller (and user) only see the final output, so all relevant info must be included there.
-* **Custom output formatting** – adjust or enrich the subagent's response in code before handing it back to the main agent.
-  * Example: pass specific state keys back to the main agent in addition to the final text.
-  * This requires wrapping the result in a @[`Command`] (or equivalent structure) so you can merge custom state with the subagent’s response.
+1. **Prompt the sub-agent**: Specify exactly what should be returned. A common failure mode is that the sub-agent performs tool calls or reasoning but doesn't include results in its final message—remind it that the supervisor only sees the final output.
+2. **Format in code**: Adjust or enrich the response before returning it. For example, pass specific state keys back in addition to the final text using a `Command`.
+
+<Accordion title="Subagent outputs">
```
lnhsingh:
```suggestion
<Accordion title="Subagent outputs example">
```

Or could use the `expandable` tag here on the codeblock so it shows some of the code but not all of it? https://www.mintlify.com/docs/create/code#expandable

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
 
-In **handoffs**, agents can directly pass control to each other. The “active” agent changes, and the user interacts with whichever agent currently has control.
-
-Flow:
+In the **handoffs** architecture, behavior changes dynamically based on state. The core mechanism: tools update a state variable (e.g., `current_step` or `active_agent`) that persists across turns, and the system reads this variable to adjust behavior—either applying different configuration (system prompt, tools) or routing to a different agent. This pattern supports both handoffs between distinct agents and dynamic configuration changes within a single agent.
 
-1. The **current agent** decides it needs help from another agent.
-2. It passes control (and state) to the **next agent**.
-3. The **new agent** interacts directly with the user until it decides to hand off again or finish.
+<Tip>
```
lnhsingh: Idk if this is needed

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+**Key characteristics:**
+
+* Prompt-driven specialization: Skills are primarily defined by specialized prompts
+* Progressive disclosure: Skills become available based on context or user needs
+* Team distribution: Different teams can develop and maintain skills independently
+* Lightweight composition: Skills are simpler than full sub-agents
+
+Use the skills pattern when you want a single agent with many possible specializations, you don't need to enforce specific constraints between skills, or different teams need to develop capabilities independently. Common examples include coding assistants (skills for different languages or tasks), knowledge bases (skills for different domains), and creative assistants (skills for different formats).
+
+<Tip>
```
lnhsingh: Also not sure if this is needed. Or at least the parenthetical isnt

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+This pattern is conceptually identical to [llms.txt](https://llmstxt.org/) (introduced by Jeremy Howard), which uses tool calling for progressive disclosure of documentation. The skills pattern applies the same approach to specialized prompts and domain knowledge rather than just documentation pages.
+</Tip>
+
+<Accordion title="Loading skills on-demand">
```
lnhsingh: could also use `expandable` here

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+```
+:::
+
+</Accordion>
+
+### Extending the pattern
+
+When writing custom implementations, you can extend the basic skills pattern in several ways:
+
+**Dynamic tool registration**: Combine progressive disclosure with state management to register new tools as skills load. For example, loading a "database_admin" skill could both add specialized context and register database-specific tools (backup, restore, migrate). This uses the same tool-and-state mechanisms used across multi-agent patterns—tools updating state to dynamically change agent capabilities.
```
lnhsingh:
```suggestion
- **Dynamic tool registration**: Combine progressive disclosure with state management to register new tools as skills load. For example, loading a "database_admin" skill could both add specialized context and register database-specific tools (backup, restore, migrate). This uses the same tool-and-state mechanisms used across multi-agent patterns—tools updating state to dynamically change agent capabilities.
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+
+**Hierarchical skills**: Skills can define other skills in a tree structure, creating nested specializations. For instance, loading a "data_science" skill might make available sub-skills like "pandas_expert", "visualization", and "statistical_analysis". Each sub-skill can be loaded independently as needed, allowing for fine-grained progressive disclosure of domain knowledge. This hierarchical approach helps manage large knowledge bases by organizing capabilities into logical groupings that can be discovered and loaded on-demand.
```
lnhsingh:
```suggestion
- **Hierarchical skills**: Skills can define other skills in a tree structure, creating nested specializations. For instance, loading a "data_science" skill might make available sub-skills like "pandas_expert", "visualization", and "statistical_analysis". Each sub-skill can be loaded independently as needed, allowing for fine-grained progressive disclosure of domain knowledge. This hierarchical approach helps manage large knowledge bases by organizing capabilities into logical groupings that can be discovered and loaded on-demand.
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+Two approaches:
+* [**Stateless routers**](#stateless) address each request independently
+* [**Stateful routers**](#stateful) maintain conversation history across requests
+
+### Stateless
+
+Each request is routed independently—no memory between calls. For multi-turn conversations, see [Stateful routers](#stateful).
+
+<Tip>
+**Stateless router vs Subagents**: The [subagents pattern](#subagents) can also route to multiple agents. Use the stateless router when you need specialized preprocessing or custom routing logic. Use the subagents pattern when you want the LLM to decide which agents to call dynamically.
```
lnhsingh:
```suggestion
**Stateless router vs. Subagents**: The [subagents pattern](#subagents) can also route to multiple agents. Use the stateless router when you need specialized preprocessing or custom routing logic. Use the subagents pattern when you want the LLM to decide which agents to call dynamically.
```

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+    B -->|path_a| C[Deterministic step]
+    B -->|path_b| D((Agentic step))
+    C --> G([Output])
+    D --> G([Output])
+```
+
+The [router pattern](#router) is an example of a custom workflow.
+
+<Tip>
+**Calling a LangChain agent from a LangGraph node**: The main insight when mixing LangChain and LangGraph is that you can call a LangChain agent directly inside any LangGraph node. This lets you combine the flexibility of custom workflows with the convenience of pre-built agents:
```
lnhsingh: I feel like this might deserve to be it's own subsection rather than a callout?

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+  const result = await agent.invoke({
+    messages: [{ role: "user", content: state.query }]
+  });
+  return { answer: result.messages.at(-1)?.content };
+}
+```
+:::
+</Tip>
+
+**Example: RAG pipeline** — A common use case is combining retrieval with an agent. This example builds a WNBA stats assistant that retrieves from a knowledge base and can fetch live news.
```
lnhsingh: yasssssssssss

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
-    * Creating specialized sub-agents for different domains (calendar and email)
-    * Wrapping sub-agents as tools for centralized orchestration
-    * Adding human-in-the-loop review for sensitive actions
+    Learn how to build a personal assistant using the subagents pattern, where a central main agent (supervisor) coordinates specialized worker agents.
 </Card>
 
+### Sync vs. async
 
-## Choosing a pattern
+By default, subagent calls are **synchronous**—the main agent waits for each subagent to complete before continuing. This is simple and works well for most cases.
```
eyurtsev: This will be confusing to readers. we need to clarify that we're not talking about sync/async as in python, but more of a non blocking paradigm where the task runs in the background (usually in a separate process or service)

src/oss/langchain/multi-agent.mdx
sydney-runkle: they don't render unfortunately :( but good idea

src/oss/langchain/multi-agent.mdx
```diff
@@ -2,162 +2,381 @@
...
+### Context engineering
 
-There are several points where you can control how context is passed between the main agent and its subagents:
+Control how context flows between the main agent and its subagents:
 
-1. **Subagent name** (`"subagent1_name"`): This is how the main agent refers to the subagent. Since it influences prompting, choose it carefully.
-2. **Subagent description** (`"subagent1_description"`): This is what the main agent “knows” about the subagent. It directly shapes how the main agent decides when to call it.
-3. **Input to the subagent**: You can customize this input to better shape how the subagent interprets tasks. In the example above, we pass the agent-generated `query` directly.
-4. **Output from the subagent**: This is the **response** passed back to the main agent. You can adjust what is returned to control how the main agent interprets results. In the example above, we return the final message text, but you could return additional state or metadata.
+| Category | Purpose | Impacts |
```
eyurtsev: We're missing context engineering information at the level of the main agent. 

Especially w/ the single dispatch tool, we need to add a bit of information about how context engineering is done to discover the sub-agents.

1. System prompt OK with the caveat that if list of agents is very long, we'll need to truncate probably and we don't want this to change for every user request in the **same** conversation** so we don't invalidate the prompt cahce
2. via a tool to discover tools
3. If list of agents isn't big, can also use enum fields on the dispatch tool to further helper the main agent

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+  {
+    name: "transfer_to_sales",
+    description: "Transfer to the sales agent.",
+    schema: z.object({})
+  }
+);
+```
+:::
+
+<Accordion title="Complete example: Sales and support with handoffs">
```
eyurtsev: This example is probably not exactly correct

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+```python
+from typing import Literal
+from langchain.agents import AgentState, create_agent
+from langchain.tools import tool
+from langgraph.graph import StateGraph, START
+from langgraph.types import Command
+
+# 1. Define state with active_agent tracker
+class MultiAgentState(AgentState):
+    active_agent: str = "sales_agent"  # Track which agent is active
```
eyurtsev: this is not correct (i think it's a typed dict, so not defaults)

src/oss/langchain/multi-agent.mdx
```diff
@@ -295,23 +511,1231 @@ const callSubagent1 = tool(
...
+
+# 2. Create handoff tools
+@tool
+def transfer_to_sales():
```
eyurtsev: We should probably include an update of message history to include the AIMessage and the ToolMessage associated w/ the transfer

src/oss/langchain/multi-agent.mdx
sydney-runkle: I think we decided we wanted to keep this section pretty simple and we don't have a ton of other content here at the moment. Going to leave for now, can add a section if we have more material later 👍

src/oss/langchain/multi-agent.mdx
sydney-runkle: leaving for now, @eyurtsev wanted to call out swarm stuff specifically

src/oss/langchain/multi-agent.mdx
sydney-runkle: same thing, @eyurtsev wanted to callout explicitly

sydney-runkle:
- [x] Rename supervisor to subagents (@sydney-runkle)
- [x] Review language for each pattern for accuracy / inaccuracy
- [x] State machine (owner @Eugene)
   - [x] More explanation about handoffs
   - [x] Explanation about conversation fluidity
   - [x] Expanation about handoffs from a subagent (i.e., jumping to parent) 
   - [x] Relax tip statement
- [x] Skills potentially include a bit of information about llms.txt (as it's the same idea)
- [x] Stateless router (owner @sydney-runkle):
   - [x] Explain how to include conversation history as context (this doesn't mean it's stateful)
- [x] Custom workflow (owner @sydney-runkle)
  - [x] Intro shouldn't suggest it's only for sequential pipelines (explain that branching / loops agentic logic is supported)
  - [x] Refine example as agents typically need tools (otherwise it's just an LLM call)

PR #1943, opened 2025-12-16
update agents streaming guide

ccurme:


src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+Below we include short, end-to-end snippets representing common use-cases for streaming.
+
+### Streaming tool calls
+
+You may want to stream both:
+
+1. Partial JSON as [tool calls](/oss/langchain/models#tool-calling) are generated
+2. The completed, parsed tool calls that are executed
+
+To do this, apply both [`"messages"`](#llm-tokens) and [`"updates"`](#agent-progress) streaming modes. The `"messages"` streaming mode will include [message chunks](/oss/langchain/messages#streaming-and-chunks) from all LLM calls in the agent. The `"updates"` mode will include completed messages with tool calls before they are routed to tools for execution.
```
eyurtsev: Could we avoid relying on updates for this information and show how to aggregate the tool message? There's no guarantee that the message comes from the same source?

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+def _process_message_chunk(token: AIMessage) -> None:
+    if token.text:
+        print(token.text, end="|")
+    if token.tool_call_chunks:
+        print(token.tool_call_chunks)
+    # N.B. all content is available through token.content_blocks
+
+
+def _process_completed_message(source: str, message: AnyMessage) -> None:
+    if source == "model" and isinstance(message, AIMessage) and message.tool_calls:
```
eyurtsev: I don't think readers will know what source="model" is , probably need a bit more context

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+Tool response: [{'type': 'text', 'text': "It's always sunny in Boston!"}]
+The| weather| in| Boston| is| **|sun|ny|**|.|
+```
+
+### Streaming with human-in-the-loop
+
+To handle human-in-the-loop [interrupts](/oss/langchain/human-in-the-loop), we build on the [above example](#streaming-tool-calls):
+
+1. We configure the agent with [human-in-the-loop middleware and a checkpointer](/oss/langchain/human-in-the-loop#configuring-interrupts)
+2. We collect interrupts generated during the `"updates"` stream mode
```
eyurtsev, line 747: Do interrupts surface if we do only messages?

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+from langchain.agents import create_agent
+from langchain.agents.middleware import HumanInTheLoopMiddleware
+from langchain.messages import AIMessage, AnyMessage
+from langgraph.checkpoint.memory import InMemorySaver
+from langgraph.types import Command, Interrupt
+
+
+def get_weather(city: str) -> str:
+    """Get weather for a given city."""
+
```
eyurtsev, line 762: Empty line

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+    "openai:gpt-5.2",
+    tools=[get_weather],
+    middleware=[  # [!code highlight]
+        HumanInTheLoopMiddleware(interrupt_on={"get_weather": True}),  # [!code highlight]
+    ],  # [!code highlight]
+    checkpointer=checkpointer,  # [!code highlight]
+)
+
+
+def _process_message_chunk(token: AIMessage) -> None:
```
eyurtsev: Are we able to remove the _process_* helpers as they act like sinks into print but usually users will not want a sync but to rewrite into a different generator

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,386 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+[{'type': 'edit',
+  'edited_action': {'name': 'get_weather', 'args': {'city': 'Boston, U.K.'}}},
+ {'type': 'approve'}]
+```
+
+We can then resume by passing a [command](/oss/langchain/human-in-the-loop#responding-to-interrupts) into the same streaming loop:
+```python
+interrupts = []
+for stream_mode, data in agent.stream(
+    Command(resume={"decisions": decisions}),
```
eyurtsev: We need to show how to resume using interrupt IDs, i.e. so user code supports responding to multiple interrupts in parallel (e.g. two agents running in parallel each using HIL)

src/oss/langchain/streaming.mdx
ccurme: There are alternatives although I'm not sure any of them is great. Issue is that atm there is no ID available during `stream_mode="messages"` that will work to group streams from individual LLM calls.

`langgraph_checkpoint_ns` is available and will work for sub-agents but not for direct chat model calls in nodes.

Some options:

- Update integrations to attach meaningful, consistent IDs on all chunks
  - To avoid breaking changes, we'd need to make sure the aggregated ID is unchanged, which means making sure we collect the provider's message ID on the first chunk.
  - Can update major integrations in ~1 day, assuming they provide the ID on the first chunk, but will be unreliable generally.
- Use `run_id`, which does appear to identify LLM calls
  - Can likely expose in `stream_mode="messages"` metadata
  - Available via callback system / astream_events
- Apply tags (at model init, or with something like `model.with_config({"tags": [str(uuid.uuid4())]}).invoke(...)`)
  - Band-aid / ugly, points to needed functionality
 
The pro of using `stream_mode="updates"` is that if you're using `create_agent`, it works in typical cases and captures the actual tool call that is sent to tools (let me know if there are failure modes with `create_agent`).

Also, unless we use callback system (e.g., as in `astream_events`) or `stream_mode="updates"`, we're duplicating work aggregating the messages, so ideally we figure out a way to expose the aggregated message-- may be possible to update `stream_mode="messages"` to do this.

src/oss/langchain/streaming.mdx
ccurme: No

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,535 @@ for await (const [streamMode, chunk] of await agent.stream(
 ```
 :::
 
+:::python
+## Common patterns
+
+Below we include short, end-to-end snippets representing common use-cases for streaming.
```
lnhsingh:
```suggestion
Below are examples showing common use cases for streaming.
```

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,535 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+
+1. Partial JSON as [tool calls](/oss/langchain/models#tool-calling) are generated
+2. The completed, parsed tool calls that are executed
+
+Specifying [`stream_mode="messages"`](#llm-tokens) will stream incremental [message chunks](/oss/langchain/messages#streaming-and-chunks) generated by all LLM calls in the agent. To access the completed messages with parsed tool calls:
+
+1. If those messages are tracked in the [state](/oss/langchain/agents#memory) (as in the model node of [`create_agent`](/oss/langchain/agents)), use `stream_mode=["messages", "updates"]` to access completed messages through [state updates](#agent-progress).
+2. If those messages are not tracked in the state, use [custom updates](#custom-updates) or aggregate the chunks during the streaming loop.
+
+Below, we demonstrate (1). We also provide an [end-to-end example of (2)](#streaming-completed-messages-that-are-not-in-agent-state) below.
```
lnhsingh: It would be better to put the code examples under H4s. Then from the numbered list above, you can add a "For an example of this, see x." and link to the section

src/oss/langchain/streaming.mdx
```diff
@@ -509,6 +511,535 @@ for await (const [streamMode, chunk] of await agent.stream(
...
+[{'name': None, 'args': 'city', 'id': None, 'index': 0, 'type': 'tool_call_chunk'}]
+[{'name': None, 'args': '":"', 'id': None, 'index': 0, 'type': 'tool_call_chunk'}]
+[{'name': None, 'args': 'Boston', 'id': None, 'index': 0, 'type': 'tool_call_chunk'}]
+[{'name': None, 'args': '"}', 'id': None, 'index': 0, 'type': 'tool_call_chunk'}]
+Tool calls: [{'name': 'get_weather', 'args': {'city': 'Boston'}, 'id': 'call_D3Orjr89KgsLTZ9hTzYv7Hpf', 'type': 'tool_call'}]
+Tool response: [{'type': 'text', 'text': "It's always sunny in Boston!"}]
+The| weather| in| Boston| is| **|sun|ny|**|.|
+```
+
+<Accordion title="Streaming completed messages that are not in agent state">
```
lnhsingh: Re: above comment -- I think this deserves it's own section/heading

PR #1984, opened 2025-12-19
chore: patch for mermaid diagrams in tabs

sydney-runkle:
kept the mermaid files as a ref, should just be able to revert this PR once mintlify fixes the mermaid rendering issue in tabs

proof that this is working, but slightly uglier: https://langchain-5e9cc07a-preview-srdocs-1766157608-70f551c.mintlify.app/oss/python/langchain/multi-agent#visual-overview


lnhsingh (approved):
we should revert as soon as they're working again

PR #2029, opened 2025-12-26
feat: database iam authentication [INF-1316]

joaquin-borggio-lc:
## Overview
Adding IAM support for database connections (postgres and redis).

## Type of change

## Related issues/PRs

- GitHub issue:
- Feature PR:

- Linear issue:
- Slack thread:

## Checklist

- [ ] I have read the [contributing guidelines](README.md)
- [ ] I have used **root relative** paths for internal links

## Additional notes

Screenshots from local testing. all looked as expected:
Screenshot 2025-12-26 at 12 53 12 PM
Screenshot 2025-12-26 at 12 53 32 PM
Screenshot 2025-12-26 at 12 53 53 PM
Screenshot 2025-12-26 at 12 54 17 PM

### DIFF
--- a/src/langsmith/self-host-external-postgres.mdx
+++ b/src/langsmith/self-host-external-postgres.mdx
@@ -15,6 +15,8 @@ LangSmith uses a PostgreSQL database as the primary data store for transactional
 
 * Note: We only officially support PostgreSQL versions >= 14.
 
+* We support password and [IAM/Workload Identity](#iam-authentication) authentication.
+
 * A user with admin access to the PostgreSQL database. This user will be used to create the necessary tables, indexes, and schemas.
 
 * This user will also need to have the ability to create extensions in the database. We use/will try to install the `btree_gin`, `btree_gist`, `pgcrypto`, `citext`, `ltree`, and `pg_trgm` extensions.
--- a/src/langsmith/self-host-external-redis.mdx
+++ b/src/langsmith/self-host-external-redis.mdx
@@ -15,7 +15,7 @@ LangSmith uses Redis to back our queuing/caching operations. By default, LangSmi
 
 * Note: We only officially support Redis versions >= 5.
 * We support both Standalone and Redis Cluster. See the appropriate sections for deployment instructions.
-
+* We support no authentication, password, and [IAM/Workload Identity](#iam-authentication) authentication.
 * By default, we recommend an instance with at least 2 vCPUs and 8GB of memory. However, the actual requirements will depend on your tracing workload. We recommend monitoring your Redis instance and scaling up as needed.
 
 ## Standalone Redis


src/langsmith/self-host-external-postgres.mdx
```diff
@@ -33,7 +33,7 @@ You will need to provide a connection string to your PostgreSQL database. This c
 * Port
 * Database
 * Username
-* Password (Make sure to url encode this if there are any special characters)
+* Password (Make sure to url encode this if there are any special characters) - **Note:** When using IAM authentication, the password is not required in the connection string. More below.
```
langchain-infra, line 38: should we make it more apparent we support 
IAM auth? either here or in the section above?

src/langsmith/self-host-external-postgres.mdx
```diff
@@ -181,3 +187,108 @@ commonPodSecurityContext:
...
+  iam_connection_url: "<identity-name>@<host>:5432/<database>?sslmode=require"
+```
+
+<Warning>
+IAM authentication requires TLS. You must include `sslmode=require` in your connection string.
+</Warning>
+
+### Required annotations
+
+You must apply the ServiceAccount annotations and pod labels required by your cloud provider's workload identity to all LangSmith components that connect to PostgreSQL. This includes:
```
langchain-infra, line 252: wonder if there is an easier way to add this to all services? will need to update this list anytime we add services or new jobs.

src/langsmith/self-host-external-postgres.mdx
```diff
@@ -181,3 +187,108 @@ commonPodSecurityContext:
...
+    labels:
+      azure.workload.identity/use: "true"
+...
+```
+
+</CodeGroup>
+
+See the [Helm values reference](https://github.com/langchain-ai/helm/blob/main/charts/langsmith/values.yaml) for the full list of configurable services and their annotation/label options.
+
+<Warning>
```
langchain-infra: would add these warnings higher up (maybe at the part where we have the helm chart version

src/langsmith/self-host-external-redis.mdx
```diff
@@ -321,3 +327,111 @@ commonPodSecurityContext:
...
+
+```yaml Kubernetes Secret
+apiVersion: v1
+kind: Secret
+metadata:
+  name: redis-secret
+type: Opaque
+stringData:
+  # IAM connection URL - identity as username, no password
+  iam_connection_url: "rediss://<identity>@<host>:6380"
```
langchain-infra: is it a different secret value? i thought we just use connection_url everywhere

src/langsmith/self-host-external-postgres.mdx
```diff
@@ -181,3 +187,108 @@ commonPodSecurityContext:
...
+IAM authentication requires TLS. You must include `sslmode=require` in your connection string.
+</Warning>
+
+### Required annotations
+
+You must apply the ServiceAccount annotations and pod labels required by your cloud provider's workload identity to all LangSmith components that connect to PostgreSQL. This includes:
+
+**Deployments:** `backend`, `queue`, `platformBackend`, `hostBackend`
+
+**Jobs:** `migrations`, `authBootstrap`, `feedbackConfigMigration`, `feedbackDataMigration`, `e2eTest`
```
langchain-infra, line 256: i think some of these jobs will just share the same service account right? might rephrase this as "these service accounts will need to be granted access".

src/langsmith/self-host-external-redis.mdx
joaquin-borggio-lc: it can be `connection_url`. ill replace.

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: so yes if its by service account then you dont have to do anything here. but for azure, you actually need to add pod labels too on top of service account annotations so that requires adding labels like this:
```
  feedbackConfigMigration:
    labels:
      azure.workload.identity/use: "true"
  feedbackDataMigration:
    labels:
      azure.workload.identity/use: "true"
  ...
```

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: we could ask for the AWS role ARN, or Azure client id, or GCP service account id as a field in the helm value files and then automatically add the right annotations to the jobs, deployments, and service accounts under the hood. The customer wouldnt deal with any annotations/labels at all and would all happen under the hood. Would need to add a couple helm fields that the customer can set.

joaquin-borggio-lc: something like:
```
postgres:
  external:
    enabled: true
    connectionUrl: "blah"
    iamAuthProvider: "azure"
    azureIamClientID: "blah"
    awsIamRoleARN: "blah"
    gcpServiceAccountID: "blah"
```
and they would only need to set 1 of the 3 fields (or we could even come up with some unified name like `iamRoleID`)

but then that restricts what the customer can do. For example, you could have a single IAM role that you use for every service or you could split it up and be more granular with roles. Right now you can do what you want. If we do the above, then we are forcing them to set a single service account for all services. Not sure how big of a deal that would be.

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: I can add a comment that they use the backend service account (except for e2eTest) so this is only relevant if pod labels are required (ex. azure)

langchain-infra: gotcha ya lets just leave it. it's also going to be tricky since you need to merge with blob storage workload identity too. we do also have commonAnnotations and commonLabels but prob not flexible enough

src/langsmith/self-host-external-postgres.mdx
langchain-infra: yea lets leave this for now, prefer not losing the flexibility. will also conflict with blob

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: adding another top bullet points where we also mention things like min versions and cluster support

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: done, added note about this to explain when you need to add pod labels vs just relying on the shared service account

src/langsmith/self-host-external-postgres.mdx
joaquin-borggio-lc: i think its better to err on the side of flexibility/configurability for this security/auth stuff even if it requires the customer to set a few more things. wouldnt want to be in a place where we cant support some auth policies customers have. but i dont feel too too strongly bc it would be nice to have it all done under the hood.

langchain-infra: agreed

langchain-infra (approved):

PR #2102, opened 2026-01-06
Change out deprecated chat dataset type

katmayb:
Fixes DOC-492

Switched out the `chat` dataset type for `kv` in this example.


katmayb:
LGTM'd in Slack

PR #2103, opened 2026-01-06
Clarify the project UUID for langsmith-fetch further

katmayb:
Fixes DOC-581

Additional callouts and clarifications to the project UUID descriptions included on the existing page. Emphasized that the project UUID does not need to be specified unless a user wants to override the default project set at the environment. Also, included more links to sections referring to this on the page.

## Preview

https://langchain-5e9cc07a-preview-additi-1767724954-f64548d.mintlify.app/langsmith/langsmith-fetch

### DIFF
--- a/src/langsmith/langsmith-fetch.mdx
+++ b/src/langsmith/langsmith-fetch.mdx
@@ -19,18 +19,27 @@ pip install langsmith-fetch
 
 ## Setup
 
-Set your [LangSmith API key](/langsmith/create-account-api-key) and project name:
+1. Set your [LangSmith API key](/langsmith/create-account-api-key):
 
-```bash
-export LANGSMITH_API_KEY=lsv2_...
-export LANGSMITH_PROJECT=your-project-name
-```
+    ```bash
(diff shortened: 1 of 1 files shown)


katmayb:
@mdrxy Let me know if you think this needs any further clarification and where you wanted to find it on the page. Thanks!

mdrxy:
@katmayb got it so `--project-uuid <uuid>` should only be used if you're not specifying a trace ID?

katmayb:
> @katmayb got it so `--project-uuid <uuid>` should only be used if you're not specifying a trace ID?

Ah ok, get your point now — taking a look through fetch repo, if you want to specify a specific trace in a specific project (not configured with `LANGSMITH_PROJECT`) then you would want to specify both.

Also, I think threads require a project UUID (set either with the env var or the flag) otherwise it will return an error. 

I'm going to make some more updates on this PR according to this ^, will ping you when I'm done

katmayb:
@mdrxy ptal when you have time, I've tried to address all of the possible cases re traces & threads with project uuid.

PR #2111, opened 2026-01-06
Add note thread-level ttls delete associated checkpoints

katmayb:
Fixes DOC-467


src/langsmith/configure-ttl.mdx
jdrogers940, line 13: I think worth making it more explicitly clear this handles both thread and checkpoint TTLs
```suggestion
## Configuring thread and checkpoint TTL
```

jdrogers940, line 15:
```suggestion
Checkpoints capture the state of conversation threads. Setting a TTL ensures old checkpoints and thread metadata are automatically deleted.
```

katmayb:
TFTR @jdrogers940 added your feedback in!

PR #2137, opened 2026-01-08
add info on tracing to deep agents docs

npentrel:
Please check if you agree with the placement of these config options. I considered putting them under configuration but since they're env variables this feels more appropriate to me


lnhsingh (approved):
Makes sense to me!

PR #2192, opened 2026-01-12
feat(deepagent): add memory feature

npentrel:
## Checklist

### DIFF
--- a/src/oss/deepagents/customization.mdx
+++ b/src/oss/deepagents/customization.mdx
@@ -249,38 +249,6 @@ result = agent.invoke(
 )
 ```
 
-```python FilesystemBackend
-from deepagents import create_deep_agent
-from langgraph.checkpoint.memory import MemorySaver
-from deepagents.backends.filesystem import FilesystemBackend
-
-# Checkpointer is REQUIRED for human-in-the-loop
-checkpointer = MemorySaver()
-
-agent = create_deep_agent(
-    backend=FilesystemBackend(root_dir="/Users/user/{project}"),
-    skills=["/Users/user/{project}/skills/"],
-    interrupt_on={
-        "write_file": True,  # Default: approve, edit, reject
-        "read_file": False,  # No interrupts needed
-        "edit_file": True    # Default: approve, edit, reject
-    },
-    checkpointer=checkpointer,  # Required!
-)
-
-result = agent.invoke(
-    {
-        "messages": [
-            {
-                "role": "user",
-                "content": "What is langgraph?",
-            }
-        ]
-    },
-    config={"configurable": {"thread_id": "12345"}},
-)
-```
-
 ```python StoreBackend
 from urllib.request import urlopen
 from deepagents import create_deep_agent
@@ -304,6 +272,38 @@ agent = create_deep_agent(
     backend=(lambda rt: StoreBackend(rt)),
     store=store,
     skills=["./skills/"]
+)
+
+result = agent.invoke(
+    {
+        "messages": [
+            {
+                "role": "user",
+                "content": "What is langgraph?",
+            }
+        ]
+    },
+    config={"configurable": {"thread_id": "12345"}},
+)
+```
+
+```python FilesystemBackend
+from deepagents import create_deep_agent
+from langgraph.checkpoint.memory import MemorySaver
+from deepagents.backends.filesystem import FilesystemBackend
+
+# Checkpointer is REQUIRED for human-in-the-loop
+checkpointer = MemorySaver()
+
+agent = create_deep_agent(
+    backend=FilesystemBackend(root_dir="/Users/user/{project}"),
+    skills=["/Users/user/{project}/skills/"],
+    interrupt_on={
+        "write_file": True,  # Default: approve, edit, reject
+        "read_file": False,  # No interrupts needed
+        "edit_file": True    # Default: approve, edit, reject
+    },
+    checkpointer=checkpointer,  # Required!
 )
 
 result = agent.invoke(
@@ -378,49 +378,6 @@ let result = await agent.invoke(
 );
 ```
 
-```typescript FilesystemBackend
-import {
-  createDeepAgent,
-  createSkillsMiddleware,
-  createSettings,
-  FilesystemBackend,
-} from "deepagents";
-import { MemorySaver } from "@langchain/langgraph";
-
-const settings = createSettings({
-});
-
-const agent = await createDeepAgent({
-  backend: (config) =>
-    new FilesystemBackend({ rootDir: "/Users/user/{project}" }),
-  skills: [path.join(process.cwd(), ".deepagents/skills")],
-  interruptOn: {
-    read_file: true,
-    write_file: true,
-    delete_file: true,
-  },
-  checkpointer, // Required!
-});
-
-const config = {
-  configurable: {
-    thread_id: `thread-${Date.now()}`,
-  },
-};
-
-let result = await agent.invoke(
-  {
-    messages: [
-      {
-        role: "user",
-        content: "what is langraph? Use the langgraph-docs skill if available.",
-      },
-    ]
-  } as any,
-  config
-);
-```
-
 ```typescript StoreBackend
 import { createDeepAgent, StoreBackend, type FileData } from "deepagents";
 import {
@@ -484,6 +441,49 @@ let result = await agent.invoke(
 );
 ```
 
+```typescript FilesystemBackend
+import {
+  createDeepAgent,
+  createSkillsMiddleware,
+  createSettings,
+  FilesystemBackend,
+} from "deepagents";
+import { MemorySaver } from "@langchain/langgraph";
+
+const settings = createSettings({
+});
+
+const agent = await createDeepAgent({
+  backend: (config) =>
+    new FilesystemBackend({ rootDir: "/Users/user/{project}" }),
+  skills: [path.join(process.cwd(), ".deepagents/skills")],
+  interruptOn: {
+    read_file: true,
+    write_file: true,
+    delete_file: true,
+  },
+  checkpointer, // Required!
+});
+
+const config = {
+  configurable: {
+    thread_id: `thread-${Date.now()}`,
+  },
+};
+
+let result = await agent.invoke(
+  {
+    messages: [
+      {
+        role: "user",
+        content: "what is langraph? Use the langgraph-docs skill if available.",
+      },
+    ]
+  } as any,
+  config
+);
+```
+
 </CodeGroup>
 
 :::
@@ -498,13 +498,14 @@ You can pass one or more file paths to the `memory` parameter when creating your
 
 <CodeGroup>
 ```python StateBackend
-from pathlib import Path
+from urllib.request import urlopen
 
 from deepagents import create_deep_agent
 from deepagents.backends.utils import create_file_data
 from langgraph.checkpoint.memory import MemorySaver
 
-agents_md = Path("AGENTS.md").read_text(encoding="utf-8")
+with urlopen("https://raw.githubusercontent.com/langchain-ai/deepagents/refs/heads/master/examples/text-to-sql-agent/AGENTS.md") as response:
+    agents_md = response.read().decode("utf-8")
 checkpointer = MemorySaver()
 
 agent = create_deep_agent(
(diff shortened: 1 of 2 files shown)


src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,47 @@ const agent = createDeepAgent({
...
+agent = create_deep_agent(
+    backend=FilesystemBackend(root_dir="/Users/user/{project}")
+    memory=[
+        "./AGENTS.md"
+    ],
+)
+```
+:::
+
+:::js
```
npentrel, line 600: Please confirm that for TypeScript it's currently not possible to do this without middleware. I don't see a memory arg, so I think this is the best we can do?

src/oss/deepagents/cli.mdx
```diff
@@ -191,6 +190,18 @@ It remembers for future sessions:
 # Applies conventions without prompting
 ```
 
+## Provide project or user context
+
+[AGENTS.md files](https://agents.md/) files provide persistent memory that is always loaded at session start.
```
mdrxy:
```suggestion
[`AGENTS.md` files](https://agents.md/) files provide persistent memory that is always loaded at session start.
```

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,47 @@ const agent = createDeepAgent({
 });
 ```
 :::
+
+## Memory
+
+:::python
+Use [AGENTS.md files](https://agents.md/) to provide extra context to your deep agent.
```
mdrxy:
```suggestion
Use [`AGENTS.md` files](https://agents.md/) to provide extra context to your deep agent.
```

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,47 @@ const agent = createDeepAgent({
...
+    backend=FilesystemBackend(root_dir="/Users/user/{project}")
+    memory=[
+        "./AGENTS.md"
+    ],
+)
+```
+:::
+
+:::js
+Use [agent.md files](https://agents.md/) to provide extra context to your deep agent.
```
mdrxy:
```suggestion
Use [`agent.md` files](https://agents.md/) to provide extra context to your deep agent.
```

src/oss/deepagents/middleware.mdx
```diff
@@ -352,3 +352,48 @@ const agent = createAgent({
 :::
 
 In addition to any user-defined subagents, the main agent has access to a `general-purpose` subagent at all times. This subagent has the same instructions as the main agent and all the tools it has access to. The primary purpose of the `general-purpose` subagent is context isolation—the main agent can delegate a complex task to this subagent and get a concise answer back without bloat from intermediate tool calls.
+
+## Memory middleware
+
+:::python
+Use [AGENTS.md files](https://agents.md/) to provide extra context to your deep agent.
```
mdrxy:
```suggestion
Use [`AGENTS.md` files](https://agents.md/) to provide extra context to your deep agent.
```

src/oss/deepagents/middleware.mdx
```diff
@@ -352,3 +352,48 @@ const agent = createAgent({
...
+    backend=FilesystemBackend(root_dir="/"),
+    sources=[ "~/.deepagents/AGENTS.md" ],
+)
+
+agent = create_deep_agent(middleware=[memory])
+```
+:::
+
+:::js
+Use [agent.md files](https://agents.md/) to provide extra context to your deep agent.
```
mdrxy:
```suggestion
Use [`agent.md` files](https://agents.md/) to provide extra context to your deep agent.
```

src/oss/deepagents/customization.mdx
vtrivedy: @christian-bromann on this one

src/oss/deepagents/middleware.mdx
```diff
@@ -352,3 +352,48 @@ const agent = createAgent({
 :::
 
 In addition to any user-defined subagents, the main agent has access to a `general-purpose` subagent at all times. This subagent has the same instructions as the main agent and all the tools it has access to. The primary purpose of the `general-purpose` subagent is context isolation—the main agent can delegate a complex task to this subagent and get a concise answer back without bloat from intermediate tool calls.
+
+## Memory middleware
```
vtrivedy: Similar to Skills, it'd help to add examples for how these are used across different backends.

The main reason is that the default backend is the state backend which will be ephemeral so users know their memories won't be updated in State backend but will with filesystem

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,34 @@ const agent = createDeepAgent({
...
+
+Use [`AGENTS.md` files](https://agents.md/) to provide extra context to your deep agent.
+You can pass one or more file paths to the `memory` parameter when creating your deep agent:
+
+:::python
+```python
+from deepagents import create_deep_agent
+
+agent = create_deep_agent(
+    backend=FilesystemBackend(root_dir="/Users/user/{project}")
```
npentrel: Need to use a different backend for safety or enable HIL

src/oss/deepagents/middleware.mdx
```diff
@@ -352,3 +352,80 @@ const agent = createAgent({
...
```
npentrel: Do we want the middleware here? The JS version automatically loads some files so maybe we want to document that but for skills we discussed not adding middleware because it wasn't adding anything for users. Is this something people are likely to want to customize?

christian-bromann: IMHO let's stay as close to the Python interface as we can. @eyurtsev thoughts?

src/oss/deepagents/cli.mdx
```diff
@@ -215,6 +190,18 @@ It remembers for future sessions:
 # Applies conventions without prompting
 ```
 
+## Provide project or user context
+
+[`AGENTS.md` files](https://agents.md/) files provide persistent memory that is always loaded at session start.
+
+To add global-level context, add it to `~/.deepagents/agent/AGENTS.md`.
+To add project-level context, add it to `.deepagents/AGENTS.md` in the project root.
```
eyurtsev: Do we have a definition of what project root is in this document already? If not, I think it's whatever the root where .git folder is (although I don't remember if there's no .git)

src/oss/deepagents/cli.mdx
```diff
@@ -215,6 +190,18 @@ It remembers for future sessions:
...
+
+Both global and project-level `AGENTS.md` files are loaded together and injected into the system prompt at startup.
```
eyurtsev: `injected` -> `appended`

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,61 @@ const agent = createDeepAgent({
...
+```python
+from deepagents import create_deep_agent
+from langgraph.checkpoint.memory import MemorySaver
+from deepagents.backends import FilesystemBackend
+
+# Checkpointer is REQUIRED for human-in-the-loop
+checkpointer = MemorySaver()
+
+agent = create_deep_agent(
+    backend=FilesystemBackend(root_dir="/Users/user/{project}"),
```
eyurtsev: Given that most users would not (and should not) use the `FilesystemBackend` in production with the SD, we should either:

1. Show examples of how to use with each backend (I think this is the desired end state potentially?) <-- this is probably ideal, but if too long to write, let's do option (2)
 
2. We can start with the default backend -- StateBackend (you don't need to specify it) -- it'll use the backend in the checkpointer. And file information is passed via `.invoke(files=[])`

If you clone deepagents and look at the tests directory, you should be able to get sample code snippets pretty quickly.


---

Basically, it's fine that we're showing HIL example which shows how to use filesystembacken safely, but this is still not the production default that developers would want to use

src/oss/deepagents/cli.mdx
```diff
@@ -191,6 +190,18 @@ It remembers for future sessions:
...
+
+Beyond `AGENTS.md`, you can create additional memory files in `.deepagents/` for structured project knowledge. These work similarly to [Anthropic's Memory Tool](https://platform.claude.com/docs/en/agents-and-tools/tool-use/memory-tool).
```
vtrivedy: i think we should clear what these other memory files are and what it means to reference them, for example in the CLI only agents.md will be read so we should make it clear to users that other memory files should be referenced in agents.md for the agent to be aware of them

vtrivedy: and in SDK adding memory file loads the whole thing

src/oss/deepagents/cli.mdx
```diff
@@ -191,6 +190,18 @@ It remembers for future sessions:
...
 ```
 
+## Provide project or user context
+
+[`AGENTS.md` files](https://agents.md/) files provide persistent memory that is always loaded at session start.
+
+To add global-level context, add it to `~/.deepagents/agent/AGENTS.md`.
+To add project-level context, add it to `.deepagents/AGENTS.md` in the project root.
+
+Both global and project-level `AGENTS.md` files are loaded together and injected into the system prompt at startup.
```
npentrel:
```suggestion
Both global and project-level `AGENTS.md` files are loaded together and appended to the system prompt at startup.
```

src/oss/deepagents/cli.mdx
npentrel: If there's no .git folder it silently fails. I ran into this issue when I was testing and it took me a while to figure out. Maybe we could check for the presence of .deepagents/AGENTS.md files in at least the current folder and print a warning if there is no .git folder?

src/oss/deepagents/cli.mdx
npentrel: Oh interesting

src/oss/deepagents/middleware.mdx
npentrel: but this doesn't necessarily have to be done with middleware I guess? Is there any reason currently why someone would want to use the middleware directly?

npentrel:
@eyurtsev I've updated the code samples. Thank you for pointing me in the right direction with invoke. I was struggling to get that to work. 

I'm still unclear about whether there is a reason to keep middleware for this, that means, is there a use case for users to want to go the middleware route rathe than use the higher level creat_deep_agent() to pass in memory info? I've removed it for now. cc: @vtrivedy

src/oss/deepagents/customization.mdx
npentrel: I've added more codesamples - hope I am using them all correctly. I tested the examples and they work

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,228 @@ const agent = createDeepAgent({
...
+
+<CodeGroup>
+```python StateBackend
+from pathlib import Path
+
+from deepagents import create_deep_agent
+from deepagents.backends.utils import create_file_data
+from langgraph.checkpoint.memory import MemorySaver
+
+agents_md = Path("AGENTS.md").read_text(encoding="utf-8")
```
eyurtsev: Could just inline the agents.md instead of reading from file

src/oss/deepagents/customization.mdx
```diff
@@ -198,3 +198,228 @@ const agent = createDeepAgent({
...
+            }
+        ],
+        # Seed the default StateBackend's in-state filesystem (virtual paths must start with "/").
+        "files": {"/AGENTS.md": create_file_data(agents_md)},
+    },
+    config={"configurable": {"thread_id": "123456"}},
+)
+```
+
+```python FilesystemBackend
```
eyurtsev: I'd suggest putting this as the **last part** of the code group to line up with when it would be used in the SDK (it's the least likely backend for the SDK) and we likely want to explain what are the appropriate usages of file system backend

eyurtsev (approved):

src/oss/deepagents/customization.mdx
npentrel: It's a lot more lines. But I'll get one from online like we did with skills

src/oss/deepagents/customization.mdx
npentrel: ok changing the order for skills too

src/oss/deepagents/cli.mdx
```diff
@@ -219,6 +218,22 @@ It remembers for future sessions:
 # Applies conventions without prompting
 ```
 
+## Provide project or user context
+
+[`AGENTS.md` files](https://agents.md/) files provide persistent memory that is always loaded at session start.
+
+You can provide global user memory for the agent in `~/.deepagents/agent/AGENTS.md`.
```
lnhsingh, line 225: nit: are these line breaks intentional? (e.g., after lines 225, 228, 233)

src/oss/deepagents/customization.mdx
```diff
@@ -422,6 +484,171 @@ let result = await agent.invoke(
 );
 ```
 
+</CodeGroup>
+
+:::
+
+## Memory
+
+Use [`AGENTS.md` files](https://agents.md/) to provide extra context to your deep agent.
```
lnhsingh, line 493: nit/question: Should `deep agent` be `Deep Agent`? 

I guess it's not really a proper noun-- the implied full name would be `Deep Agents agent` right?

lnhsingh (approved):
Just a couple nits/comments but nothing blocking

src/oss/deepagents/cli.mdx
npentrel: yes - something we can discuss. I find they make reviews easier than having whole paragraphs on a line. Someone reviewing and adding suggestions only has to edit that sentence rather than having more text. The rendered html is the same so will merge as is for now and put as an item for us to review in the next team meeting

src/oss/deepagents/customization.mdx
npentrel: good questions. I will double check

npentrel: Will make any fixes that result from me checking in a separate PR

PR #2206, opened 2026-01-13
add global install option for deepagents-cli

npentrel:


src/oss/deepagents/cli.mdx
```diff
@@ -38,9 +38,16 @@ The Deep Agents CLI has the following built-in capabilities:
     </Step>
 
     <Step title="Run the CLI" icon="terminal">
-        ```bash
-        uvx deepagents-cli
-        ```
+        <CodeGroup>
+            ```bash Install globally
```
eyurtsev, line 42: These are both using UV so I think title is maybe a bit misleading in terms of titles

`uvx` = `uv tool run` -> runs without attempting to do a global install
`uv tool install` -> does a global install


More information is here:

https://docs.astral.sh/uv/concepts/tools/#relationship-to-uv-run

There's also `uv tool upgrade` to upgrade the tool version

src/oss/deepagents/cli.mdx
```diff
@@ -38,9 +38,16 @@ The Deep Agents CLI has the following built-in capabilities:
...
+            uv tool install deepagents-cli
+            deepagents
+            ```
+
+            ```bash Use with uv
```
npentrel:
```suggestion
            ```bash Run (without global install)
```

src/oss/deepagents/cli.mdx
npentrel: Changed to `Run (without global install)` - do you think that works?

npentrel:
Making exec decision to merge. Fairly sure this is good.

PR #2211, opened 2026-01-13
Auto trigger publish when pushing to main

lnhsingh:
Originally when we set up this docs repo, we created a manual workflow for publishing in the event that we wanted to merge something but not release it yet. In practice, we haven't had to do that ever and it has caused issues with people merging and then forgetting to publish.

This PR automates the trigger to publish docs.


npentrel (approved):
AWESOME

eyurtsev (approved):
perfect

PR #2229, opened 2026-01-14
Self-hosted renav for cloud providers

katmayb:
Fixes DOC-544

Improves cloud-specific discoverability in LangSmith self-hosted docs by creating a centralized page for each cloud provider

- Provides initial setup steps for AWS (EKS), GCP (GKE), and Azure (AKS)
- Links to cloud-managed services (S3/RDS/ElastiCache, GCS/Cloud SQL/Memorystore, Azure Blob/Database/Cache)
- Includes authentication setup guidance (IRSA, Workload Identity, Azure Workload Identity)
- Consolidates with the existing architecture guides.
- Added to navigation under "Self-hosted cloud guides" section

### Approach 

This proposed solution provides cloud-specific discoverability in the places where it really matters (like initial setup, authentication) while keeping the detailed configuration guides clean and maintainable.

## Preview

https://langchain-5e9cc07a-preview-selfho-1768586885-555e1d7.mintlify.app/langsmith/aws-self-hosted


src/langsmith/cloud-provider-guides.mdx
```diff
@@ -0,0 +1,202 @@
...
+  </Card>
+  <Card title="GCP" icon="google" href="#gcp-google-cloud-platform">
+    Google Cloud Platform (GKE)
+  </Card>
+  <Card title="Azure" icon="microsoft" href="#azure-microsoft-azure">
+    Microsoft Azure (AKS)
+  </Card>
+</CardGroup>
+
+## AWS (Amazon Web Services)
```
katmayb: Actually think we can get rid of the parenthesis full name from the cloud provider headers

lnhsingh:
I like the general direction.

image

Maybe these could be elevated a level? So it would be something like:

```
- Cloud quickstart or Get started
    - AWS architecture
    - Azure architecture
    - GCP architecture
```

langchain-karan:
Overall, I think having specific guides to clouds (while it might be slightly redundant) is good because we are being prescriptive for customers and our partners.

- A lot of times customers have a primary cloud of choice - its hard to sift through common setups and AWS setups vs having 1 place for that cloud provider
- Most mature companies like us have done this historically
- As we deepen our relationships with cloud providers - its will be super convenient to give them 1 guide as opposed to them figuring out from our docs.

The way you have it setup works actually. One suggestion, similar to Lauren's suggestion:
On the left pane, under "Get started by cloud provider", have the individual cloud links - AWS setup, Azure setup, GCP setup
Combined the content under "Self-hosted architecture" to the individual section?
Screenshot 2026-01-15 at 10 42 45 AM

katmayb:
image

@langchain-karan I've consolidated the get started page with the architecture pages, so we have the attached nav.

PR #2247, opened 2026-01-15
oss: generalize versioning policy

mdrxy:
I think we should potentially also explicitly state where `deepagents` falls?

### DIFF
--- a/src/oss/versioning.mdx
+++ b/src/oss/versioning.mdx
@@ -2,7 +2,7 @@
 title: Versioning
 ---
 
-LangChain OSS version numbers follow the format: `MAJOR.MINOR.PATCH`, as defined by [Semantic Versioning](https://semver.org/).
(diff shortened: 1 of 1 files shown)


src/oss/versioning.mdx
```diff
@@ -2,17 +2,13 @@
 title: Versioning
 ---
 
-
-
-Each LangChain and LangGraph version number follows the format: `MAJOR.MINOR.PATCH`
+LangChain OSS version numbers follow the format: `MAJOR.MINOR.PATCH`, as defined by [Semantic Versioning](https://semver.org/).
```
lnhsingh:
```suggestion
Our OSS version numbers follow the format: `MAJOR.MINOR.PATCH`, as defined by [Semantic Versioning](https://semver.org/).
```
Maybe to make less confusing with the whole LangChain the Company vs LangChain the Package

lnhsingh:
> I think we should potentially also explicitly state where `deepagents` falls?

in that it's also an oss package? we could add a list of all the os packages if that's the case

PR #2264, opened 2026-01-16
feat(deepagents): add new cli commands

npentrel:
Probably not complete but at least adds the new commands

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -288,16 +288,10 @@
                   "pages": [
                     "oss/python/deepagents/overview",
                     {
-                      "group": "Deep Agents SDK",
+                      "group": "Get started",
                       "pages": [
                         "oss/python/deepagents/quickstart",
                         "oss/python/deepagents/customization"
-                      ]
-                    },
-                    {
-                      "group": "Deep Agents CLI",
-                      "pages": [
-                        "oss/python/deepagents/cli"
                       ]
                     },
                     {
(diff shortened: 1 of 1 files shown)


src/oss/deepagents/cli.mdx
```diff
@@ -126,9 +127,11 @@ API keys can be set as environment variables or in a `.env` file.
         |------------------------|-------------------------------------------------------------|
         | `--agent NAME`         | Use named agent with separate memory                        |
         | `--auto-approve`       | Skip tool confirmation prompts (toggle with `Ctrl+T`)       |
-        | `--sandbox TYPE`       | Execute in remote sandbox: `modal`, `daytona`, or `runloop` |
+        | `-r`                   | Resume most recent session |
```
eyurtsev: --resume for consistency here (we can abbreviate like --resume (-r) I think

src/docs.json
```diff
@@ -288,12 +288,18 @@
                   "pages": [
                     "oss/python/deepagents/overview",
                     {
-                      "group": "Get started",
+                      "group": "Deep Agents SDK",
```
eyurtsev: Would you mind generating a preview on mintlify so we can see how this would look?

npentrel: I don't seem to be able right now, I'll have to see why the GitHub Action didn't fire properly. In the meantime, this is what it looks like:
Screenshot 2026-01-16 at 22 09 04

PR #2327, opened 2026-01-22
Add traces/ skills section, and link for memory to AB Essentials

lnhsingh:
https://langchain.slack.com/archives/C08N2LV7HM2/p1769034274408239

Fixes DOC-624

### DIFF
--- a/src/langsmith/agent-builder-essentials.mdx
+++ b/src/langsmith/agent-builder-essentials.mdx
@@ -29,6 +29,12 @@ Here are some popular ways to trigger your agent:
   </Card>
 </CardGroup>
 
+## Skills
+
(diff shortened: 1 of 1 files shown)


src/langsmith/agent-builder-essentials.mdx
```diff
@@ -29,6 +29,12 @@ Here are some popular ways to trigger your agent:
   </Card>
 </CardGroup>
 
+## Skills
+
+Skills are a way to bundle capabilities together into larger actions and provide additional context beyond single tool descriptions.
```
nhuang-lc, line 34: Ik this merged! 

Skills I think are more akin to memory than tools (in the context of agent builder). I like to think of them as memories or instructions which are not **always** necessary and are thus not worth putting in the system prompt **every time**. But the agent can still discover and read these easily through the lens of skills

PR #2368, opened 2026-01-26
add structured output info

npentrel:


src/oss/deepagents/subagents.mdx
```diff
@@ -311,6 +311,20 @@ In this case the subagent with the name `"research-agent"`, will have `{'lc_agen
 
 ![LangSmith Example trace showing the metadata](/oss/images/deepagents/deepagents-langsmith.png)
 
+## Structured output
+
+All subagents support [structured ouput](/oss/langchain/structured-output).
+:::python
+You can set a desired structured output schema by passing it as the `response_format` argument to the call to `create_agent()`.
```
sydney-runkle, line 319: Sometimes you create subagents w/ a dict spec, so we should mention that as well (not the compiled version)

I'm not actually sure what we do if a subagent returns structured output - ex, do we pass it back to the main model? Should investigate and document!

npentrel: The structured response is actually dropped - you need to add the content to the Toolmessage if you want to use it: https://github.com/langchain-ai/deepagents/blob/79003c2629b75a8867f12b7cee92e2ccae9d5c70/libs/deepagents/deepagents/middleware/subagents.py#L113.  Will update

src/oss/deepagents/customization.mdx
```diff
@@ -748,3 +748,31 @@ You can pass one or more file paths to the `memory` parameter when creating your
 </Tabs>
 
 :::
+
+## Structured ouput
+
+Deep agents support [structured ouput](/oss/langchain/structured-output).
+:::python
+You can set a desired structured output schema by passing it as the `response_format` argument to the call to `create_deep_agent()`.
```
sydney-runkle, line 756: can we expand this example a bit? like define a pydantic model to use that makes sense in the context of the agent? can be as simple as `WeatherReport` or something like that

src/oss/deepagents/subagents.mdx
sydney-runkle: Interesting, we probably should include it, let's loop in @eyurtsev here... maybe as an artifact on the tool message, would require a code change

src/oss/deepagents/customization.mdx
npentrel: done!

src/oss/deepagents/subagents.mdx
npentrel: @sydney-runkle @eyurtsev - what are we thinking here?

PR #2370, opened 2026-01-26
add exceptions for large tool result eviction

npentrel:
This may be more information than needed. I wonder if we even need to mention the exceptions since for most of them this behaviour might be expected?

### DIFF
--- a/src/oss/deepagents/harness.mdx
+++ b/src/oss/deepagents/harness.mdx
@@ -46,7 +46,7 @@ The [`FilesystemMiddleware`](/oss/deepagents/middleware) automatically evicts la
 - When exceeded, writes the result using the configured backend
 - Replaces the tool result with a truncated preview and file reference
 - Agent can read the full result from the file system as needed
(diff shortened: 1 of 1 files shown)


src/oss/deepagents/harness.mdx
```diff
@@ -46,6 +46,15 @@ The [`FilesystemMiddleware`](/oss/deepagents/middleware) automatically evicts la
 - When exceeded, writes the result using the configured backend
 - Replaces the tool result with a truncated preview and file reference
 - Agent can read the full result from the file system as needed
+- If the CLI /large_tool_results/ produces large results, they are written to `/large_tool_results/`.
```
npentrel:
```suggestion
- If the CLI produces large results, they are written to `/large_tool_results/`.
```

eyurtsev, line 49: We use a composite backend to remap this to another folder.

And we don't need the list of exceptions

npentrel:
decided this was actually a no-op. Making a separate ticket for FilesystemBackend usage

PR #2431, opened 2026-02-02
add table to products overview

npentrel:
Screenshot 2026-02-02 at 18 22 13

### DIFF
--- a/src/langsmith/agent-server-changelog.mdx
+++ b/src/langsmith/agent-server-changelog.mdx
@@ -11,6 +11,13 @@ rss: true
 
 [Agent Server](/langsmith/agent-server) is an API platform for creating and managing agent-based applications. It provides built-in persistence, a task queue, and supports deploying, configuring, and running assistants (agentic workflows) at scale. This changelog documents all notable updates, features, and fixes to Agent Server releases.
 
+<Update label="2026-02-02" tags={["agent-server"]}>
+## v0.7.14
+
+- Ensured selected system fields are excluded from custom encryption to prevent unnecessary encryption of non-sensitive data.
+- Introduced a custom checkpointer adapter with unit tests to validate implementation checks.
+</Update>
+
 <Update label="2026-01-29" tags={["agent-server"]}>
 ## v0.7.13
 
--- a/src/oss/deepagents/quickstart.mdx
+++ b/src/oss/deepagents/quickstart.mdx
@@ -26,10 +26,6 @@ Before you begin, make sure you have an API key from a model provider (e.g., Ant
     uv add deepagents tavily-python
     uv sync
     ```
-
-    ```bash poetry
-    poetry add deepagents tavily-python
-    ```
 </CodeGroup>
 :::
 


src/oss/concepts/products.mdx
```diff
@@ -6,6 +6,12 @@ description: Understand the differences between LangChain, LangGraph, and DeepAg
 
 LangChain maintains several open source packages to help you build agents. Each serves a different purpose in the agent development stack. Understanding the distinctions between [agent frameworks](#agent-frameworks-like-langchain),[agent runtimes](#agent-runtimes-like-langgraph), and [agent harnesses](#agent-harnesses-like-the-deep-agents-sdk) helps you choose the right tool for your needs.
 
+|   | Framework | Runtime | Harness |
```
ccurme, line 9: I'd do `Harness | Framework | Runtime` or `Runtime | Framework | Harness` (harness is built on top of the framework, which is built on top of the runtime).

mdrxy:
Screenshot

For "When to use" maybe we shouldn't use number list since it looks crowded

lnhsingh:
To echo Mason's comment, I think you can remove the bullet points and just use `<br>` for the Value Add / When to use columns

src/oss/concepts/products.mdx
```diff
@@ -6,6 +6,12 @@ description: Understand the differences between LangChain, LangGraph, and DeepAg
...
+| - | --------- | ------- | ------- |
+| Value add | <ul><li>Abstractions</li><li>Integrations</li></ul> | <ul><li>Durable execution</li><li>Streaming</li><li>HITL</li><li>Persistence</li></ul> | <ul><li>Predefined tools</li><li>Prompts</li><li>Subagents</li></ul> |
+| When to use | <ol><li>Getting started quickly</li><li>Standardizing how a team builds</li></ol> | <ol><li>Low-level control</li><li>Long running, stateful workflows and agents</li></ol> | <ol><li>More autonomous agents</li><li>Agents faced with complex, non-deterministic tasks</li></ol> |
```
ccurme: What is meant by non-deterministic here?

Maybe another point on when to use harness: when working with search results, scripts, or other artifacts in state.

src/oss/concepts/products.mdx
```diff
@@ -6,6 +6,12 @@ description: Understand the differences between LangChain, LangGraph, and DeepAg
...
+| Other options | Vercel’s AI SDK, CrewAI, OpenAI Agents SDK, Google ADK, LlamaIndex | Temporal, Inngest | Claude Agent SDK, Manus |
```
ccurme: Maybe instead of "Other options" we can have "Implementations", "Products", "Options" or similar, and put LangChain, LangGraph, and Deep Agents as the first one in each?

src/oss/concepts/products.mdx
ccurme: can ignore this though given the ordering of sections in the page

src/oss/concepts/products.mdx
npentrel: I'll add the latter to the text below - that otherwise gets too detailed compared to the high level overview we're showing for the others. 
Non-deterministic tasks is a reference to how software used to be built fairly deterministic but agents, and especially multi-step agents open us up to more non-deterministic tasks.

src/oss/concepts/products.mdx
npentrel: done

PR #2434, opened 2026-02-02
fix(deepagents): fix JS skills example

christian-bromann:
## Before

- example was bad formatted
- contained unnecessary imports
- used `as any`

docsbefore

## After
docs


npentrel (approved):
Thank you for catching and improving those!

PR #2455, opened 2026-02-03
add HIL for subagents example

npentrel:


src/oss/deepagents/human-in-the-loop.mdx
```diff
@@ -432,6 +432,229 @@ const agent = createDeepAgent({
 
 When a subagent triggers an interrupt, the handling is the same – check for `__interrupt__` and resume with `Command`.
 
+## Tools that use `interrupt()` directly
```
sydney-runkle: I think should be a ### (related to subagent interrupts)

npentrel:
```suggestion
### Tools that use `interrupt()` directly
```

PR #2492, opened 2026-02-05
move middleware page to LangChain built-in

npentrel:
Fairly easy move because I don't actually think any of this info should be with deep agents. If you are using this info you're not using deep agents. 
The only question is if we need both to do list middlewares or if they are the same anyway. I haven't looked into it. I assume you probably know the answer immediately


npentrel:
https://langchain-5e9cc07a-preview-naomid-1770298845-2294868.mintlify.app/oss/python/langchain/middleware/built-in

PR #2504, opened 2026-02-05
fill in customization page

npentrel:
Decided to focus next on filling in the customization page. I left out structured ouput and context for now, will add that, possibly in a separate PR. We don't currently have info on it so we're not losing anything.

https://langchain-5e9cc07a-preview-naomid-1770324835-64a7eea.mintlify.app/oss/python/deepagents/customization

### DIFF
--- a/src/oss/deepagents/customization.mdx
+++ b/src/oss/deepagents/customization.mdx
@@ -301,8 +301,7 @@ For more information, see [Subagents](/oss/deepagents/subagents).
 
 ## Backends
 
-For [skills](#skills) and [memory](#memory), you must provide a filesystem backend.
-You may also find it helpful to use the backend with custom tools.
+You can provide your deep agent with one of the following virtual filesystems:
 
 <Tabs>
   <Tab title="StateBackend">
@@ -313,6 +312,12 @@ You may also find it helpful to use the backend with custom tools.
   </Tab>
   <Tab title="FilesystemBackend">
     The local machine's filesystem.
+
+    <Warning>
+    This backend grants agents direct filesystem read/write access.
+    Use with caution and only in appropriate environments.
+    For more information, see [FilesystemBackend](/oss/deepagents/backends#filesystembackend-local-disk).
+    </Warning>
 
     <BackendFilesystem />
   </Tab>


src/oss/deepagents/customization.mdx
```diff
@@ -4,123 +4,84 @@ sidebarTitle: Customization
...
+import HitlBasicConfig from '/snippets/hitl-basic-config.mdx';
+import BackendState from '/snippets/backend-state.mdx';
+import BackendFilesystem from '/snippets/backend-filesystem.mdx';
+import BackendStore from '/snippets/backend-store.mdx';
+import BackendComposite from '/snippets/backend-composite.mdx';
+import SubagentBasic from '/snippets/subagent-basic.mdx';
 
-## Model
+:::python
+`create_deep_agent` has the following core configuration options:
```
ccurme, line 17: what is the difference between core configuration options and configurable features?

the model is maybe the only thing that I'd call "core". `tools` we don't necessarily need (the agent comes equipped with some from the filesystem already). there's also a default system promt.

src/oss/deepagents/customization.mdx
```diff
@@ -4,123 +4,84 @@ sidebarTitle: Customization
...
-    from langchain.chat_models import init_chat_model
-    from deepagents import create_deep_agent
+And the following configurable features:
 
-    model = init_chat_model(model="openai:gpt-5")
-    agent = create_deep_agent(model=model)
-    ```
+- [Middleware](#middleware)
+- [Subagents](#subagents)
+- [Filesystem backend](#filesystem-backend)
```
ccurme: naming here is tricky. "filesystem backend" helps explain the concept, but `backend` is a general thing and `filesystem backend` is one particular instance of a backend. maybe "Backend (virtual filesystem)" or similar?

src/oss/deepagents/customization.mdx
```diff
@@ -4,123 +4,84 @@ sidebarTitle: Customization
...
+  tools?: TTools | StructuredTool[],
+  systemPrompt?: string | SystemMessage,
 });
 ```
+
+For more information, see [Customizing Deep Agents](https://reference.langchain.com/javascript/modules/deepagents.html#customizing-deep-agents).
 :::
 
-## System prompt
+### Model
```
ccurme: OK for now but we might expand on this bit.

IMO we should show this explicitly:
```python
create_deep_agent(model="openai:gpt-5.2")
```
But explain that this method initializes with default model params. If users want to apply any specific model params, they should use `init_chat_model` (as shown here). For example:
```python
model = init_chat_model("openai:gpt-5.2", use_responses_api=True, store=False)
```
^ that particular example is important because for OpenAI, we may soon update the default in the deepagents SDK to use the responses API, which has some consequences for data retention, so we want to surface to users how to adjust that (via `store=False`)

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
 ```
 :::
 
-## Skills
+### System prompt
+
+Deep agents come with a built-in system prompt inspired by Claude Code's system prompt. The default system prompt contains detailed instructions for using the built-in planning tool, file system tools, and subagents.
```
ccurme: Is it important to say it's inspired by Claude Code's system prompt? It is also inspired by system prompts from other tools and our own optimizations. Wondering if this is a hold-over from early days of deepagents.

ccurme: Probably important to mention here that when middleware add special tools, like the filesystem tools, it can [append to the system prompt](https://github.com/langchain-ai/deepagents/blob/82de717323e3f9b7d431c78f2bdfae01c3491eaa/libs/deepagents/deepagents/middleware/filesystem.py#L928-L943). Might be worth a diagram or bullets explaining what composes the final prompt (can be a todo from eng)

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+  systemPrompt: researchInstructions,
+});
+```
+:::
+
+## Features
+
+### Middleware
+
+Middleware provides a way to more tightly control what happens inside an agent.
```
ccurme:
```suggestion
[Middleware](/oss/langchain/middleware/overview) provides a way to more tightly control what happens inside an agent.
```

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+You can provide additional middleware to extend functionality, add tools, or implement custom hooks:
```
ccurme, line 197: Should we list what middleware deepagents ships with by default?

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+You may also find it helpful to use the backend with custom tools.
+
+<Tabs>
+  <Tab title="StateBackend">
+    An ephemeral filesystem backend stored in `langgraph` state.
+    This filesystem only persists _for a single thread_.
+
+    <BackendState />
+  </Tab>
+  <Tab title="FilesystemBackend">
```
ccurme, line 313: seems wrong for this tab to be the same as the section title

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+
+<SubagentBasic />
+
+For more information, see [Subagents](/oss/deepagents/subagents).
+
+{/* TODO(https://github.com/langchain-ai/docs/pull/2368/) #### Structured response format */}
+
+{/* ### Context - You can persist agent state between runs to store information like user IDs. */}
+
+### Filesystem backend
```
ccurme: IMO it's confusing to call this Filesystem backend (see comment above)

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+
+For [skills](#skills) and [memory](#memory), you must provide a filesystem backend.
```
ccurme: Maybe a nitpick but it would be good to clarify the behavior here. `create_deep_agent` defaults to state backend, so there's always a backend. If using skills and memory, you need to add these to the backend yourself, then pass that in when creating the agent.

src/oss/deepagents/customization.mdx
ccurme: Everywhere we use Filesystem backend we should probably clarify that it is not recommended for production use.

src/oss/deepagents/customization.mdx
```diff
@@ -199,7 +160,193 @@ const agent = createDeepAgent({
...
+
+Some tool operations may be sensitive and require human approval before execution.
+You can configure the approval for each tool:
+
+<HitlBasicConfig />
+
+You can configure interrupt for agents and subagents on tool call as well as from within tool calls.
+For more information, see [Human-in-the-loop](/oss/deepagents/human-in-the-loop).
+
+### Skills
```
ccurme: Maybe outside of scope of this PR but we should improve this section.

State backend is a special case where skills can be passed through invoking the agent. For the others, I think we want to create the backend and then use `backend.upload_files` to make it consistent.

Where we show Filesystem backend we should use in virtual mode.

npentrel: Yes outside scope - but I will make a ticket. https://linear.app/langchain/issue/DOC-696/improve-skills

src/oss/deepagents/customization.mdx
npentrel:
```suggestion
### Backends
```

src/oss/deepagents/customization.mdx
npentrel:
```suggestion
- [Backends (virtual filesystems)](#backends)
```

src/oss/deepagents/customization.mdx
npentrel: will remove - it was split that way on the page before, but yeah some of it isn't even required

src/oss/deepagents/customization.mdx
npentrel: ok

src/oss/deepagents/customization.mdx
npentrel:
```suggestion
Deep agents come with a built-in system prompt. The default system prompt contains detailed instructions for using the built-in planning tool, file system tools, and subagents.
When middleware add special tools, like the filesystem tools, it appends them to the system prompt.
```
Will add and make follow up ticket for diagram etc

src/oss/deepagents/customization.mdx
npentrel: yes - making a ticket because I don't know the answer directly

src/oss/deepagents/customization.mdx
npentrel: I think probably actually better to leave that wclarification with skills and memory. I'll make a ticket though

PR #2542, opened 2026-02-09
add more dynamic tool examples

npentrel:
Slight reordering and adding more examples

https://langchain-5e9cc07a-preview-naomid-1770642567-d5193bc.mintlify.app/oss/python/langchain/agents#tools

### DIFF
--- a/src/oss/langchain/agents.mdx
+++ b/src/oss/langchain/agents.mdx
@@ -294,7 +294,7 @@ If an empty tool list is provided, the agent will consist of a single LLM node w
 
 #### Dynamic tools
 
-Dynamic tools modify the set of tools available to the agent at runtime rather than defining them all upfront. Not every tool is appropriate for every situation. Too many tools may overwhelm the model (overload context) and increase errors; too few limit capabilities. Dynamic tool selection adapts the available toolset based on authentication state, user permissions, feature flags, or conversation stage.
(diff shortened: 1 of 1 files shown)


src/oss/langchain/agents.mdx
```diff
@@ -290,215 +292,304 @@ const agent = createAgent({
...
-agent = create_agent(
-    model="gpt-4.1",
-    tools=[search, get_weather],
-    middleware=[handle_tool_errors]
-)
-```
+#### Dynamic tools
 
-The agent will return a @[`ToolMessage`] with the custom error message when a tool fails:
+Dynamic tools modify the set of tools available to the agent at runtime rather than defining them all upfront. Not every tool is appropriate for every situation. Too many tools may overwhelm the model (overload context) and increase errors; too few limit capabilities. Dynamic tool selection adapts the available toolset based on authentication state, user permissions, feature flags, or conversation stage.
```
sydney-runkle:
```suggestion
With dynamic tools, the set of tools available to the agent is modified at runtime rather than defined all upfront. Not every tool is appropriate for every situation. Too many tools may overwhelm the model (overload context) and increase errors; too few limit capabilities. Dynamic tool selection enables adapting the available toolset based on authentication state, user permissions, feature flags, or conversation stage.
```

PR #2569, opened 2026-02-10
add sandboxes to customization

npentrel:


src/oss/deepagents/customization.mdx
```diff
@@ -1036,3 +1037,54 @@ console.log(result.structuredResponse);
 :::
 
 For more information and examples, see [response format](/oss/langchain/structured-output#response-format).
+
+## Sandboxes
+
+Sandboxes are specialized [backends](/oss/deepagents/backends) that run agent code in an isolated environment with their own filesystem and an `execute` tool for shell commands.
```
eyurtsev, line 1045: Should `sandboxes` here link to the sandboxes page?

Is there a way to make it clear that we support different integrations not just modal? I'd prefer having a switcher here that shows how to do it with every provider or make it clear that we can switch and link to the appropriate page for more information

npentrel: can do

npentrel:
https://github.com/langchain-ai/docs/pull/2574

PR #2574, opened 2026-02-10
Add sandboxes customization

npentrel:
https://langchain-5e9cc07a-preview-addsan-1770744332-3d759b5.mintlify.app/oss/python/deepagents/customization#sandboxes

### DIFF
--- a/src/oss/deepagents/customization.mdx
+++ b/src/oss/deepagents/customization.mdx
@@ -32,7 +32,6 @@ import SandboxBasicJs from '/snippets/deepagents-sandbox-basic-js.mdx';
 - [Human-in-the-loop](#human-in-the-loop)
 - [Skills](#skills)
 - [Memory](#memory)
-- [Sandboxes](#sandboxes)
 
 :::python
 ```python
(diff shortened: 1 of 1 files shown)


src/oss/deepagents/customization.mdx
```diff
@@ -1036,3 +1039,20 @@ console.log(result.structuredResponse);
 :::
 
 For more information and examples, see [response format](/oss/langchain/structured-output#response-format).
+
+## Sandboxes
```
ccurme: should this be a sub-section of the `Backends` section?

src/oss/deepagents/sandboxes.mdx
```diff
@@ -4,6 +4,9 @@ sidebarTitle: Sandboxes
 description: Execute code in isolated environments with sandbox backends
 ---
 
+import SandboxBasicPy from '/snippets/deepagents-sandbox-basic-py.mdx';
```
ccurme, line 7: will we re-use these?

npentrel: they're used in two places in this PR to avoid duplicating them

src/oss/deepagents/customization.mdx
npentrel: Sandboxes have their own left nav item, so I went with giving theme their own but I'll change it to a subsection for now.

PR #2575, opened 2026-02-10
deepagents: add missing pip installs for langchain-modal

ccurme:


ccurme:
Would conflict with https://github.com/langchain-ai/docs/pull/2574

PR #2584, opened 2026-02-10
Papercuts

lnhsingh:
Fixes DOC-699
Fixes DOC-523
Fixes DOC-524

### DIFF
--- a/src/langsmith/observability-llm-tutorial.mdx
+++ b/src/langsmith/observability-llm-tutorial.mdx
@@ -5,7 +5,7 @@ sidebarTitle: Trace an LLM application
 
 import project from '/snippets/langsmith/trace-ingestion-project.mdx';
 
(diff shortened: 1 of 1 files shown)


src/langsmith/langgraph-python-sdk.mdx
```diff
@@ -1,5 +1,5 @@
 ---
 title: LangGraph Python SDK
```
npentrel, line 2: Can we change the titles too? They should probably match where they are going right?
"LangSmith Deployment SDK reference"

src/langsmith/observability-llm-tutorial.mdx
npentrel: Should you'd also be LLM then?

npentrel (approved):
See comments but leaving final decision to you

PR #2618, opened 2026-02-12
add warning around mutating attributes in custom middleware

npentrel:


src/oss/deepagents/customization.mdx
```diff
@@ -195,25 +195,44 @@ const agent = createDeepAgent({
...
-- `HumanInTheLoopMiddleware`: pauses for human approval or input at specified points when the `interrupt_on` argument is provided
+- `MemoryMiddleware`: if the `memory` argument is provided
+- `SkillsMiddleware`: if the `skills` argument is provided
+- `HumanInTheLoopMiddleware`: if the `interrupt_on` argument is provided
 
 You can provide additional middleware to extend functionality, add tools, or implement custom hooks:
 
+<Warning>
+Middleware implementations should never mutate any attributes post initialization.
+If mutation is needed, leverage graph state instead.
```
eyurtsev: This needs a bit of paraphrasing I think to be clearer

Basically mutation is used to achieve some goal -- is there a good way to call that goal and then 

If you're trying to do X you should rely on updating langgraph short-term memory / state which is scoped to a thread by design instead of using mutation in place patterns in the middleware. Mutation in place patterns can lead to ...,

Example could be useful:

DO NOT DO THIS

```python
class CustomMiddleware(..):
   def __init__(self):
     self.x = 1


  def before_agent(self):
     self.x += 1
```

Probably should do this:
 
```python
class CustomMiddleware(..):
   def __init__(self):
    pass



  def before_agent(self, state):
    return {
      'x': state['x'] + 1
    }
```

---

It could also be a more general recommendation of being careful with respect to race condition and just making sure that if users develop custom middleware with mutation in place then they should do it carefully and think about what would happen if the middleware is used concurrently

npentrel: OK - iterating on this

PR #2625, opened 2026-02-12
add nextjs as an example of a custom transport

npentrel:


src/oss/langchain/streaming/frontend.mdx
```diff
@@ -2467,6 +2467,99 @@ function CustomAPIChat({ apiKey }: { apiKey: string }) {
...
+import { MemorySaver } from "@langchain/langgraph";
+import type { BaseMessage } from "@langchain/core/messages";
+
+const checkpointer = new MemorySaver();
+
+const agent = createAgent({
+  model: new ChatAnthropic({ model: "claude-sonnet-4-5-20250929" }),
+  tools: [/* your tools */],
+  checkpointer,
+});
```
christian-bromann: Let's move this into the `POST` function to avoid triggering the same agent multiple times. I am not even sure what would happen here as we would have multiple LLMs modifying the agent state.

src/oss/langchain/streaming/frontend.mdx
```diff
@@ -2467,6 +2467,99 @@ function CustomAPIChat({ apiKey }: { apiKey: string }) {
...
+**When to use your own API routes vs LangSmith**
+
+For basic agent interaction (streaming, tool calls), rolling your own Next.js API routes works well. For persisting conversations, loading thread history, and conversation branching, consider [LangSmith deployments](/langsmith/deployments), which provide these features out of the box.
+</Tip>
+
+#### Server: Stream from an API route
+
+In your Next.js API route, stream the agent's output by returning a `Response` with `text/event-stream` encoding:
+
+:::js
```
christian-bromann: I would suggest just wrapping everything here into `:::js` .. I don't think this applies in any way to folks that work with Python.

npentrel: yeah good point, I thought I had done that already

christian-bromann (approved):

PR #2643, opened 2026-02-13
Add override model name parameter to trace with langchain

katmayb:
Fixes DOC-729

## Preview

https://langchain-5e9cc07a-preview-langch-1770997142-fbbbd19.mintlify.app/langsmith/trace-with-langchain#override-model-name-in-traces


jacoblee93 (approved):
LGTM, I wonder if linking out to our existing docs on the `ls_` fields in LangSmith in general

PR #2651, opened 2026-02-13
fix broken linkcheck

npentrel:


src/oss/deepagents/sandboxes.mdx
```diff
@@ -587,5 +587,5 @@ Even with these safeguards, this remains an unsafe workaround. A sufficiently cr
 
 - Review sandbox outputs before acting on them in your application
 - Block sandbox network access when not needed
-- Use [middleware](/oss/deepagents/middleware) to filter or redact sensitive patterns in tool outputs
+- Use [middleware](/oss/langchain/middleware/built-in) to filter or redact sensitive patterns in tool outputs
```
npentrel:
```suggestion
- Use [middleware](/oss/langchain/middleware) to filter or redact sensitive patterns in tool outputs
```

lnhsingh:
@npentrel fixed what I could, but have some open questions.

The following are correct but maybe we need to add `https://docs.langchain.com` since these are generated on build?

```
langsmith/api-ref-control-plane.mdx
 ⎿  /api-reference/integrations-v1/list-github-integrations
 ⎿  /api-reference/deployments-v2
 ⎿  /api-reference/listeners-v2
 ⎿  /api-reference/auth-service-v2
```

I'm not sure what the links are here that are wrong since it's empty:

```
langsmith/server-api-ref.mdx

langsmith/use-threads.mdx

oss/javascript/deepagents/long-term-memory.mdx

oss/python/deepagents/long-term-memory.mdx
```

Also unsure about how to fix this one because (from what I can tell) in the file, it says `/oss/langchain/agents` (not `../`):

```
snippets/oss/agent-chat-ui.mdx
 ⎿  ../langchain/agents
```

npentrel:
The issue is that `mint broken-links` checks snippets and doesn't actually first build the docs. So the best we can do is filter those false positives out. Which leaves us open to hidden broken links. Maybe I'll implement a secondary link test with `htmltest` at some point

lnhsingh:
I’ll mention to Mintlify as well. Seems like a gapOn Feb 16, 2026, at 9:52 AM, Naomi Pentrel ***@***.***> wrote:﻿npentrel left a comment (langchain-ai/docs#2651)
The issue is that mint broken-links checks snippets and doesn't actually first build the docs. So the best we can do is filter those false positives out. Which leaves us open to hidden broken links. Maybe I'll implement a secondary link test with htmltest at some point

PR #2679, opened 2026-02-17
Document integration patterns for agents with sandboxes

npentrel:
Added integration patterns for agents with sandboxes, detailing two architecture patterns: 'Agent in sandbox' and 'Sandbox as tool'. Included benefits and trade-offs for each pattern.

### DIFF
--- a/src/oss/deepagents/sandboxes.mdx
+++ b/src/oss/deepagents/sandboxes.mdx
@@ -142,30 +142,30 @@ import { createDeepAgent } from "deepagents";
 const sandbox = await DaytonaSandbox.create();
 
 const agent = createDeepAgent({
-backend: sandbox,
-systemPrompt:
+  backend: sandbox,
+  systemPrompt:
     "You are a coding assistant with sandbox access. You can create and run code in the sandbox.",
 });
 
 try {
-const result = await agent.invoke({
+  const result = await agent.invoke({
     messages: [
-    {
+      {
         role: "user",
         content: "Create a hello world Python script and run it",
-    },
+      },
     ],
-});
-const lastMessage = result.messages[result.messages.length - 1];
-console.log(
+  });
+  const lastMessage = result.messages[result.messages.length - 1];
+  console.log(
     typeof lastMessage.content === "string"
-    ? lastMessage.content
-    : String(lastMessage.content),
-);
-} catch (err) {
-// Optional: delete the sandbox proactively on an exception
-await sandbox.close();
-throw err;
+      ? lastMessage.content
+      : String(lastMessage.content),
+  );
+} finally {
+  // Optional: delete the sandbox proactively when the agent finishes
+  await sandbox.close();
+  throw err;
 }
 ```
 


src/oss/deepagents/sandboxes.mdx
```diff
@@ -38,6 +38,47 @@ graph LR
...
+
+Trade-offs:
+
+- API keys must live inside the sandbox (security risk).
+- Updates require rebuilding images.
+- Requires infrastructure for communication (WebSocket or HTTP layer).
+
+**Sandbox as tool pattern**
+
+The agent runs on your machine or server. When it needs to execute code, it calls a remote sandbox via the provider's API.
```
hntrl: We should illustrate that the agent does this through tool calling the sandbox directly which uses the provider APIs

src/oss/deepagents/sandboxes.mdx
```diff
@@ -38,6 +38,47 @@ graph LR
...
+- Pay only for execution time.
+
+Trade-offs:
+
+- Network latency on each execution call.
+
+The examples in this doc use the sandbox as a tool pattern.
+Choose the agent in sandbox pattern when your provider's SDK handles the communication layer and you want production to mirror local development.
+Choose the sandbox as tool pattern when you need to iterate quickly on agent logic, keep API keys outside the sandbox, or prefer cleaner separation of concerns.
+For a deeper dive with trade-offs and implementation details, see [The two patterns by which agents connect sandboxes](https://blog.langchain.com/the-two-patterns-by-which-agents-connect-sandboxes/).
```
hntrl: does it make sense to backlink to blog here? I almost think we should be pulling out all the relevant concepts and inlining them here instead. E.g. docs are meant to an authoritative source and our blogs are more positioned as "op-eds"

src/oss/deepagents/sandboxes.mdx
```diff
@@ -38,6 +38,47 @@ graph LR
...
+Benefits:
+
+- Mirrors local development closely.
+- Tight coupling between agent and environment.
+
+Trade-offs:
+
+- API keys must live inside the sandbox (security risk).
+- Updates require rebuilding images.
+- Requires infrastructure for communication (WebSocket or HTTP layer).
```
hntrl: I like the visual candy we have the interrupt docs https://docs.langchain.com/oss/python/langgraph/interrupts#do-not-wrap-interrupt-calls-in-try/except

```suggestion
Benefits:

- ✅ Mirrors local development closely.
- ✅ Tight coupling between agent and environment.

Trade-offs:

- 🔴 API keys must live inside the sandbox (security risk).
- 🔴 Updates require rebuilding images.
- 🔴 Requires infrastructure for communication (WebSocket or HTTP layer).
```

src/oss/deepagents/sandboxes.mdx
npentrel: Do you mean in an example or by replacing this text? 

```suggestion
The agent runs on your machine or server. When it needs to execute code, it calls sandbox tools (such as `execute`, `read_file`, or `write_file`) which invoke the provider's APIs to run operations in a remote sandbox.
```

hntrl: ya this is perfect

src/oss/deepagents/sandboxes.mdx
```diff
@@ -42,6 +42,141 @@ Sandboxes are especially useful for:
...
+console.log(
+    typeof lastMessage.content === "string"
+    ? lastMessage.content
+    : String(lastMessage.content),
+);
+} catch (err) {
+// Optional: delete the sandbox proactively on an exception
+await sandbox.close();
+throw err;
+}
```
hntrl, line 169:
```suggestion
const agent = createDeepAgent({
  backend: sandbox,
  systemPrompt:
    "You are a coding assistant with sandbox access. You can create and run code in the sandbox.",
});

try {
  const result = await agent.invoke({
    messages: [
      {
        role: "user",
        content: "Create a hello world Python script and run it",
      },
    ],
  });
  const lastMessage = result.messages[result.messages.length - 1];
  console.log(
    typeof lastMessage.content === "string"
      ? lastMessage.content
      : String(lastMessage.content),
  );
} finally {
  // Optional: delete the sandbox proactively when the agent finishes
  await sandbox.close();
  throw err;
}
```

PR #2682, opened 2026-02-17
Update the supported ttl strategies

katmayb:
Fixes DOC-735

Adds `keep_latest` to TTL strategy configuration

## Preview

https://langchain-5e9cc07a-preview-ttlstr-1771345811-5c31d57.mintlify.app/langsmith/configure-ttl#configuring-thread-and-checkpoint-ttl

### DIFF
--- a/src/langsmith/configure-ttl.mdx
+++ b/src/langsmith/configure-ttl.mdx
@@ -33,7 +33,7 @@ Add a `checkpointer.ttl` configuration to your `langgraph.json` file:
 ```
 
 - `strategy`: Specifies the action taken on expiration.
-    - `"delete"`: Removes the entire thread (all associated run and checkpoint data) when the TTL expires.
+    - `"delete"`: Removes the entire thread including all associated run and checkpoint data when the TTL expires.
     - `"keep_latest"`: Retains the thread and latest checkpoint, but deletes older checkpoint data that subsequent runs won't need.
 - `sweep_interval_minutes`: Defines how often, in minutes, the system checks for expired checkpoints.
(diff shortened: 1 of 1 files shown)


src/langsmith/configure-ttl.mdx
```diff
@@ -32,7 +32,15 @@ Add a `checkpointer.ttl` configuration to your `langgraph.json` file:
 }
 ```
 
-- `strategy`: Specifies the action taken on expiration. Only `"delete"` is supported, which deletes all checkpoints in the thread upon expiration.
+- `strategy`: Specifies the action taken on expiration.
+    - `"delete"`: Removes the entire thread and all associated data when the TTL expires, including:
+        - All checkpoint data
+        - Associated runs
+        - Thread metadata
```
hinthornw: Doesn't add much

src/langsmith/configure-ttl.mdx
```diff
@@ -32,7 +32,15 @@ Add a `checkpointer.ttl` configuration to your `langgraph.json` file:
...
+    - `"keep_latest"`: Deletes historical checkpoints when the TTL expires but preserves:
+        - The thread itself
+        - The most recent checkpoint
+        - Thread metadata
```
jdrogers940: Can we maybe make these more parallel for the list? We could combine the thread itself and metadata into one for both?
```suggestion
        - The most recent checkpoint
        - Associated runs
        - Thread metadata
```

src/langsmith/configure-ttl.mdx
katmayb: My thought was to keep the descriptions explicit to avoid questions of what is/isn't included here. I'm OK to remove the sub-bullets if we think it's too much

PR #2701, opened 2026-02-18
add docs for resume multiple interrupts

npentrel:


src/oss/langgraph/interrupts.mdx
```diff
@@ -130,6 +128,7 @@ await graph.invoke(new Command({ resume: true }), config);
 The key thing that interrupts unlock is the ability to pause execution and wait for external input. This is useful for a variety of use cases, including:
 
 - <Icon icon="circle-check" /> [Approval workflows](#approve-or-reject): Pause before executing critical actions (API calls, database changes, financial transactions)
+- <Icon icon="link" /> [Map interrupt IDs to resume values](#map-interrupt-ids-to-resume-values): Pair interrupt IDs with resume values when resuming multiple interrupts in a single invocation
```
sydney-runkle: maybe something like "resuming multiple interrupts" or "handling multiple interrupts"

npentrel:
> a bit misleading on my part -- let's provide a more simple example where there's 2 nodes in parallel that each fire an interrupt, then we resume both with one invoke call
> 
> we used to have this in the old docs :/

How do you feel about something like this? 

```
import asyncio
from typing import List
import uuid

from langchain_core.runnables import RunnableConfig
from langgraph.checkpoint.memory import InMemorySaver
from langgraph.func import entrypoint, task
from langgraph.types import Command, Interrupt, StateSnapshot, interrupt


def get_pending_interrupts(state: StateSnapshot) -> List[Interrupt]:
    pending_interrupts = []
    for state_task in state.tasks:
        if not state_task.result:
            for i in state_task.interrupts:
                pending_interrupts.append(i)
    return pending_interrupts


def ask_pending_hitl_questions(state: StateSnapshot) -> dict[str, str] | None:
    resume_map = {}
    for pending in get_pending_interrupts(state):
        answer = input(f"Q: {pending.value} ")
        resume_map[pending.id] = answer
    return resume_map or None


@task
async def ask_a() -> str:
    return interrupt("Question from node A?")


@task
async def ask_b() -> str:
    return interrupt("Question from node B?")


@entrypoint(checkpointer=InMemorySaver())
async def workflow(_input: dict | None = None) -> dict:
    # 2 nodes in parallel, each fires an interrupt
    a_result, b_result = await asyncio.gather(ask_a(), ask_b())
    return {"a": a_result, "b": b_result}


def get_interrupts_from_result(result: dict) -> list[Interrupt]:
    """Extract interrupts from an invoke/stream result."""
    interrupts = result.get("__interrupt__")
    if interrupts is None:
        return []
    return list(interrupts) if isinstance(interrupts, (list, tuple)) else [interrupts]


async def main():
    thread_id = uuid.uuid4().hex
    config = RunnableConfig(configurable={"thread_id": thread_id})

    # Run until first interrupt
    async for e in workflow.astream({}, config=config):
        print(f"===> {e}")

    # Loop: resume all pending interrupts until the workflow completes
    while True:
        state = await workflow.aget_state(config=config)
        resume_map = ask_pending_hitl_questions(state)
        if not resume_map:
            break
        result = await workflow.ainvoke(Command(resume=resume_map), config=config)
        more = get_interrupts_from_result(result)
        if not more:
            print(f"Done: {result}")
            break
        print(f"More interrupts to handle: {more}")


if __name__ == "__main__":
    asyncio.run(main())

```

src/oss/langgraph/interrupts.mdx
```diff
@@ -180,6 +179,157 @@ async for metadata, mode, chunk in graph.astream(
...
 
+### Handling multiple interrupts
+
+When resuming multiple interrupts with a single invocation, map each interrupt ID to its resume value.
+This ensures each response is paired with the correct interrupt at runtime.
+
+:::python
+
+```python
+"""Minimal example: resuming multiple parallel interrupts."""
```
sydney-runkle:
```suggestion
```

PR #2707, opened 2026-02-18
docs: Add reference to LocalShellBackend

rk-yen:
## Overview
Adding documentation for `LocalShellBackend`

## Type of change

## Related issues/PRs
- https://github.com/langchain-ai/deepagents/pull/930

## Checklist


npentrel:
Thank you for the PR. I've reviewed and unfortunately we won't be merging this. The docs are intended as is in this case because the localshell backend is not a Sandbox, so adding it to that list would be confusing.

PR #2793, opened 2026-02-24
Remove index-datasets-for-dynamic-few-shot-example file

katmayb:


src/docs.json
```diff
@@ -2206,6 +2205,10 @@
     {
       "source": "/oss/javascript/integrations/chat/nvidia_ai_endpoints",
       "destination": "/oss/javascript/integrations/chat"
+    },
+    {
+      "source": "/langsmith/index-datasets-for-dynamic-few-shot-example-selection",
+      "destination": "/langsmith/manage-datasets"
```
katmayb, line 2214: @catherine-langchain There might be a better page to redirect to? I've just used the Manage datasets page right now?

catherine-langchain: yes! i think manage datasets are fine

PR #2815, opened 2026-02-26
docs: warn that @auth.on filter fields must be left unencrypted

cwlbraa:
Adds a callout in the custom encryption docs noting that fields used as auth filters must stay unencrypted — otherwise the stored encrypted value won't match the plaintext filter and auth isolation silently breaks. LangGraph Server (as of 0.7.57) will return a 422 at write time if it detects this.

### DIFF
--- a/src/langsmith/encryption.mdx
+++ b/src/langsmith/encryption.mdx
@@ -197,7 +197,7 @@ async def decrypt_json(ctx: EncryptionContext, data: dict) -> dict:
 
 Common fields to leave **unencrypted** for search and filtering:
 
-- Fields used as filters in your [`@auth.on` handlers](./custom_auth) (e.g., `owner`, `tenant_id`) — if these are encrypted, auth isolation silently breaks because the stored value no longer matches the plaintext filter. LangGraph Server will return a 422 error if it detects this at write time.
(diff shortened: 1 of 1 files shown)


src/langsmith/encryption.mdx
```diff
@@ -212,7 +197,8 @@ async def decrypt_json(ctx: EncryptionContext, data: dict) -> dict:
 
 Common fields to leave **unencrypted** for search and filtering:
 
-- User-defined fields for access control queries (e.g., `tenant_id`, `owner`)
+- Fields used as filters in your [`@auth.on` handlers](./custom_auth) (e.g., `owner`, `tenant_id`) — if these are encrypted, auth isolation silently breaks because the stored value no longer matches the plaintext filter. LangGraph Server will return a 422 error if it detects this at write time.
```
katmayb:
```suggestion
- Fields used as filters in your [`@auth.on` handlers](/langsmith/set-up-custom-auth) (e.g., `owner`, `tenant_id`) — if these are encrypted, auth isolation silently breaks because the stored value no longer matches the plaintext filter. LangGraph Server will return a 422 error if it detects this at write time.
```

Is that the page you wanted to link to?

cwlbraa: yep, sorry! was gonna check the links before merging i swear

PR #2822, opened 2026-02-26
fix: add redis sharing warning [closes DOC-782]

lnhsingh:
## Description
Clarifies that each LangSmith installation needs a dedicated Redis instance to prevent deployment tasks from routing to the wrong cluster, especially during migrations. Adds a warning callout to the external Redis setup doc.

AI agent assisted with drafting this change.

## Test Plan
- [ ] `make -C /workspace/docs format` (fails: ruff check errors in pipeline/core/builder.py, pipeline/preprocessors/handle_auto_links.py, reference/python/serve_subset.py, scripts/check_removed_pages_redirects.py)
- [ ] `make -C /workspace/docs lint`

### DIFF
--- a/src/langsmith/self-host-external-redis.mdx
+++ b/src/langsmith/self-host-external-redis.mdx
@@ -6,7 +6,7 @@ sidebarTitle: Connect to an external Redis database
 LangSmith uses Redis to back our queuing/caching operations. By default, LangSmith Self-Hosted will use an internal Redis instance. However, you can configure LangSmith to use an external Redis instance. By configuring an external Redis instance, you can more easily manage backups, scaling, and other operational tasks for your Redis instance.
 
 <Warning>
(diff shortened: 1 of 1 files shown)


src/langsmith/self-host-external-redis.mdx
```diff
@@ -5,6 +5,10 @@ sidebarTitle: Connect to an external Redis database
 
 LangSmith uses Redis to back our queuing/caching operations. By default, LangSmith Self-Hosted will use an internal Redis instance. However, you can configure LangSmith to use an external Redis instance. By configuring an external Redis instance, you can more easily manage backups, scaling, and other operational tasks for your Redis instance.
 
+<Warning>
+Each LangSmith installation must use its own dedicated Redis instance. Redis cannot be shared across separate LangSmith installations (for example, between an old and new cluster during a migration). Sharing it across installations causes deployment tasks to be routed to the wrong cluster.
```
katmayb:
```suggestion
Each LangSmith installation must use its own dedicated Redis instance. Redis cannot be shared across separate LangSmith installations (for example, between an existing and new cluster during a migration). Sharing it across installations causes deployment tasks to be routed to the wrong cluster.
```

PR #2936, opened 2026-03-05
feat: add streaming thinking/reasoning tokens example

npentrel:
- Added a new "Streaming thinking / reasoning tokens" subsection to the Common patterns section of the streaming overview page
- Added python and javascript code samples

https://langchain-5e9cc07a-preview-opensw-1772748614-696205f.mintlify.app/oss/javascript/langchain/streaming/overview#streaming-thinking-%2F-reasoning-tokens

### DIFF
--- a/src/oss/langchain/streaming/overview.mdx
+++ b/src/oss/langchain/streaming/overview.mdx
@@ -18,7 +18,7 @@ What's possible with LangChain streaming:
 
 * <Icon icon="brain" size={16} /> [**Stream agent progress**](#agent-progress) — get state updates after each agent step.
 * <Icon icon="binary" size={16} /> [**Stream LLM tokens**](#llm-tokens) — stream language model tokens as they're generated.
-* <Icon icon="bulb" size={16} /> [**Stream thinking / reasoning tokens**](#streaming-thinking--reasoning-tokens) — surface model reasoning as it's generated.
+* <Icon icon="bulb" size={16} /> [**Stream thinking / reasoning tokens**](#streaming-thinking-/-reasoning-tokens) — surface model reasoning as it's generated.
 * <Icon icon="table" size={16} /> [**Stream custom updates**](#custom-updates) — emit user-defined signals (e.g., `"Fetched 10/100 records"`).
 * <Icon icon="stack-push" size={16} /> [**Stream multiple modes**](#stream-multiple-modes) — choose from `updates` (agent progress), `messages` (LLM tokens + metadata), or `custom` (arbitrary user data).
 
@@ -528,6 +528,8 @@ Some models perform internal reasoning before producing a final answer. You can 
     Reasoning output must be enabled on the model.
 
     See the [reasoning section](/oss/langchain/models#reasoning) and your [provider's integration page](/oss/integrations/providers/overview) for configuration details.
+
+    To quickly check a model's reasoning support, see [models.dev](https://models.dev).
 </Note>
 
 :::python


src/oss/langchain/streaming/overview.mdx
```diff
@@ -15,6 +18,7 @@ What's possible with LangChain streaming:
 
 * <Icon icon="brain" size={16} /> [**Stream agent progress**](#agent-progress) — get state updates after each agent step.
 * <Icon icon="binary" size={16} /> [**Stream LLM tokens**](#llm-tokens) — stream language model tokens as they're generated.
+* <Icon icon="bulb" size={16} /> [**Stream thinking / reasoning tokens**](#streaming-thinking--reasoning-tokens) — surface model reasoning as it's generated.
```
mdrxy: link is broken

mdrxy: at least on local dev

src/oss/langchain/streaming/overview.mdx
```diff
@@ -512,11 +516,54 @@ for await (const [streamMode, chunk] of await agent.stream(
...
 
-:::python
 ## Common patterns
 
 Below are examples showing common use cases for streaming.
 
+### Streaming thinking / reasoning tokens
+
+Some models perform internal reasoning before producing a final answer.
+You can stream these thinking / reasoning tokens as they're generated by filtering [standard content blocks](/oss/langchain/messages#standard-content-blocks) for the `type` `"reasoning"`.
```
mdrxy: `type = "reasoning"` instead? weird having two side by side code formatted words

src/oss/langchain/streaming/overview.mdx
```diff
@@ -512,11 +516,54 @@ for await (const [streamMode, chunk] of await agent.stream(
...
 
+### Streaming thinking / reasoning tokens
+
+Some models perform internal reasoning before producing a final answer. You can stream these thinking / reasoning tokens as they're generated by filtering [standard content blocks](/oss/langchain/messages#standard-content-blocks) for the `type` `"reasoning"`.
+
+<Note>
+    Reasoning output must be enabled on the model.
+
+    See the [reasoning section](/oss/langchain/models#reasoning) and your [provider's integration page](/oss/integrations/providers/overview) for configuration details.
+</Note>
```
mdrxy, line 533: should we callout to models.dev as a resource to quickly check a given model's reasoning support? e.g. making it easy to find out if a model supports reasoning in the first place

src/snippets/code-samples/streaming-reasoning-tokens-py.mdx
```diff
@@ -0,0 +1,36 @@
+```python
```
mdrxy, line 1: why is this file duplicated from src/code-samples?

npentrel: It's a new setup we're trying, I want the code samples to be continuously tested. The code-samples in src/code-samples are tested on change and every week, that way if there are API changes and the code samples there break, we'll know we need to fix something. Then there's a tool + script that extracts only the relevant bits for the snippets we want to show in the docs (leaving out some setup and teardown and tests if applicable) and then puts that into an mdx file which is what then gets included in mintlify. I know it's duplicating them, but these mdx snippets are autogenerated - so it's only the initial python and typescript files that we need to care about. I know we could not check these into the code too and have them built as part of the build process but it would lengthen the build process, especially once there are hundreds of these - and these files shouldn't change often. So I think it's worth the tradeoff

src/oss/langchain/streaming/overview.mdx
npentrel: I wish I could trust our link checking :)

npentrel: thanks!

PR #2947, opened 2026-03-05
Clean up navigation in get started tab langsmith

katmayb:
https://langchain-5e9cc07a-preview-additi-1772741372-d15a08d.mintlify.app/langsmith/home

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -226,14 +226,14 @@
                       ]
                     },
                     {
-                      "group": "FAQ & support",
+                      "group": "FAQ",
                       "pages": [
                         "langsmith/faq",
                         "langsmith/regions-faq",
-                        "langsmith/pricing-faq",
-                        "langsmith/status"
+                        "langsmith/pricing-faq"
                       ]
-                    }
+                    },
+                    "langsmith/status"
                   ]
                 }
               ]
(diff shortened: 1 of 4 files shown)


src/docs.json
```diff
@@ -177,43 +177,63 @@
...
                   ]
                 },
                 {
-                  "group": "Additional resources",
+                  "group": "Tools",
                   "pages": [
                     "langsmith/polly",
                     "langsmith/langsmith-cli",
-                    "langsmith/skills",
+                    "langsmith/skills"
```
lnhsingh, line 214: Should this page be titled "Agent skills" or "Skills"? I feel like we should reserve "LangSmith + [feature name]" for the big ones

src/docs.json
```diff
@@ -177,43 +177,63 @@
...
-                         "langsmith/authentication-methods",
-                         "langsmith/jit-invite-sso"
-                       ]
-                     },
-                    "langsmith/scalability-and-resilience",
-                    "langsmith/faq",
-                    "langsmith/regions-faq",
-                    "langsmith/pricing-faq",
-                    "langsmith/status"
+                      "group": "FAQ & support",
```
lnhsingh: I think this would be cleaner as "FAQ" and LangSmith status is on its own

PR #2988, opened 2026-03-08
feat(js): add docs for deepagents-acp in JS

christian-bromann:
Since we now also support ACP in JS, let's document it.


src/oss/deepagents/acp.mdx
```diff
@@ -118,6 +254,392 @@ toad acp "python path/to/your_server.py" .
...
+
+:::
+
+:::js
+
+## Features
+
+### Multiple agents
+
+When you define multiple agents, the client selects which one to use at session creation time by passing `configOptions.agent` in the `session/new` ACP request. If not specified, the first agent is used by default.
```
npentrel:
```suggestion
When you define multiple agents, the client selects which one to use at session creation time by passing `configOptions.agent` in the `session/new` ACP request. If not specified, the client uses the first agent by default.
```

src/oss/deepagents/acp.mdx
```diff
@@ -118,6 +254,392 @@ toad acp "python path/to/your_server.py" .
...
+When using models with extended thinking (e.g., Claude with `thinking: { type: "enabled" }`), the server streams reasoning tokens to the IDE as `thought_message_chunk` updates. This gives users visibility into the agent's chain-of-thought process in clients that support it.
+
+### Tool call enhancements
+
+The server provides rich tool call reporting to the IDE:
+
+- **Tool call kinds** — each tool call is categorized using [ACP-standard kinds](https://agentclientprotocol.com/protocol/tool-calls) (`read`, `edit`, `search`, `execute`, `think`, etc.) so the IDE can display appropriate icons
+- **File locations (follow-along)** — tool calls that operate on files (e.g., `read_file`, `edit_file`, `grep`) report `{ path, line }` locations, enabling IDEs to open and highlight files the agent is working with in real time
+- **Diff content** — when the agent edits a file, the tool call update includes diff content so the IDE can render inline diffs
+- **Raw input/output** — tool call notifications include the raw tool arguments and results for transparency
```
npentrel: is this needed? Feels like possibly implementation details? If I wanted to use this I'd just expect all of this wouldn't I?

npentrel (commented):
I am a little unsure who the audience is here - is it people using deepagents who want to configure it in their IDE or people developing with it?

christian-bromann:
Thanks for the review! Addressed both points:

- Refocused the page on developers building with `deepagents-acp` who want to expose their agents to IDEs. The structure now follows a clear progression: install → quickstart → connect to an IDE → CLI reference → programmatic API → customization options.
- Removed the sections that describe behavior users don't configure and would just expect to work. The remaining "Customization" section now only covers things the developer actively configures: multiple agents, slash commands, human-in-the-loop, custom tools, custom backends, and skills/memory.

Let me know your thoughts

PR #2992, opened 2026-03-08
Add external integration: langchain-aro (execution integrity artifacts)

joy7758:
This PR adds an external integration page for langchain-aro.

langchain-aro is a small middleware-first project that captures LangChain run events, exports a portable artifact, and verifies the artifact via execution hashing.

The integration example shows how to attach an ARORecorder callback handler to a LangChain runnable, export a run artifact, and verify it with verify_artifact().

The project focuses on execution integrity artifacts rather than tracing or observability replacement.

The current MVP is experimental and has been validated against a real langchain-core callback flow.


npentrel (approved):

PR #3008, opened 2026-03-09
Add ABAC section on self-hosted enablement

katmayb:
Fixes DOC-843

## Preview

https://langchain-5e9cc07a-preview-abacse-1773079806-e643a6d.mintlify.app/langsmith/abac#enable-abac-for-self-hosted-deployments


src/langsmith/abac.mdx
```diff
@@ -20,6 +20,35 @@ Roles and resource tags can be managed via the UI or API. ABAC policies are curr
...
+        ```
+
+    - **Enable for all organizations:** Add the following environment variable to `commonEnv` in your `values.yaml`:
+
+        ```yaml
+        DEFAULT_ORG_FEATURE_CAN_USE_ABAC: "true"
+        ```
+
+        <Note>
+        This environment variable has no effect on personal organizations, which do not have [RBAC](/langsmith/rbac) enabled.
```
katmayb: @bvs-langchain Is this trying to say this has no effect on personal organizations because RBAC isn't enabled. Or, this has no effect on personal organizations that don't have RBAC enabled?

This is just for Enterprise, so assuming the former?

bvs-langchain: yes the former. and we should update the sql to say `... where not is_personal` to match this fact (even though it would have no effect for personal orgs, better to avoid doubt).

PR #3042, opened 2026-03-10
docs: adding guide for cancelling runs on LSD

jdrogers940:
## Overview
Adding docs for the cancel and bulk cancel APIs on LSD. Provides examples of using the various methods with the relevant parameters and reasons why users might want to do so. Also links out to the various other docs and guides we have today about why cancellation is useful.

## Type of change

**Type:** New documentation page

## Related issues/PRs

- Slack thread: https://langchain.slack.com/archives/C04F8K3HAR1/p1773169736380269?thread_ts=1773080463.567999&cid=C04F8K3HAR1

## Checklist

- [X] I have read the [contributing guidelines](README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)
- [X] All code examples have been tested and work correctly
- [X] I have used **root relative** paths for internal links
- [X] I have updated navigation in `src/docs.json` if needed

### DIFF
--- a/src/langsmith/cancel-run.mdx
+++ b/src/langsmith/cancel-run.mdx
@@ -3,7 +3,7 @@ title: How to cancel a run
 sidebarTitle: Cancel a run
 description: Cancel a single run or multiple runs via the API, and choose between interrupt and rollback actions.
 ---
-This guide covers how to cancel runs for your agent via the LangSmith Deployment API. You can cancel a single run by ID or cancel multiple runs by thread or status. Cancellation is useful for stopping long-running or stuck runs, or when a user abandons a request.
+This guide covers how to cancel runs for your agent via the [LangSmith Deployment API](/langsmith/server-api-ref). You can cancel a single run by ID or cancel multiple runs by thread or status. Cancellation is useful for stopping long-running or stuck runs, or when a user abandons a request.
 
 ## Setup
 
@@ -62,7 +62,7 @@ Use **interrupt** when you want to stop a run but keep it for debugging, auditin
     )
     await client.runs.cancel(thread["thread_id"], run["run_id"])
 
-    run_after = await client.runs.get(thread["thread_id"], run["run_id"])
+    run_after = await client.runs.get(thread["thread_id"], run["run_id"], wait=True)
     print(run_after["status"])   # "interrupted"
     ```
     </Tab>
@@ -73,7 +73,7 @@ Use **interrupt** when you want to stop a run but keep it for debugging, auditin
         assistantID,
         { input: { messages: [{ role: "user", content: "Long task" }] } }
     );
-    await client.runs.cancel(thread["thread_id"], run["run_id"]);
+    await client.runs.cancel(thread["thread_id"], run["run_id"], wait=true);
 
     const runAfter = await client.runs.get(thread["thread_id"], run["run_id"]);
     console.log(runAfter["status"]);   // "interrupted"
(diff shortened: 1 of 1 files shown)


src/langsmith/cancel-run.mdx
```diff
@@ -0,0 +1,308 @@
...
+    # Throws an error because the run is deleted
+    curl --request GET \
+      --url <DEPLOYMENT_URL>/threads/<THREAD_ID>/runs/<RUN_ID>
+    ```
+    </Tab>
+</Tabs>
+
+### Cancel with wait
+
+By default, the cancel request returns after the cancellation is requested and the run is cancelled asynchronously. `wait=True` makes the cancel request block until the run has been fully cancelled. This is useful when you want to know the final state of the run after it has been cancelled (e.g. what checkpoints were created, what the final output was, etc.).
```
katmayb:
```suggestion
By default, the cancel request returns after the cancellation is requested and the run is cancelled asynchronously. `wait=True` makes the cancel request block until the run has been fully cancelled. This is useful when you want to know the final state of the run after it has been cancelled (e.g., what checkpoints were created, what the final output was).
```

src/langsmith/cancel-run.mdx
```diff
@@ -0,0 +1,308 @@
...
+      --url "<DEPLOYMENT_URL>/runs/cancel?action=interrupt" \
+      --header 'Content-Type: application/json' \
+      --data '{"thread_id": "<THREAD_ID>", "status": "running"}'
+    ```
+    </Tab>
+</Tabs>
+
+## Cancel on disconnect
+
+When starting a run with streaming, you can set `cancel_on_disconnect` so that the run is cancelled if the client disconnects. This avoids leaving runs in progress when a user closes the app or loses connection. See the stream/connect options in the API you use (for example, `client.runs.stream` or `client.runs.wait`) for the exact parameter name and behavior.
```
katmayb: The final sentence here is a little confusing, want to break into two sentences, one linking to the API options and then the sentence re using the exact parameter name?

katmayb: Or something like this, if it is what you intended?

```suggestion
When starting a run with streaming, you can set `cancel_on_disconnect` so that the run is cancelled if the client disconnects. This avoids leaving runs in progress when a user closes the app or loses connection. For the exact parameter name and behavior, see the stream or connect options for your API method (for example, `client.runs.stream` or `client.runs.wait`).
```

src/langsmith/cancel-run.mdx
```diff
@@ -0,0 +1,308 @@
...
+    # Get the final state of the thread
+    curl --request GET \
+      --url <DEPLOYMENT_URL>/threads/<THREAD_ID>/state
+    ```
+    </Tab>
+</Tabs>
+
+## Cancel multiple runs
+
+The bulk cancel endpoint can be used to cancel multiple runs by thread ID and (optionally) run IDs or status. `status` filter (`pending`, `running`, or `all`).
```
katmayb:
```suggestion
Use the bulk cancel endpoint to cancel multiple runs by thread ID and (optionally) run IDs or status. `status` filter (`pending`, `running`, or `all`).
```

src/langsmith/cancel-run.mdx
```diff
@@ -0,0 +1,308 @@
+---
+title: How to cancel a run
+sidebarTitle: Cancel a run
+description: Cancel a single run or multiple runs via the API, and choose between interrupt and rollback actions.
+---
+This guide covers how to cancel runs for your agent via the LangSmith Deployment API. You can cancel a single run by ID or cancel multiple runs by thread or status. Cancellation is useful for stopping long-running or stuck runs, or when a user abandons a request.
```
katmayb:
```suggestion
This guide covers how to cancel runs for your agent via the [LangSmith Deployment API](/langsmith/server-api-ref). You can cancel a single run by ID or cancel multiple runs by thread or status. Cancellation is useful for stopping long-running or stuck runs, or when a user abandons a request.
```

Is it the LangSmith Deployment Agent Server API ref we want to link to here?

src/langsmith/cancel-run.mdx
jdrogers940: Yeah good callout! I'll just add examples here too showing the API parameter being used. No reason to point to the other page.

PR #3191, opened 2026-03-18
Improve type hints for llm_call and tool_node functions

LiTianyu11:
## Overview
Improve type hints for llm_call and tool_node functions


npentrel (approved):
Thanks!

PR #3196, opened 2026-03-18
feat(frontend): document OpenUI integration

christian-bromann:
Adding gen-ui example using OpenUI

Screenshot 2026-03-18 at 11 33 25 AM

### DIFF
--- a/src/oss/langchain/frontend/overview.mdx
+++ b/src/oss/langchain/frontend/overview.mdx
@@ -15,8 +15,7 @@ Every pattern follows the same architecture: a `createAgent` backend streams sta
     "fontFamily": "monospace",
     "flowchart": {
       "curve": "curve"
-    },
-    "themeVariables": {"edgeLabelBackground": "#ffffff"}
+    }
   }
 }%%
 graph LR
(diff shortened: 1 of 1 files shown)


src/oss/langchain/frontend/overview.mdx
```diff
@@ -35,61 +36,19 @@ On the backend, `createAgent` produces a compiled LangGraph graph that exposes a
 
 <CodeGroup>
 
-:::python
```
npentrel, line 38: why is this going?

src/oss/langchain/frontend/integrations/openui.mdx
```diff
@@ -0,0 +1,509 @@
...
+s1   = Series("Revenue", [286, 198, 147])
+ch1  = BarChart(["Gen AI", "ML Infra", "Vision"], [s1])
+marketSection = Card([CardHeader("Market Breakdown"), tbl, ch1])
+```
+
+With hoisting enabled (recommended), the `root` line is written first so the page structure appears immediately and each section fills in as the model defines it.
+
+## Progressive rendering utilities
+
+Wiring `useStream` to `Renderer` naively — re-rendering on every streaming token — produces hundreds of no-op re-parses per response and causes chart components to crash when their data hasn't arrived yet. The utilities below solve three specific problems:
```
npentrel: I don't like the word naive for docs, if someone does this by mistake they shouldn't be made to feel stupid. 
And then since we don't specify the 3 problems, let's keep it general 
```suggestion
Wiring `useStream` to `Renderer` directly causes results in re-rendering on every streaming token and produces hundreds of no-op re-parses per response. This causes chart components to crash when their data hasn't arrived yet. The utilities below solve these problems:
```

npentrel (commented):
Two minor comments

PR #3227, opened 2026-03-20
docs(deepagents): add async subagents documentation

hntrl:
Add documentation for async subagents in deepagents 0.5.0


eyurtsev:
@hntrl I updated the chart in the introduction to be a bit more precise. Most of the doc looks pretty good. I'll update SDK tools right now to make the doc ones (which are better)

PR #3283, opened 2026-03-24
Add docs ordering rules with webhooks + evaluators

katmayb:
Fixes DOC-487

- webhooks: new "Ensuring evaluations complete before the webhook fires" section with the worked example, filter syntax tip, and the within-rule ordering note
- rules: new "How automation rules execute" section explaining independent scheduling and the fixed within-rule action order, with a link to the webhook doc
- online evaluations: tip pointing users to the feedback filter pattern


src/langsmith/rules.mdx
```diff
@@ -17,6 +17,19 @@ To configure online evaluations, visit the [online evaluations](/langsmith/onlin
 
 <Note>If an automation rule matches any run within a trace, the trace will be auto-upgraded to [extended data retention](/langsmith/administration-overview#data-retention-auto-upgrades). This upgrade will impact trace pricing, but ensures that traces meeting your automation criteria (typically those most valuable for analysis) are preserved for investigation. </Note>
 
+## How automation rules execute
+
+Each automation rule runs on an independent polling schedule. If you have multiple rules on the same project, a webhook rule may process a run before an evaluator rule has scored it, or vice versa.
+
+Within a single rule, if multiple actions are configured, they execute in this order:
```
tanushree-sharma, line 24: I'm not totally sure about this order, but I assume you've checked this

tanushree-sharma (dismissed):

src/langsmith/rules.mdx
```diff
@@ -17,6 +17,19 @@ To configure online evaluations, visit the [online evaluations](/langsmith/onlin
...
+
+1. Add to annotation queue.
+1. Add to dataset.
+1. Trigger webhook.
+1. Run online evaluator.
```
katmayb, line 29:
```suggestion
1. Add to annotation queue.
1. Add to dataset.
1. Trigger webhook.
1. Run online evaluator.
1. Run custom code evaluator.
1. Trigger alert.
```

src/langsmith/rules.mdx
katmayb: I think this is good, I checked the repo: https://github.com/langchain-ai/langchainplus/blob/6b402764a39f76e8e256c399ae3db96ad34e50e6/smith-backend/app/models/runs/rules_apply.py#L1810

However, I was missing two actions at the end, so adding.

PR #3292, opened 2026-03-24
fix: language switching redirects to deep agents overview instead of equivalent page

mdrxy:
## Description
Fixes the language toggle (Python ↔ TypeScript) in the OSS docs sidebar. Switching languages from e.g. `/oss/javascript/integrations/chat/fireworks` was landing on the Deep Agents overview instead of the equivalent Python page.

Root causes: (1) Mintlify re-executes custom JS on each client-side navigation, so the in-memory `previousUrl` was reset to `null` before the redirect logic could use it — fixed by persisting in `sessionStorage`. (2) The click handler selector `[data-dropdown-item]` doesn't exist in current Mintlify HTML — updated to `.nav-dropdown-item`. (3) History API patches were stacking on re-execution — added a `window.__langTogglePatched` guard.

## Test Plan
- [ ] From https://docs.langchain.com/oss/javascript/integrations/chat/fireworks, click "Python" in the language dropdown — should navigate to `/oss/python/integrations/chat/fireworks`
- [ ] From a Python page, click "TypeScript" — should navigate to the equivalent JS page
- [ ] Verify hash anchors are preserved when switching languages

### DIFF
--- a/src/language-toggle.js
+++ b/src/language-toggle.js
@@ -7,11 +7,14 @@
  * overview page.
  *
  * How it works:
- * Mintlify's language dropdown renders <a> tags inside a Radix UI portal. Next.js
- * intercepts clicks on these <a> tags for client-side SPA navigation. We use a
- * MutationObserver to detect when the dropdown portal appears, then rewrite the
- * href on each dropdown link to point to the equivalent page in the other language.
- * Next.js then navigates directly to the correct page — no redirect or page refresh.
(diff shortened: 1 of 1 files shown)


mdrxy:
appears to work https://langchain-5e9cc07a-preview-opensw-1774396437-6a33300.mintlify.app/oss/javascript/integrations/chat/fireworks

PR #3313, opened 2026-03-25
docs(deepagents): add doc updates for backend refactor

colifran:
## Overview

We recently worked on a backend refactor for deepagents that adds support for multimodal file types.

### JS
The JS implementation for the refactor was done via a BackendProtocolV2 interface. For sandbox backends we added a SandboxBackendProtocolV2. The refactor is backwards compatible allowing users to continue using legacy file data as well as backends implementing BackendProtocolV1 via a backend protocol adapter implemented internally.

### Python
Updated existing BackendProtocol to use new structured types and added new simplified backend protocol methods.


src/oss/deepagents/backends.mdx
```diff
@@ -358,26 +488,233 @@ class PolicyWrapper(BackendProtocol):
...
+
+`SandboxBackendProtocolV2` extends `BackendProtocolV2` with:
+
+- **`execute(command: string) → ExecuteResponse`** — Run a shell command in the sandbox.
+- **`readonly id: string`** — Unique identifier for the sandbox instance.
+:::
+
+## Update existing backends to V2
+
+:::js
```
hntrl:
```suggestion

## Update existing backends to V2

```

src/oss/deepagents/long-term-memory.mdx
```diff
@@ -476,6 +477,7 @@ const agent = createDeepAgent({
 
 Files stored via `StoreBackend` use the following schema:
 
+:::python
```
npentrel:
```suggestion
:::python
Files stored via `StoreBackend` use the following schema:

```

src/oss/deepagents/long-term-memory.mdx
```diff
@@ -496,12 +497,18 @@ file_data = create_file_data("Hello\nWorld")
 :::
 
 :::js
-```typescript
-import { createFileData } from "deepagents";
+Text files use the following schema:
```
npentrel:
```suggestion
Text files stored via `StoreBackend` use the following schema:
```

src/oss/deepagents/long-term-memory.mdx
npentrel: ^ Otherwise this is a bit awkward for JS

src/oss/deepagents/backends.mdx
```diff
@@ -185,11 +189,25 @@ Use with extreme caution and only in appropriate environments.
...
+- Pass a backend to `createDeepAgent({ backend: ... })`. The filesystem middleware uses it for all tooling.
+- You can pass either:
+    - An instance implementing `AnyBackendProtocol` (`BackendProtocolV1` or `BackendProtocolV2`) — for example, `new FilesystemBackend({ rootDir: "." })`, or
+    - A `BackendFactory` function `(stateAndStore: StateAndStore) => AnyBackendProtocol` — for backends that need runtime state or a store, like `StateBackend` or `StoreBackend`.
+- If omitted, the default is `(config) => new StateBackend(config)`.
+
+<Note>
+V1 backends are automatically adapted to V2 at runtime via `adaptBackendProtocol()`. No code changes are required to keep using existing V1 backends.
+</Note>
```
npentrel, line 336: Reading this before any explanation of v1 vs v2 is confusing. It was previously called `BackendProtocol` right? So maybe this text? I assume I would need to change the name right if I were to use 1.9.0?

```suggestion
<Note>
Before version 1.9.0, only `BackendProtocol` was supported which is now `BackendProtocolV1`. V1 backends are automatically adapted to V2 at runtime via `adaptBackendProtocol()`. No code changes are required to keep using existing V1 backends. To update to v2, see [update existing backends to v2](#update-existing-backends-to-v2).
</Note>
```

src/oss/deepagents/backends.mdx
npentrel: I think actually this? There's no reason this header shoudl show up for python right?
```suggestion
:::

:::js
## Update existing backends to V2

```

colifran: Yeah I purposely put that inside of JS so it doesn't show up for Python since Python doesn't have a V2 backend.

npentrel: you didn't though. The title will still appear. Just do this


```suggestion

## Update existing backends to V2

```

src/oss/deepagents/backends.mdx
```diff
@@ -243,16 +261,27 @@ Build a custom backend to project a remote or database filesystem (e.g., S3 or P
...
+- For external persistence (S3, Postgres, etc.), return `files_update=None` (Python) or omit `filesUpdate` (JS) in write/edit results — only in-memory state backends need to return a files update dict.
+
+:::python
+- Use `ls` and `glob` as the method names.
+- Return structured result types with an `error` field for missing files or invalid patterns (do not raise).
+:::
+
+:::js
+- Use `ls` and `glob` as the method names.
+- All query methods (`ls`, `read`, `readRaw`, `grep`, `glob`) must return structured Result objects (e.g., `LsResult`, `ReadResult`) with an optional `error` field — do not throw.
```
ccurme:
```suggestion
- All query methods (`ls`, `read`, `readRaw`, `grep`, `glob`) must return structured Result objects (e.g., `LsResult`, `ReadResult`) with an optional `error` field.
```
I'm not sure we want to never throw?

src/oss/deepagents/backends.mdx
```diff
@@ -243,16 +261,27 @@ Build a custom backend to project a remote or database filesystem (e.g., S3 or P
 Design guidelines:
 
 - Paths are absolute (`/x/y.txt`). Decide how to map them to your storage keys/rows.
-- Implement `ls_info` and `glob_info` efficiently (server-side listing where available, otherwise local filter).
-- Return user-readable error strings for missing files or invalid regex patterns.
-- For external persistence, set `files_update=None` in results; only in-state backends should return a `files_update` dict.
+- Implement listing and glob efficiently (server-side filtering where available, otherwise local filter).
```
ccurme:
```suggestion
- Implement`ls` and `glob` efficiently (server-side filtering where available, otherwise local filter).
```
? (nit)

src/oss/deepagents/backends.mdx
```diff
@@ -339,14 +469,14 @@ class PolicyWrapper(BackendProtocol):
     def _deny(self, path: str) -> bool:
         return any(path.startswith(p) for p in self.deny_prefixes)
 
-    def ls_info(self, path: str) -> list[FileInfo]:
-        return self.inner.ls_info(path)
-    def read(self, file_path: str, offset: int = 0, limit: int = 2000) -> str:
+    def ls(self, path: str) -> LsResult:
+        return self.inner.ls(path)
+    def read(self, file_path: str, offset: int = 0, limit: int = 2000) -> ReadResult:
```
ccurme, line 602:
```suggestion
        return self.inner.ls(path)

    def read(self, file_path: str, offset: int = 0, limit: int = 2000) -> ReadResult:
```
(nit) consider adding whitespace

src/oss/deepagents/long-term-memory.mdx
ccurme:
```suggestion
"created_at": "2024-01-15T10:30:00Z",       # ISO 8601 timestamp (optional)
"modified_at": "2024-01-15T11:45:00Z"       # ISO 8601 timestamp (optional)
```

src/oss/deepagents/long-term-memory.mdx
```diff
@@ -482,7 +482,8 @@ Files stored via `StoreBackend` use the following schema:
 {
     "content": ["line 1", "line 2", "line 3"],  # List of strings (one per line)
     "created_at": "2024-01-15T10:30:00Z",       # ISO 8601 timestamp
-    "modified_at": "2024-01-15T11:45:00Z"       # ISO 8601 timestamp
+"created_at": "2024-01-15T10:30:00Z",       # ISO 8601 timestamp (optional)
+"modified_at": "2024-01-15T11:45:00Z"       # ISO 8601 timestamp (optional)
```
npentrel: please indent these like the other rows. And I think it's probably not right to have created_at twice

colifran: Updated.

src/oss/deepagents/backends.mdx
```diff
@@ -11,7 +11,11 @@ import BackendComposite from '/snippets/backend-composite.mdx';
 
 Deep Agents expose a filesystem surface to the agent via tools like `ls`, `read_file`, `write_file`, `edit_file`, `glob`, and `grep`. These tools operate through a pluggable backend. The `read_file` tool natively supports image files (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`) across all backends, returning them as multimodal content blocks.
 
-Sandboxes and the @[`LocalShellBackend`] also provide an `execute` tool.
+:::js
+The `read_file` tool natively supports binary files (images, PDFs, audio, video) across all backends, returning a `ReadResult` with typed `content` and `mimeType`.
+:::
+
+Sandboxes and the @[`LocalShellBackend`] also provide and `execute` tool.
```
npentrel:
```suggestion
Sandboxes and the @[`LocalShellBackend`] also provide an `execute` tool.
```

src/oss/deepagents/backends.mdx
```diff
@@ -340,16 +358,27 @@ Build a custom backend to project a remote or database filesystem (e.g., S3 or P
 Design guidelines:
 
 - Paths are absolute (`/x/y.txt`). Decide how to map them to your storage keys/rows.
-- Implement `ls_info` and `glob_info` efficiently (server-side listing where available, otherwise local filter).
-- Return user-readable error strings for missing files or invalid regex patterns.
-- For external persistence, set `files_update=None` in results; only in-state backends should return a `files_update` dict.
+- Implement`ls` and `glob` efficiently (server-side filtering where available, otherwise local filter).
```
npentrel:
```suggestion
- Implement `ls` and `glob` efficiently (server-side filtering where available, otherwise local filter).
```

src/oss/deepagents/harness.mdx
```diff
@@ -93,7 +93,7 @@ For configuration details, see [Context engineering](/oss/deepagents/context-eng
 When you use a [sandbox backend](/oss/deepagents/sandboxes), the harness exposes an `execute` tool that lets the agent run shell commands in an isolated environment. This enables the agent to install dependencies, run scripts, and execute code as part of its task.
 
 **How it works:**
-- Sandbox backends implement @[`SandboxBackendProtocol`] — when detected, the harness adds the `execute` tool to the agent's available tools
+- Sandbox backends implement the `SandboxBackendProtocolV2` — when detected, the harness adds the `execute` tool to the agent's available tools
```
npentrel, line 96: I think this should work?
```suggestion
- Sandbox backends implement the @[`SandboxBackendProtocolV2`] — when detected, the harness adds the `execute` tool to the agent's available tools
```

colifran: I tried to add this previously but I got an invalid cross ref error in the build. Any idea?

PR #3380, opened 2026-03-31
fix(deepagents): add TS translation to customization doc

christian-bromann:
Adding some missing translations.

### DIFF
--- a/src/oss/deepagents/customization.mdx
+++ b/src/oss/deepagents/customization.mdx
@@ -428,7 +428,12 @@ const customMiddleware = createMiddleware({
 ```
 ::::
 
-Mutation in place, such as modifying `self.x` in `before_agent`, mutating a shared variable in `beforeAgent`, or changing other shared values in hooks, can lead to subtle bugs and race conditions because many operations run concurrently (subagents, parallel tools, and parallel invocations on different threads).
+:::python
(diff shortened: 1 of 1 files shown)


src/oss/deepagents/customization.mdx
```diff
@@ -399,8 +413,22 @@ class CustomMiddleware(AgentMiddleware):
...
+  name: "CustomMiddleware",
+  beforeAgent: async () => {
+    x += 1; // Mutation causes race conditions
+  },
+});
+```
+::::
 
-Mutation in place—such as modifying `self.x` in `before_agent` or other hooks—can lead to subtle bugs and race conditions, because many operations run concurrently (subagents, parallel tools, and parallel invocations on different threads).
+Mutation in place, such as modifying `self.x` in `before_agent`, mutating a shared variable in `beforeAgent`, or changing other shared values in hooks, can lead to subtle bugs and race conditions because many operations run concurrently (subagents, parallel tools, and parallel invocations on different threads).
```
npentrel: probably want to gate this - please check whether I got this right
```suggestion
:::python
Mutation in place, such as modifying `self.x` in `before_agent` or changing other shared values in hooks, can lead to subtle bugs and race conditions because many operations run concurrently (subagents, parallel tools, and parallel invocations on different threads).
:::
:::js
Mutation in place, such as modifying `state.x` in `beforeAgent`, mutating a shared variable in `beforeAgent`, or changing other shared values in hooks, can lead to subtle bugs and race conditions because many operations run concurrently (subagents, parallel tools, and parallel invocations on different threads).
:::
```

christian-bromann: Good catch 👌

PR #3438, opened 2026-04-04
docs: Add SupraWall Callback Handler Integration Guide

wiserautomation:
This PR adds documentation for the **SupraWall** integration, a deterministic security and compliance layer for LangChain agents. SupraWall provides a `SupraWallCallbackHandler` that allows developers to:
- Enforce human-in-the-loop approvals for destructive tools.
- - Maintain tamper-evident audit logs (essential for EU AI Act compliance).
- - Block prompt injections at the tool-execution level.


npentrel (approved):

PR #3534, opened 2026-04-10
docs(deepagents): add filesystem permissions page

nick-hollon-lc:
Add documentation for the `FilesystemPermission` API, covering declarative path-based access control for agent filesystem tools. Includes rule structure, evaluation order, subagent inheritance, and composite backend scoping.


src/oss/deepagents/permissions.mdx
```diff
@@ -0,0 +1,280 @@
+---
+title: Permissions
+description: Control filesystem access with declarative permission rules for Deep Agents
+---
+
+Control which files and directories an agent can read or write using declarative permission rules. Pass a list of rules to `permissions=` and the agent's built-in filesystem tools respect them automatically, without writing custom [backend policy hooks](/oss/deepagents/backends#add-policy-hooks).
```
npentrel: I think we don't need to mention the previous workaround?
```suggestion
Control which files and directories an agent can read or write to using declarative permission rules. Pass a list of rules to `permissions=` and the agent's built-in filesystem tools respect them.
```

src/oss/deepagents/permissions.mdx
```diff
@@ -0,0 +1,280 @@
...
+
+## Limitations
+
+<Warning>
+Permissions only apply to the built-in filesystem tools (`ls`, `read_file`, `glob`, `grep`, `write_file`, `edit_file`). Custom tools and MCP tools that access the filesystem are not covered.
+</Warning>
+
+<Warning>
+Sandbox backends support arbitrary command execution, which can bypass path-based restrictions. Using permissions with a sandbox backend raises `NotImplementedError` unless all permission paths are scoped to [composite backend](/oss/deepagents/backends) routes.
+</Warning>
```
npentrel: - Generally we don't do multiple warning callouts in a row. I think in this case we need these warnings also further above. At the bottom of the page is too easy to miss.

src/oss/deepagents/permissions.mdx
npentrel: Add the warning text here as normal text to reiterate it, I think that's enough. 
```suggestion
Control which files and directories an agent can read or write using declarative permission rules. Pass a list of rules to `permissions=` and the agent's built-in filesystem tools respect them.

Permissions only apply to the built-in filesystem tools (`ls`, `read_file`, `glob`, `grep`, `write_file`, `edit_file`). Custom tools and MCP tools that access the filesystem are not covered.
```

src/oss/deepagents/permissions.mdx
```diff
@@ -0,0 +1,280 @@
...
+)
+```
+
+:::
+
+## Limitations
+
+<Warning>
+Permissions only apply to the built-in filesystem tools (`ls`, `read_file`, `glob`, `grep`, `write_file`, `edit_file`). Custom tools and MCP tools that access the filesystem are not covered.
+</Warning>
```
npentrel:
```suggestion
```

src/oss/deepagents/permissions.mdx
npentrel: Is this a limitation or is this desired? Either way I think there's no danger so this doesn't need to be a warning does it? I think if you want to point out Permissions do not apply to sandbox backends you can put this in the top or with the sandbox examples or both

```suggestion
```

eyurtsev (approved):

PR #3580, opened 2026-04-14
docs: fix interrupt handling example in SQL agent tutorial

### LINKED ISSUE #3208: doc bug: a code problem in official tutorial
### Type of issue

issue / bug

### Language

Python

### Description

https://docs.langchain.com/oss/python/langchain/sql-agent: In this tutorial, in the "6. Implement human-in-the-loop review" part, the third code example has a problem. The original code is given below.

from langgraph.types import Command 

for step in agent.stream(
    Command(resume={"decisions": [{"type": "approve"}]}),
    config,
    stream_mode="values",
):
    if "messages" in step:
        step["messages"][-1].pretty_print()
    elif "__interrupt__" in step:
        print("INTERRUPTED:")
        interrupt = step["__interrupt__"][0]
        for request in interrupt.value["action_requests"]:
            print(request["description"])
    else:
        pass

The code hopes to proceed with the "approve" command after the program is interrupted. It works if there's only one interruption. But when there are multiple interruptions, like when the llm makes one query after another(which is unpredictable), this piece of code doesn't work when it is put in a loop.
The solution is simple. Replace elif with if, or write if "__interrupt__" in step first. Otherwise,  when the step is ["massages", "__interrupt__"], the propram can't recognize the existence of "__interrupt__" and ends awkwardly.

monish-rgb:
## Summary
Fixes the interrupt-handling example in the SQL agent tutorial.

## Changes
- Replaced `elif "__interrupt__" in step:` with `if "__interrupt__" in step:`
- Preserved the rest of the example logic

## Why
When a step contains both `"messages"` and `"__interrupt__"`, the previous `elif` branch can skip interrupt handling. This update matches the issue report and makes the example work correctly for multiple interruptions.

Fixes #3208

### DIFF
--- a/src/oss/langchain/sql-agent.mdx
+++ b/src/oss/langchain/sql-agent.mdx
@@ -521,7 +521,7 @@ for step in agent.stream(
 ):
     if "messages" in step:
         step["messages"][-1].pretty_print()
-    elif "__interrupt__" in step:
+    if "__interrupt__" in step:
         print("INTERRUPTED:")
         interrupt = step["__interrupt__"][0]
         for request in interrupt.value["action_requests"]:


monish-rgb:
Hi @npentrel, pull request is ready for review. It fixes issue #3208 by updating the SQL agent tutorial example from 'elif' to 'if'

npentrel (approved):
Thank you!

PR #3659, opened 2026-04-21
Remove deprecated docker install + upgrade instructions from the docs

katmayb:
Fixes DOC-1016


src/langsmith/kubernetes.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Self-host LangSmith on Kubernetes
 sidebarTitle: Install on Kubernetes
+icon: "download"
```
fjmorris: I looked for the specific `kubernetes` wheel icon, but https://tabler.io/icons does not have one.

With Kubernetes, you really don't "download" to a cluster, you more "upload" to a cloud Kubernetes cluster. So maybe the `cloud-upload` icon might be more appropriate, like this:

Image

fjmorris (approved):
lgtm pending 1 suggestion

src/langsmith/kubernetes.mdx
katmayb: I guess, but I think we want to be careful having anything "cloud-y" in the self-hosted docs. I'll take a look at the icons

katmayb: I'm going to remove the icons from both of these pages for now!

PR #3663, opened 2026-04-21
clarify `subgraphs=True` needed when wrapping `create_agent`

mdrxy:
Document a subtle streaming gotcha: when `create_agent` (a compiled graph) is added as a node in a parent `StateGraph`, it becomes a subgraph, and `stream_mode="messages"` won't emit inner LLM tokens without `subgraphs=True`. This commonly surfaces when code that streams tokens fine via `agent.astream(...)` suddenly goes silent after wrapping.

### DIFF
--- a/src/oss/langchain/streaming.mdx
+++ b/src/oss/langchain/streaming.mdx
@@ -256,6 +256,10 @@ for await (const [token, metadata] of await agent.stream(
 ```
 :::
 
+<Warning>
+    **Wrapping an agent as a node in a parent `StateGraph`?** @[`create_agent`] returns a compiled graph, so using it as a node makes it a subgraph. `stream_mode="messages"` on the parent graph will not emit tokens from subgraphs unless you pass `subgraphs=True` (`subgraphs: true` in JS). See [Subgraph outputs](/oss/langgraph/streaming#subgraph-outputs) and [Streaming from sub-agents](#streaming-from-sub-agents).
+</Warning>
+
 ## Custom updates
 
 :::python
--- a/src/oss/langchain/streaming.mdx
+++ b/src/oss/langchain/streaming.mdx
@@ -256,9 +256,17 @@ for await (const [token, metadata] of await agent.stream(
 ```
 :::
 
-<Warning>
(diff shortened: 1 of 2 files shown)


src/oss/langchain/streaming.mdx
```diff
@@ -337,6 +337,10 @@ for await (const [token, metadata] of await agent.stream(
 ```
 :::
 
+<Warning>
+    **Wrapping an agent as a node in a parent `StateGraph`?** @[`create_agent`] returns a compiled graph, so using it as a node makes it a subgraph. `stream_mode="messages"` on the parent graph will not emit tokens from subgraphs unless you pass `subgraphs=True` (`subgraphs: true` in JS). See [Subgraph outputs](/oss/langgraph/streaming#subgraph-outputs) and [Streaming from sub-agents](#streaming-from-sub-agents).
```
hntrl: use fences instead of JS call out

src/oss/langchain/streaming.mdx
```diff
@@ -337,6 +337,10 @@ for await (const [token, metadata] of await agent.stream(
 ```
 :::
 
+<Warning>
```
hntrl: Is this a warning? This just seems like a call out instead of unsafe behavior

hntrl (approved):

src/oss/langgraph/streaming.mdx
```diff
@@ -1280,6 +1280,65 @@ for await (const chunk of await graph.stream(
 ```
 :::
 
+:::python
+<Note>
+    This applies to every `stream_mode`, including `"messages"`. Agent builders like @[`create_agent`] return a **compiled graph**, so adding one as a node turns it into a subgraph. Without `subgraphs=True`, `stream_mode="messages"` on the parent graph will not emit token chunks from the inner agent's LLM calls. Invoking `agent.astream(...)` directly will, which is why this often shows up only after wrapping.
```
npentrel:
```suggestion
<Note>
    This applies to every `stream_mode`, including `"messages"`. Agent builders like @[`create_agent`] return a **compiled graph**, so adding one as a node turns it into a subgraph. Without `subgraphs=True`, `stream_mode="messages"` on the parent graph will not emit token chunks from the inner agent's LLM calls. Invoking `agent.stream(...)` directly will, which is why this often shows up only after wrapping.
```

npentrel (approved):

PR #3717, opened 2026-04-24
chore: ⿁ scary loops ⿁

hinthornw:


src/langsmith/env-var.mdx
```diff
@@ -8,6 +8,12 @@ The Agent Server supports specific environment variables for configuring a deplo
 
 Set `BG_JOB_ISOLATED_LOOPS` to `True` to execute background runs in an isolated event loop separate from the serving API event loop.
 
+<Warning>
+Enabling this flag does not remove the underlying problem. It moves synchronous blocking work off the serving API's event loop so health checks stop failing, but the blocking code continues to run on the background loop and may continue to cause issues in production, like degraded throughput, tail-latency spikes, starved workers, or connection pool exhaustion (see the pool-size caveat below), and poor scaling under load.
```
cwlbraa:
```suggestion
Enabling this flag does not remove the underlying problem. It moves synchronous blocking work off the serving API's event loop so health checks stop failing, but the blocking code continues to run on the background loop and will continue to cause issues in production, like degraded throughput, tail-latency spikes, starved workers, or connection pool exhaustion (see the pool-size caveat below), and poor scaling under load.
```

cwlbraa (approved):

phvash (approved):

src/langsmith/env-var.mdx
hinthornw: Oops pushed the same but put in bold ha

hinthornw: ty!

PR #3818, opened 2026-04-30
feat(oss): hitl config expansion

sydney-runkle:
updating based on https://github.com/langchain-ai/langchain/pull/37095

### DIFF
--- a/src/oss/langchain/frontend/human-in-the-loop.mdx
+++ b/src/oss/langchain/frontend/human-in-the-loop.mdx
@@ -194,7 +194,7 @@ interface ActionRequest {
 }
 
 interface ReviewConfig {
-  allowedDecisions: ("approve" | "reject" | "edit")[];
+  allowedDecisions: ("approve" | "reject" | "edit" | "respond")[];
 }
 ```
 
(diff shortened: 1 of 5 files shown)


src/oss/langchain/frontend/human-in-the-loop.mdx
```diff
@@ -379,6 +424,7 @@ After the user makes a decision, the full cycle looks like this:
 3. The agent receives the `HITLResponse` and continues execution
 4. If approved, the tool runs with the original (or edited) arguments
 5. If rejected, the agent receives the reason and decides its next step
+6. If responded, the human's message is returned directly as the tool result without executing the tool
```
lnhsingh: This is misnumbered. For auto incrementing numbered steps, use all `1.`s

lnhsingh: Actually, looking at this, they shouldn't be numbered bc they're not sequential. The `If approved`/ `If rejected`, `If responded` should all prob be sub bullets under the `3. The agent receives the HITLResponse...`

lnhsingh (approved):
Conditionally approved once the numbered list issue is resolved

PR #3848, opened 2026-05-03
Document Context Hub (quickstart, concepts, manage contexts programmatically)

fjmorris:
based on work by emanbareket in #3662, #3666 and #3667

Fixes DOC-1063


vishnu-ssuresh (approved):
LGTM! Will update images as designs change.

PR #3865, opened 2026-05-04
Add general LS int tracing docs for openai compatible providers

katmayb:
Fixes DOC-662


katmayb:
@bwhiting2356 you opened an issue for this a few weeks ago, can you review and check that it covers what you were thinking? Thanks!

lnhsingh (approved):

PR #3935, opened 2026-05-09
fix(docs): prioritize interrupts in SQL agent streaming examples

mahirdesai2004:
## Summary

Fixes the SQL agent tutorial streaming example where the interrupt handling branch becomes unreachable when using stream_mode="values".

Since streamed steps may contain both:
- messages
- __interrupt__

the previous condition order caused the interrupt branch to never execute.

This PR updates the example logic to prioritize interrupt handling before message printing in both affected Python examples.

## Changes
- Reordered conditional checks in sql-agent.mdx
- Removed unreachable else: pass
- No runtime or library behavior changes (docs-only fix)

## Related Issues/PRs
Fixes #37251

### DIFF
--- a/src/oss/langgraph/sql-agent.mdx
+++ b/src/oss/langgraph/sql-agent.mdx
@@ -871,15 +871,13 @@ for step in agent.stream(
     config,
     stream_mode="values",
 ):
-    if "messages" in step:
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thanks!

PR #3940, opened 2026-05-10
docs(middleware): LangGraph workflow composition section

mdrxy:
Added a new section to the middleware overview explaining that `create_agent` returns a compiled LangGraph, which means an agent configured with middleware can be dropped into a larger `StateGraph` as a node or subgraph. Includes Python and TypeScript examples showing how to compose an agent with HITL and other middleware alongside classification and routing steps.

## Changes

- **Document agent reuse in larger LangGraph topologies** — new section on `src/oss/langchain/middleware/overview.mdx` clarifies that middleware hooks (HITL, retries, PII redaction, etc.) continue to run when the agent is used as a subgraph node, and points readers to the subgraph composition guide.

### DIFF
--- a/src/oss/langchain/middleware/overview.mdx
+++ b/src/oss/langchain/middleware/overview.mdx
@@ -66,6 +66,57 @@ Middleware exposes hooks before and after each of those steps:
     className="rounded-lg mx-auto"
 />
 
+## Use middleware inside a LangGraph workflow
+
+Middleware is not a separate runtime—it compiles down into the same [LangGraph](/oss/langgraph/overview) nodes and edges that power @[`create_agent`]. Because `create_agent` returns a compiled graph, you can drop the whole agent (middleware and all) into a larger `StateGraph` as a node or subgraph, and every middleware hook continues to run.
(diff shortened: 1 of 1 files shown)


src/oss/langchain/middleware/overview.mdx
```diff
@@ -66,6 +66,62 @@ Middleware exposes hooks before and after each of those steps:
...
+
+# Assumes read_email, send_email, classify_node, and route are defined elsewhere.
+email_agent = create_agent(
+    model="claude-sonnet-4-6",
+    tools=[read_email, send_email],
+    middleware=[HumanInTheLoopMiddleware(interrupt_on={"send_email": True})],
+)
+
+graph = (
+    StateGraph(MessagesState)
```
sydney-runkle: we don't recommend using this naymore

mdrxy: what should we switch to?

sydney-runkle:
@open-swe please switch to AgentState instead of MessagesState

PR #3958, opened 2026-05-12
docs: add Managed Deep Agents private preview docs

victorm-lc:
## Overview
Adds private preview docs for Managed Deep Agents:

- Adds a LangSmith deployment guide for deploying with `deepagents deploy` and invoking through `/v1/deepagents/threads`.
- Adds a manually authored Managed Deep Agents API overview under the LangSmith Deployment reference section.
- Replaces the OSS Deep Agents deploy page with a shorter Managed Deep Agents page and updates related launch wording.
- Clarifies that customers can still deploy Deep Agents with a standard LangSmith Deployment when they need custom code, custom routes, advanced auth, full Agent Server APIs, stronger isolation controls, or maximum scalability.

The goal is to document the private preview path without implying generated OpenAPI coverage or the old bundled frontend/custom auth behavior for Managed Deep Agents.

## Type of change

**Type:** New documentation page / Update existing documentation

## Related issues/PRs
- GitHub issue:
- Feature PR: #24821
- Linear issue:
- Slack thread:

## Checklist
- [x] I have read the [contributing guidelines](README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)
- [x] I have tested my changes locally using `docs dev`
- [x] All code examples have been tested and work correctly
- [x] I have used **root relative** paths for internal links
- [x] I have updated navigation in `src/docs.json` if needed

## Tests

- [x] `make lint_prose FILES="src/langsmith/deploy-managed-deep-agent.mdx src/langsmith/managed-deep-agents-api.mdx src/oss/deepagents/deploy.mdx src/oss/deepagents/going-to-production.mdx src/oss/deepagents/comparison.mdx src/oss/deepagents/cli/overview.mdx src/langsmith/deployment.mdx src/langsmith/deployment-quickstart-da.mdx"`
- [x] `git diff --check HEAD~2..HEAD`
- [ ] `make broken-links` fails on existing broken links outside the changed files. The reported failures are in existing pages such as `langsmith/assistants.mdx`, `langsmith/fleet/code.mdx`, `oss/*/concepts/products.mdx`, `oss/*/deepagents/streaming.mdx`, `oss/*/langchain/streaming.mdx`, and `snippets/oss/requires-langgraph-server.mdx`; the new Managed Deep Agents pages were not listed.

## Review notes

Please review the private preview positioning, especially:

- The route shape: `/v1/deepagents/agents/*` and `/v1/deepagents/threads/*`.
- The `X-Api-Key` auth header wording.
[...]

PR #4002, opened 2026-05-13
Update Managed Deep Agents private preview docs

fjmorris:
- Incorporated feedback from Parker from PR 3958.

Fixes DOC-1116


fjmorris:
@pjrule Please do an initial review for changes to address your comments from https://github.com/langchain-ai/docs/pull/3958.

pjrule (commented):
Markdown changes LGTM. I think we should sync with @victorm-lc / @rx5ad about the `/auth-scopes/{scope_slug}/credentials` endpoint; it's likely that its omission from the `/v1/deepagents` endpoints was simply an oversight on my part (if this holds, adding an alias to expose it should be super easy, and we can avoid removing it from the docs here).

(update: decided to remove)

pjrule (approved):

PR #4011, opened 2026-05-14
docs(event-streaming): fix raw event payload, text-delta type, and async examples

nick-hollon-lc:
## Summary

- **deepagents**: fix raw protocol event iteration in the \"Consume concurrently\" section — `event[\"params\"][\"data\"]` is a `(event_dict, metadata)` tuple so the payload is at `data[0]`, and the delta type is `\"text-delta\"` not `\"text\"`
- **all three pages** (deepagents, langchain, langgraph): reframe the Python multiple-projections section to lead with the async path (`astream_events` + `asyncio.gather`) and relegate `stream.interleave()` to a sync-only fallback
- **async examples verified**: `astream_events()` returns a coroutine and requires `await`; `message.text` in async projections is an `AsyncProjection` and also requires `await` — both confirmed by running the examples against the live API

## Test plan

- [x] All three async examples (`deepagents`, `langchain`, `langgraph`) run successfully against live API
- [x] Raw event structure confirmed: `data` is a 2-tuple, `data[0]` is the event dict, delta type is `\"text-delta\"`
- [x] `make lint_prose` passes on all changed files

## AI involvement

Changes were implemented and verified with Claude Code.

### DIFF
--- a/src/oss/deepagents/event-streaming.mdx
+++ b/src/oss/deepagents/event-streaming.mdx
@@ -264,7 +264,26 @@ for await (const subagent of stream.subagents) {
 Coordinator and subagent output often interleave. Consume projections concurrently when you need live UI updates.
 
 :::python
-Use `stream.interleave(...)` when you want one sync loop over multiple projections:
+For concurrent consumption in async code, use `astream_events` with `asyncio.gather`:
+
+```py
+import asyncio
+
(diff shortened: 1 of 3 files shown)


nick-hollon-lc:
Closing in favor of a new PR — branch renamed to remove the + character that was failing the create-preview CI check.

PR #4012, opened 2026-05-14
docs(event-streaming): fix raw event payload, text-delta type, and async examples

nick-hollon-lc:
## Summary

- **deepagents**: fix raw protocol event iteration in the "Consume concurrently" section — `event["params"]["data"]` is a `(event_dict, metadata)` tuple so the payload is at `data[0]`, and the delta type is `"text-delta"` not `"text"`
- **all three pages** (deepagents, langchain, langgraph): reframe the Python multiple-projections section to lead with the async path (`astream_events` + `asyncio.gather`) and relegate `stream.interleave()` to a sync-only fallback
- **async examples verified**: `astream_events()` returns a coroutine and requires `await`; `message.text` in async projections is an `AsyncProjection` and also requires `await` — both confirmed by running the examples against the live API

## Test plan

- [x] All three async examples (deepagents, langchain, langgraph) run successfully against live API
- [x] Raw event structure confirmed: `data` is a 2-tuple, `data[0]` is the event dict, delta type is `"text-delta"`
- [x] `make lint_prose` passes on all changed files

### DIFF
--- a/src/oss/deepagents/event-streaming.mdx
+++ b/src/oss/deepagents/event-streaming.mdx
@@ -357,7 +357,8 @@ for await (const event of stream) {
 
   const block = data.delta ?? {};
   if (block.type === "text-delta") {
-    const source = event.params.namespace.length > 0 ? "subagent" : "coordinator";
+    const isSubagent = event.params.namespace.some((seg) => seg.startsWith("tools:"));
+    const source = isSubagent ? "subagent" : "coordinator";
     console.log(`[${source}] ${block.text}`);
   }
 }
--- a/src/oss/deepagents/event-streaming.mdx
+++ b/src/oss/deepagents/event-streaming.mdx
@@ -345,6 +345,25 @@ for event in stream:
 ```
 :::
 
+:::js
+```ts
+const stream = await agent.streamEvents(input, { version: "v3" });
+
+for await (const event of stream) {
+  if (event.method !== "messages") continue;
+
+  const data = event.params.data;
+  if (data.event !== "content-block-delta") continue;
+
+  const block = data.delta ?? {};
+  if (block.type === "text-delta") {
+    const source = event.params.namespace.length > 0 ? "subagent" : "coordinator";
+    console.log(`[${source}] ${block.text}`);
+  }
+}
+```
+:::
+
 ## Subagents versus subgraphs
 
 `stream.subgraphs` shows graph execution structure. `stream.subagents` shows product-level Deep Agents task delegations. Use `stream.subagents` for user-facing UI because it hides internal graph nodes and exposes the subagent concept directly.
--- a/src/oss/deepagents/event-streaming.mdx
+++ b/src/oss/deepagents/event-streaming.mdx
@@ -332,7 +332,7 @@ for event in stream:
     if event.get("method") != "messages":
         continue
 
-    payload = event["params"]["data"][0]  # 2-tuple; [0] is a protocol event dict or a message object
+    payload = event["params"]["data"][0]
     if not isinstance(payload, dict):
         continue
     if payload.get("event") != "content-block-delta":
--- a/src/oss/deepagents/event-streaming.mdx
+++ b/src/oss/deepagents/event-streaming.mdx
@@ -264,7 +264,26 @@ for await (const subagent of stream.subagents) {
 Coordinator and subagent output often interleave. Consume projections concurrently when you need live UI updates.
 
 :::python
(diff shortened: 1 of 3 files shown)


src/oss/deepagents/event-streaming.mdx
```diff
@@ -264,7 +264,26 @@ for await (const subagent of stream.subagents) {
...
+
+async def consume_subagents():
+    async for subagent in stream.subagents:
+        async for message in subagent.messages:
+            print(f"[{subagent.name}]", await message.text)
+
+await asyncio.gather(consume_coordinator(), consume_subagents())
+```
+
+For synchronous code, use `stream.interleave(...)` instead:
```
ccurme, line 286: needs py / js fencing? (I see `interLeave` here for python)

src/oss/deepagents/event-streaming.mdx
```diff
@@ -313,12 +332,12 @@ for event in stream:
...
         continue
 
-    payload = event["params"]["data"]
+    payload = event["params"]["data"][0]  # data is a (event_dict, metadata) tuple
     if payload.get("event") != "content-block-delta":
         continue
 
     block = payload.get("delta") or {}
-    if block.get("type") == "text":
+    if block.get("type") == "text-delta":
```
ccurme, line 342: when I run this and a subagent is invoked, I get an error:
```python
from deepagents import create_deep_agent

agent = create_deep_agent(model="anthropic:claude-sonnet-4-6")

input_message = {
    "role": "user",
    "content": "Invoke a subagent using the `task` tool to calculate 2+2.",
}


stream = agent.stream_events({"messages": [input_message]}, version="v3")

for event in stream:
    if event.get("method") != "messages":
        continue

    payload = event["params"]["data"][0]  # data is a (event_dict, metadata) tuple
    if payload.get("event") != "content-block-delta":
        continue

    block = payload.get("delta") or {}
    if block.get("type") == "text-delta":
        source = "subagent" if event["params"]["namespace"] else "coordinator"
        print(f"[{source}] {block['text']}")
```
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 18
     15     continue
     17 payload = event["params"]["data"][0]  # data is a (event_dict, metadata) tuple
---> 18 if payload.get("event") != "content-block-delta":
     19     continue
     21 block = payload.get("delta") or {}

AttributeError: 'ToolMessage' object has no attribute 'get'
```

src/oss/deepagents/event-streaming.mdx
```diff
@@ -313,12 +332,12 @@ for event in stream:
...
         source = "subagent" if event["params"]["namespace"] else "coordinator"
         print(f"[{source}] {block['text']}", end="", flush=True)
```
ccurme:
```suggestion
        print(f"[{source}] {block['text']}")
```
this repeats the `"[source]"` within a single line, e.g.,
```
[coordinator] Hello! How can I help you[coordinator]  today?
```

src/oss/deepagents/event-streaming.mdx
nick-hollon-lc: interleave is python only purely because we have a sync path. checked the preview and things look correct given this

src/oss/deepagents/event-streaming.mdx
```diff
@@ -264,7 +264,26 @@ for await (const subagent of stream.subagents) {
 Coordinator and subagent output often interleave. Consume projections concurrently when you need live UI updates.
 
 :::python
```
npentrel, line 266: :::python
```suggestion
## Consume concurrently

Coordinator and subagent output often interleave. Consume projections concurrently when you need live UI updates.

```

src/oss/deepagents/event-streaming.mdx
npentrel: please start the fence before the headline then, otherwise there headline will be there with essentially empty content (see above comment)

src/oss/deepagents/event-streaming.mdx
nick-hollon-lc: checked the mintlify preview and there was one example lower down that didn't have a JS example with it so fixed that but now this looks fine

src/oss/deepagents/event-streaming.mdx
nick-hollon-lc: concurrent consumptions isn't python only just interleave. i checked the preview and this looks correct to me

npentrel (approved):

PR #4077, opened 2026-05-19
docs(managed-deep-agents): update MDA deploy guide to include mcp servers

fjmorris:
Update the Managed Deep Agents deploy guide and API reference for the
private preview shape.

- Rewrite deploy-managed-deep-agent.mdx end-to-end: create / get / update / delete agent, thread lifecycle, and run streaming, all as Python (httpx) and JavaScript (fetch) examples + cURL. (Python SDK / JS SDK / useStream tabs dropped — SDK support coming in a follow-up.)
- Document MCP server registration on the agent, including how credentials flow through and per-user vs. shared auth.
- Switch all example tools from local stubs to the hosted Tavily MCP server so the examples are runnable end-to-end.
- Expand managed-deep-agents-openapi.json with the new MCP server endpoints and tighten the existing schemas to match.
- Note the Context Hub-backed agent repo (instructions / skills / subagents / tools.json + the runtime /memories/ workspace) so readers understand what the create/update payload actually stores.

Fixes DOC-1139


src/langsmith/deploy-managed-deep-agent.mdx
```diff
@@ -62,115 +75,491 @@ Managed Deep Agents keeps the familiar Deep Agents project shape. Keep these fil
...
 }
 ```
 
+Each `mcp_server_url` you reference here needs to be registered first via [Configure MCP servers](#configure-mcp-servers) so its credentials are available at invocation time.
+
 The `interrupt_config` map lets you require human approval for selected tools. Each key is `{mcp_server_url}::{tool_name}` (separator is two colons); trailing slashes on the URL are stripped before matching. Additional `::`-separated parts after the tool name (such as the MCP server display name shown in the example above) are accepted but ignored when matching. Set the value to `true` to require human approval before the tool runs, or `false` to allow it without interrupt.
 
+## Configure MCP servers
+
+Tools that need credentials — custom MCP servers, private endpoints, or any bearer-authenticated API — are registered once per workspace via `POST /v1/deepagents/mcp-servers`. The platform stores the server's URL and any credential headers; when an agent's `tools[].mcp_server_url` matches a registered server, the stored headers are attached automatically each time the tool is invoked.
```
lnhsingh:
```suggestion
Tools that need credentials (like custom MCP servers, private endpoints, or any bearer-authenticated API) are registered once per workspace via `POST /v1/deepagents/mcp-servers`. The platform stores the server's URL and any credential headers; when an agent's `tools[].mcp_server_url` matches a registered server, the stored headers are attached automatically each time the tool is invoked.
```

lnhsingh (approved):
nit

PR #4124, opened 2026-05-21
docs: add LangChain tab to LLM gateway quickstart

avirath-kumar:
## Summary

- Add a LangChain tab to the gateway quickstart's "Make a call" `<CodeGroup>`, showing `init_chat_model` + `create_agent` against the Anthropic gateway path. This is the path most
LangChain users will reach for first.
- Fix a `/v1` mismatch in `OPENAI_BASE_URL`: step 1 exported `$BASE_URL/openai`, but the OpenAI Python SDK requires the `/v1` suffix on `base_url`. Anyone copying the env vars and
calling `OpenAI()` without an explicit `base_url` got 404s. Env var now exports `$BASE_URL/openai/v1`.
- Make the OpenAI SDK and Anthropic SDK tabs read their `base_url` from the env vars set in step 1, instead of hardcoding the URL. Removes the inconsistency between the env var setup
and the code samples.
- Add a `<Note>` in section 3 calling out that when an app emits its own LangSmith traces (via LangChain / LangGraph), the gateway trace and the application trace are not linked to
the same parent run.

## Test plan

- [ ] Run `make dev` (Node ≤22) and verify the quickstart page renders with four tabs (curl, OpenAI SDK, Anthropic SDK, LangChain) in step 2.
- [ ] Verify the `<Note>` callout renders correctly in section 3.
- [ ] Manually run the env var exports from step 1, then run each of the OpenAI / Anthropic / LangChain snippets end-to-end against the gateway and confirm a 200 + trace.
- [ ] `make lint_prose FILES="src/langsmith/llm-gateway-quickstart.mdx"` — clean.
- [ ] `make broken-links` — clean.

### DIFF
--- a/src/langsmith/llm-gateway-quickstart.mdx
+++ b/src/langsmith/llm-gateway-quickstart.mdx
@@ -23,7 +23,7 @@ Set the following in your terminal (or add them to your `~/.zshrc` to persist ac
 export BASE_URL="https://gateway.smith.langchain.com"
 
 export ANTHROPIC_BASE_URL="$BASE_URL/anthropic"
-export OPENAI_BASE_URL="$BASE_URL/openai"
+export OPENAI_BASE_URL="$BASE_URL/openai/v1"
 export GOOGLE_GEMINI_BASE_URL="$BASE_URL/gemini"
 
 export LANGSMITH_API_KEY="lsv2_..._....cbed3e"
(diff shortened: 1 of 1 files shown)


katmayb (approved):
lgtm from a docs perspective

ramonpetgrave64 (approved):

PR #4187, opened 2026-05-28
docs(self-hosted): add Terraform guides for AWS, GCP, and Azure

devfreddy-langchain:
## Summary

Adds Terraform deployment guides for LangSmith self-hosted across AWS, GCP, and Azure, mirroring the structure of the existing Helm setup guide. The set is one shared overview (`self-host-terraform`) plus five pages per provider:

- **Deploy**: end-to-end walkthrough from prerequisites to a running install.
- **Architecture**: platform layers, managed services, networking, ingress, secret flow, and the Terraform module graph.
- **Variables**: exhaustive reference for the `infra` (foundation) layer inputs.
- **Quick reference**: condensed commands and add-on toggles.
- **Troubleshooting**: failure symptoms, causes, and fixes.

OpenShift is not included. The OCP module is still Preview upstream, so it is left out until it matures.

## Why

Self-hosted customers using the public Terraform modules at github.com/langchain-ai/terraform have been pointed at provider READMEs. This PR brings those walkthroughs into the official docs so the Terraform path matches the experience the Helm path already has.

## Accuracy audit

Every page was audited line by line against the actual modules in github.com/langchain-ai/terraform (cloned as read-only reference). Substantive corrections include:

- **Variables pages rewritten to be exhaustive** for each provider's `infra` layer, sourced from the real `variables.tf` defaults and validation blocks rather than a curated subset.
- **Provider secret models documented as they actually differ**, not homogenized:
  - AWS syncs secrets from SSM Parameter Store into the cluster through the External Secrets Operator.
  - GCP writes the Kubernetes Secrets directly from Terraform; Secret Manager is an optional durable copy. No External Secrets Operator is installed. An earlier draft incorrectly described an ESO sync path on GCP (inherited from the upstream module docs, which carry the same error); this is now corrected.
  - Azure writes `secrets.auto.tfvars` and stores secrets in Key Vault.
- **Azure managed services corrected**: Azure Managed Redis (`Microsoft.Cache/redisEnterprise`, TLS port 10000) replaces the retired classic Azure Cache for Redis (port 6380); `amr_sku` replaces `redis_capacity`; PostgreSQL default version is 14; the `ltree` extension was removed from the enabled list (the module enables `pgcrypto`, `btree_gin`, `pg_trgm`, `btree_gist`, and `citext`).
- **AWS sizing corrected**: removed a fabricated table claiming `sizing_profile` drives instance types. Infrastructure capacity comes from the `infra` variables directly; `sizing_profile` only selects the Helm sizing overlay. EKS version updated to 1.33.
- **GCP secret names corrected** to the `-credentials` suffixed names the chart expects, and blob storage reframed around the shipped native GCS mode (Workload Identity, no HMAC keys) with the S3-compatible mode noted as optional.
- **Terminology aligned**: "stages" and "layers" throughout, matching the consolidated guides.

## Areas needing careful review

- The three variables pages have the largest surface area. They are scoped to the `infra` foundation layer, which is what these guides deploy.
- The provider secret-flow sections, since each provider genuinely differs and the differences are easy to over-normalize.
- The GCP architecture secret-flow diagram and cluster-infrastructure table, which the ESO correction touched.

## Test plan

- [x] `make lint_prose` passes (0 errors across the 11 audited terraform pages).
- [ ] `make broken-links` passes in CI. It cannot run locally here because Mintlify does not support Node 25+. The one new internal link (`/langsmith/self-host-scale`) was verified to exist and to have a `docs.json` nav entry.
- [ ] Mintlify preview renders all pages and navigation entries resolve.

## AI involvement

Drafted and audited in collaboration with Claude Code (Opus). The author directed content, scope, and accuracy decisions; Claude verified claims against the Terraform modules, rewrote the variables references, applied the corrections above, and ran Vale lint cleanup.


DavoCoder (commented):
The main aspect we need to improve here before publishing this is the diagrams. Something more clean and customer friendly

katmayb (approved):

PR #4215, opened 2026-05-29
chore(deepagent): add streaming as core capability

christian-bromann:
Allowing to stream individual subagents efficiently seems to me like a great core capability to point out.


src/oss/deepagents/overview.mdx
```diff
@@ -256,6 +256,9 @@ Use the **Deep Agents SDK** to build agents that handle complex, multi-step task
 <Card title="Subagent spawning" icon="users-group">
     A built-in `task` tool spawns general-purpose or specialized [subagents](/oss/deepagents/subagents) for context isolation on subtasks. For long-running or parallel work, [async subagents](/oss/deepagents/async-subagents) run in the background with progress checks, follow-ups, and cancellation.
 </Card>
+<Card title="Streaming" icon="broadcast">
```
npentrel, line 259: Is this high level enough to be a core capability? I want to avoid us adding new features just because they're new, hence the question. Not opposed to merging, just want to check yours and Sydney's thoughts on this. cc: @sydney-runkle

npentrel (approved):

PR #4251, opened 2026-06-02
docs: use camelCase middleware factory names in agents.mdx JS examples

emmayusufu:
The JavaScript middleware examples in the agents guide use the Python class names (`new TodoListMiddleware()`, `new ModelRetryMiddleware()`, etc.), but the `langchain` package exports these as camelCase factory functions, so the snippets fail to compile (`'langchain' has no exported member named 'TodoListMiddleware'`). This switches them to the factory functions, matching the package and the `middleware/built-in.mdx` examples.

`piiMiddleware` takes a PII type argument in JS (unlike the Python `PIIMiddleware()`), so that one becomes `piiMiddleware("email")`. The `deepagents` middleware (`FilesystemMiddleware`, `SubAgentMiddleware`) are genuinely classes and are left as-is. Verified each fixed snippet compiles against langchain 1.4.4.

### DIFF
--- a/src/oss/langchain/agents.mdx
+++ b/src/oss/langchain/agents.mdx
@@ -406,7 +406,7 @@ agent = create_agent(
 :::
 :::js
 ```typescript
-import { createAgent, TodoListMiddleware } from "langchain";
+import { createAgent, todoListMiddleware } from "langchain";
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thank you for fixing that!

PR #4289, opened 2026-06-03
docs: document sub-agent identity (lc_agent_name) in event streaming

nick-hollon-lc:
## Why

Recent work landed first-class sub-agent identity in the event-streaming stack: a sub-agent's configured `name` is bound as `lc_agent_name` and surfaces in streamed events. This is shipped across LangGraph (Python), LangChain (Python), and Deep Agents — but the three event-streaming pages either didn't explain where the name comes from, or (LangChain Python) still used the older `stream.subgraphs` + manual filter workaround. This PR brings the docs in line with what ships.

## Changes

- **LangChain** (`oss/langchain/event-streaming.mdx`): switch the Python sub-agent example to the now-shipped dedicated `stream.subagents` projection (`SubagentTransformer`, wired into `create_agent`), instead of iterating `stream.subgraphs` and filtering on `graph_name`. The handle exposes `.name` and `.cause`. Corrected the stale "there's no separate sub-agent-only projection" line.
- **Deep Agents** (`oss/deepagents/event-streaming.mdx`): explain that a handle's `name` is the sub-agent's configured name, selected via the `task` tool's `subagent_type` and bound to the delegated run.
- **LangGraph** (`oss/langgraph/event-streaming.mdx`): clarify that `subgraph.graph_name` is the compiled graph/agent name, that tool-dispatched named agents surface on `stream.subgraphs`, and that the opening `lifecycle` event carries a `cause` linking back to the dispatching tool call.

## Language parity (please review)

Verified against shipped source on local branches:
- Python `create_agent` surfaces `stream.subagents` natively (`langchain/agents/_subagent_transformer.py`, registered in `factory.py`).
- **In-process TypeScript does not yet expose `stream.subagents`** (only the deployment SDK has subagent discovery). So the JS example intentionally stays on `stream.subgraphs`, with a parity note. If the JS in-process projection has landed since, the JS example should be updated to match the Python one.

## Verification

- `make lint_prose` on the three files: 0 errors.
- `make broken-links`: ✅ No broken links.
- API surface (method names, `version="v3"`, projection accessors, `.name`/`.cause`) checked against shipped SDK source and tests. Snippets were **not** executed against a live model (needs API keys), so they are not runtime-tested end to end.


src/oss/langgraph/event-streaming.mdx
```diff
@@ -173,6 +173,8 @@ for await (const subgraph of stream.subgraphs) {
 ```
 :::
 
+`subgraph.graph_name` is the `name` of the compiled graph or agent. A named agent dispatched from a tool (for example, a `create_agent(name=...)` invoked through the Deep Agents `task` tool) surfaces here under that name, and the `lifecycle` event that opens the scope carries a `cause` linking back to the dispatching tool call (see [Lifecycle](#lifecycle)).
```
npentrel:
```suggestion
`subgraph.graph_name` is the `name` of the compiled graph or agent. A named agent dispatched from a tool (for example, a `create_agent(name=...)` invoked through the Deep Agents `task` tool) surfaces here under that name, and the `lifecycle` event that opens the scope carries a `cause` linking back to the dispatching tool call. See [Lifecycle](#lifecycle) for more information.
```

npentrel (approved):
LGTM % nit

PR #4301, opened 2026-06-04
docs(callbacks): add Spanlens callback handler

sunes26:
## Summary

Adds a new community callback handler page for [Spanlens](https://spanlens.io) — an open-source (MIT) LLM observability platform — to `src/oss/python/integrations/callbacks/`.

`SpanlensCallbackHandler` is published as part of the `spanlens` PyPI package (under the `langchain` extra) and uses LangChain's standard callback contract, so it works without modification for plain LangChain chains, LCEL pipelines, agents, and LangGraph compiled graphs.

## What's on the page

- **Installation** — `pip install \"spanlens[langchain]\"`
- **LangChain example** — prompt | llm pipeline traced into a span tree
- **LangGraph example** — multi-node compiled graph (draft → polish) traced as one trace with one chain-span per node
- **Attaching to an existing trace** — pass a `TraceHandle` so the handler records spans under a wider trace the caller owns
- **Options table** — every constructor argument (`client`, `trace`, `trace_name`, `capture_*` flags, `max_*_bytes`)
- **Related links** — Spanlens docs, PyPI, GitHub, self-host guide

## Why a separate file vs. extending an existing one

Placed alongside the existing `agentsystems_notary.mdx` and `google_bigquery.mdx` so it picks up the docs site's nav automatically, matches the surrounding tone, and stays maintainable as Spanlens evolves.

## Verification

- Code shown in the page mirrors [`spanlens.integrations.langchain.SpanlensCallbackHandler`](https://github.com/spanlens/Spanlens/blob/main/packages/sdk-python/spanlens/integrations/langchain.py) — option names, defaults, and behaviour all match the implementation
- [`spanlens` on PyPI](https://pypi.org/project/spanlens/) ships the `langchain` extra (falls back to a plain class when `langchain-core` isn't installed)
- Spanlens dashboard's [/docs/sdk#langchain](https://spanlens.io/docs/sdk#langchain) page covers the same flow for cross-reference

Happy to revise tone, trim examples, or move sections to match how the doc set evolves. Thank you for reviewing!


sunes26:
Ping @mdrxy — this is a small docs-only addition (single new `callbacks/spanlens.mdx` page, no code, no changeset) and CI is green. Ready for review when you have a moment. Happy to tighten wording or trim sections if anything reads off.

npentrel (approved):

PR #4328, opened 2026-06-05
reorder skills page

npentrel:
- Start with skill definition - and adding a metaphor that I think helps
- moved some of the "what skills are" into usage and renamed what skills are to how they work. I think that works better.
- Reframed skill permissions into sharing skills (I think permissions is how to implement it, but the task people are trying to accomplish is sharing skills)
- added a section on when to use skills, kept it short
- added section on skills with other resources. It makes it very explicit that you need to reference files and how to do so
- Added skill diagram

---
Updates:
- Expanded Skill permissions
- added more layering info to how skills work


sydney-runkle (approved):
wow, amazing!

PR #4357, opened 2026-06-08
docs: move engine webhooks page next to engine page in nav

bentanny:
Engine webhooks are fired by the Engine feature, not by automation rules. Moving the page out of the Automations group and placing it directly after the Engine page in the Observability tab so readers find it in the right context.


baskaryan:
whats Chat?

Screenshot 2026-06-08 at 6 42 08 PM

ayoung19 (approved):

PR #4369, opened 2026-06-09
Add ls_message_view exclude constant to the docs

katmayb:


src/langsmith/view-traces.mdx
```diff
@@ -86,7 +86,7 @@ You can control how runs appear in the Messages view using metadata keys on indi
   - `"langchain"`: parse as LangChain message format
   - `"anthropic"`: parse as Anthropic message format
   - `"responses"`: parse as OpenAI Responses API format
-- `ls_message_view_exclude`: Exclude an individual run from the Messages view. For code examples, refer to [Exclude runs from the Messages view](/langsmith/messages-view-integrations#exclude-runs-from-the-messages-view).
+- `LS_MESSAGE_VIEW_EXCLUDE`: Exclude an individual run from the Messages view. Import the constant from `langsmith` (Python and JS), or use the literal string `"ls_message_view_exclude"`. Child runs that execute inside the tagged run's tracing context inherit the exclusion. For code examples, refer to [Exclude runs from the Messages view](/langsmith/messages-view-integrations#exclude-runs-from-the-messages-view).
```
ybathula707: Heads up — this part is right for @traceable/traceable() (we have tests for that exact cascade in the SDK PR). But the old bullet that said "child runs aren't implicitly excluded" got removed, and the doc still has examples for wrap_openai, wrapOpenAI, wrapAISDK, RunTree.createChild, and RunnableConfig — and we never actually tested whether the cascade reaches those. Could we add a second bullet right after, something like:

- For the other surfaces shown here (\wrap_openai`/`wrapOpenAI`, `wrapAISDK`, `RunTree.createChild`, LangChain `RunnableConfig`), just set the key on each run you want hidden — don't count on it cascading.`
- 
That way the strong cascade promise stays scoped to the thing we actually verified.

src/langsmith/ls-metadata-parameters.mdx
```diff
@@ -347,6 +348,37 @@ For more details, see [Customize the Messages view](/langsmith/view-traces#custo
...
+**Example:**
+```python
+from langsmith import LS_MESSAGE_VIEW_EXCLUDE, traceable
+
+@traceable(run_type="llm", metadata={LS_MESSAGE_VIEW_EXCLUDE: True})
+def classify_intent(query: str) -> str:
+    return llm.predict(f"Classify: {query}")
+```
+
+For additional code examples across Python and JS contexts (`@traceable`, `trace`, `wrap_openai`, `RunnableConfig`, `wrapAISDK`, `RunTree.createChild`), see [Exclude runs from the Messages view](/langsmith/messages-view-integrations#exclude-runs-from-the-messages-view).
```
ybathula707, line 376: Nit-  the "What it does" paragraph earlier in this section correctly limits the cascade claim to @traceable/traceable(). But this cross-reference then lists six surfaces together, and a reader who skimmed the cascade sentence above could easily assume it applies to all of them. Could we either reorder so @traceable and trace come first and frame the list as "here are examples of tagging the run," or just drop the cascade sentence from the "What it does" paragraph and leave the cascade explanation only in the messages-view-integrations Notes section? Either way it keeps the two parts from being read together.

ybathula707 (changes requested):

ybathula707 (approved):

katmayb:
TFTR!

PR #4415, opened 2026-06-11
Update Deploy dropdown

katmayb:
Follow-on from new docs IA work


src/langsmith/deployment.mdx
```diff
@@ -1,24 +1,12 @@
 ---
 title: LangSmith Deployment
 sidebarTitle: Overview
-mode: "custom"
```
lnhsingh, line 4: set it so the right toc doesn't show up. looks squished rn

PR #4416, opened 2026-06-11
feat(docs): add docs to cover task primitive in code interpreter and subagent orchestrations

colifran:
## Overview
Reworks the interpreters docs around the two ways interpreter code reaches outside its sandbox: calling tools, and dispatching subagents.

- Adds a Subagent orchestration section for the built-in task() global, covering how to steer orchestration through subagent config and prompting, working with large inputs, and six workflow patterns (classify and act, fan-out and synthesize, adversarial verification, generate and filter, tournament, loop until done) with diagrams and collapsible Python/JS examples.
- Adds a short What interpreter code can reach intro framing the two bridges.
- Folds the recursive-language-models material into the new section, since it's the same task() mechanism over a large variable space.
- Removes interpreter skills from interpreters.mdx and skills.mdx, plus the related skills_backend wiring in the skills-usage snippets and the security table.

## Type of change

## Checklist

### DIFF
--- a/src/oss/deepagents/interpreters.mdx
+++ b/src/oss/deepagents/interpreters.mdx
@@ -20,9 +20,7 @@ Where [sandboxes](/oss/deepagents/sandboxes) are a code-first way for acting on 
 
 ## When to use an interpreter
 
-Most agent work alternates between model reasoning and tool execution. That works for simple actions, but it becomes awkward when the agent needs to compose many steps, reason over structured data, or manage intermediate state.
-
-An interpreter gives the agent a runtime for that work. Instead of asking the model to choose every next step one tool call at a time, the agent can write a small program that runs control flow, calls allowlisted tools, stores variables, and returns a compact result to the model.
+Most agent work alternates between model reasoning and single tool calls. That breaks down when the agent needs to compose many steps, reason over structured data, or hold intermediate state. An interpreter gives it a runtime to do that work in code: run control flow, call allowlisted tools, store variables, and return only a compact result to the model.
 
 Use interpreters when the agent needs to:
 
@@ -51,9 +49,7 @@ graph LR
     class Result output
 ```
 
-This works by running code against [**QuickJS**](https://github.com/quickjs-ng/quickjs), a lightweight JavaScript runtime designed for embedded execution. The runtime gives the agent a place to evaluate code without exposing host filesystem, network, shell, package, or clock APIs by default.
-
-QuickJS is the execution boundary for interpreter code. Explicit bridges, such as programmatic tool calling, decide what capabilities the code can access.
+Code runs against [**QuickJS**](https://github.com/quickjs-ng/quickjs), a lightweight JavaScript runtime with no access to the host filesystem, network, shell, or clock by default. Explicit bridges decide what the code can reach.
 
 ## Choose the right execution path
 
@@ -359,7 +355,7 @@ releaseSummary;
 
 ### Orchestration patterns
 
-The agent picks a strategy from the shape of the task. These patterns emerge from how it writes interpreter code; you don't select them yourself, and the subagents you configure determine what's available. The diagrams below show the common shapes, each with a runnable example.
+The agent picks a strategy from the shape of the task; these emerge from how it writes interpreter code, not from configuration, and the subagents you make available determine what it can do. Every pattern shares one model: hold work in JS variables, dispatch subagents with `task()`, and combine results in code. The diagrams below show the common shapes, each with a runnable example.
 
 #### Classify and act
 
@@ -439,7 +435,7 @@ const result = await agent.invoke({
 
 #### Fan-out and synthesize
 
-The agent dispatches the same kind of work across many items in parallel, then combines the results. This is the most common pattern.
+The agent dispatches the same kind of work across many items in parallel, then combines the results.
 
 ```mermaid
 graph LR
@@ -720,7 +716,7 @@ agent = create_deep_agent(
 )
 
 result = await agent.ainvoke({
-    "messages": [{"role": "user", "content": "Find all the dead code in this repo. Be thorough — I want every unused export and unreachable function."}]
+    "messages": [{"role": "user", "content": "Find all the dead code in this repo. Be thorough. I want every unused export and unreachable function."}]
 })
 ```
 :::
@@ -738,7 +734,7 @@ const agent = createDeepAgent({
 });
 
 const result = await agent.invoke({
-  messages: [{ role: "user", content: "Find all the dead code in this repo. Be thorough — I want every unused export and unreachable function." }],
+  messages: [{ role: "user", content: "Find all the dead code in this repo. Be thorough. I want every unused export and unreachable function." }],
 });
 ```
 :::
--- a/src/oss/deepagents/interpreters.mdx
+++ b/src/oss/deepagents/interpreters.mdx
@@ -61,8 +61,8 @@ QuickJS is the execution boundary for interpreter code. Explicit bridges, such a
 | --- | --- |
 | One or two simple external calls | Normal tool calling |
 | A small program that loops, branches, retries, or aggregates results | Interpreter |
-| Many selected tool calls that should run from code | Interpreter with programmatic tool calling |
-| Reusable helpers used across threads | Interpreter with [interpreter skills](/oss/deepagents/skills#interpreter-skills) |
+| Many selected tool calls that should run from code | Interpreter with [programmatic tool calling](#programmatic-tool-calling) |
+| Many independent units of work, or multiple perspectives on one task | Interpreter with [subagent orchestration](#orchestrate-subagents-from-code) |
 | Shell commands, package installs, tests, or full OS filesystem access | [Sandboxes](/oss/deepagents/sandboxes) |
 
 ## Add an interpreter to an agent
@@ -119,7 +119,7 @@ const agent = createDeepAgent({
 
 ## Run code in the interpreter
 
-The middleware adds an `eval` tool to the agent. The tool runs TypeScript in a persistent context, captures `console.log`, and returns the result of the last expression.
+The middleware adds an `eval` tool to the agent. The tool runs JavaScript in a persistent context, captures `console.log`, and returns the result of the last expression.
 
 The agent can write code like this:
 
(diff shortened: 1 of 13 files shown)


src/oss/deepagents/interpreters.mdx
```diff
@@ -201,51 +168,71 @@ import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
...
 );
 
-reports.join("\n\n");
+results.join("\n\n");
 ```
 
-Because this is code, the agent can also handle failures locally:
+<Warning>
+    PTC calls currently execute through the interpreter bridge and do not go through the normal tool calling path. As a result, `interrupt_on` approval workflows are not enforced per PTC-invoked tool call.
+</Warning>
```
hntrl, line 200: this is `interruptOn` in JS, maybe we want to put it behind a fence

src/oss/deepagents/interpreters.mdx
```diff
@@ -306,11 +291,396 @@ const releaseSummary = riskReports
...
+    ],
+    middleware=[CodeInterpreterMiddleware()],
+)
+
+result = await agent.ainvoke({
+    "messages": [{"role": "user", "content": "Go through these 30 support tickets. Categorize each one, then for bugs give me reproduction steps, and for feature requests give me a feasibility assessment."}]
+})
+```
+:::
+
```
hntrl, line 304: should add some small code snippets here

colifran: ya let's update with some lean code snippets of what the agent writes - we can do something like "What you configure" and "What the agent writes"

src/oss/deepagents/interpreters.mdx
mdrxy, line 17: should we bump to 0.2.0?

```suggestion
    Interpreters require `langchain-quickjs>=0.2.0`.
```

we don't support below 3.11

src/oss/deepagents/interpreters.mdx
```diff
@@ -201,51 +168,71 @@ import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
 
 const agent = createDeepAgent({
   model: "openai:gpt-5.4",
```
mdrxy: nit

```suggestion
  model: "openai:gpt-5.5",
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -20,9 +20,7 @@ Where [sandboxes](/oss/deepagents/sandboxes) are a code-first way for acting on
 
 ## When to use an interpreter
 
-Most agent work alternates between model reasoning and tool execution. That works for simple actions, but it becomes awkward when the agent needs to compose many steps, reason over structured data, or manage intermediate state.
-
-An interpreter gives the agent a runtime for that work. Instead of asking the model to choose every next step one tool call at a time, the agent can write a small program that runs control flow, calls allowlisted tools, stores variables, and returns a compact result to the model.
+Most agent work alternates between model reasoning and single tool calls. That breaks down when the agent needs to compose many steps, reason over structured data, or hold intermediate state. An interpreter gives it a runtime to do that work in code: run control flow, call allowlisted tools, store variables, and return only a compact result to the model.
```
mdrxy: my naive rebuttal from a user POV: _what about parallel tool calls?_

should we call out by name why they're not much better?

src/oss/deepagents/interpreters.mdx
mdrxy, line 1: would we consider adding a callout to dcode to this page linking to instructions on how to use / test it out there?

npentrel: I think that's a good idea, I think the docs for how to use interpreters in dcode should live near dcode but we can add a sentence after the version callouts to say "If you'rel ooking to use interpreters with `dcode`, see... "

npentrel: if it works in 0.1.0 then no. Does it?

src/oss/deepagents/interpreters.mdx
npentrel: so should we keep tool execution as a better way of putting that?

colifran: I think we should call out that tool calling doesn't work at scale and isn't reliable. For example, we've seen instances where when trying to analyze 500 traces the agent will choose to only analyze 75. It doesn't reliably do all 500.

src/oss/deepagents/interpreters.mdx
colifran: updated.

src/oss/deepagents/interpreters.mdx
colifran: good call

src/oss/deepagents/interpreters.mdx
colifran: updated to call this out.

src/oss/deepagents/interpreters.mdx
```diff
@@ -143,41 +120,31 @@ totals;
...
+By default, interpreter code runs against QuickJS with no access to the host filesystem, network, shell, package manager, or clock. It can compute, hold state, and write to `console.log`, and nothing more.
+
+Two explicit bridges extend that reach:
+
+- **Tools**, through [programmatic tool calling](#programmatic-tool-calling). Expose an allowlist of tools as async functions under the `tools` namespace. These can be the agent's own tools or standalone tools you define and pass in.
+- **Subagents**, through the built-in [`task()` global](#orchestrate-subagents-from-code). Dispatch configured subagents from code and orchestrate them in plain JavaScript.
 
-This is useful when intermediate tool results are only inputs to the next step. The interpreter can process, filter, or aggregate those results before anything returns to the model context, which can make multi-tool/multi-step workflows more token efficient.
+Programmatic tool calling is off until you pass an allowlist. Subagent dispatch is on by default whenever the agent has subagents, and you can turn it off. Nothing else crosses the QuickJS boundary unless you expose it.
```
npentrel:
```suggestion
Programmatic tool calling is off until you [enable it](#enable-ptc). Subagent dispatch is on by default whenever the agent has subagents, and you can turn it off. Nothing else crosses the QuickJS boundary unless you expose it.
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -143,41 +120,31 @@ totals;
...
 
-### How it works
+Programmatic tool calling (PTC) exposes selected agent tools inside the interpreter under the global `tools` namespace. Instead of asking the model to issue one tool call, wait for the result, and then decide the next call, the agent can write code that calls tools in loops, branches, retries, or parallel batches.
 
-1. You choose which tools the interpreter can call with the `ptc` allowlist.
-2. The middleware exposes those tools as async JavaScript functions under `tools`.
-3. The agent writes interpreter code that calls those functions with `await`.
-4. The interpreter runs the tool bridge, receives the tool result, and continues executing code.
-5. The model receives the final interpreter output, not every intermediate value.
+This helps when intermediate results are only inputs to the next step: the interpreter filters or aggregates them before anything returns to the model, keeping multi-step workflows token efficient. It is model-agnostic, implemented by middleware rather than a provider-specific tool-calling API.
```
npentrel:
```suggestion
This helps when intermediate results are only inputs to the next step: the interpreter filters or aggregates them before anything returns to the model, keeping multi-step workflows token-efficient. It is model-agnostic, implemented by middleware rather than a provider-specific tool-calling API.
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -201,51 +168,78 @@ import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
...
+</Warning>
+:::
+
+## Orchestrate subagents from code
+
+When an agent has [subagents](/oss/deepagents/subagents) and interpreter middleware, the interpreter exposes a built-in `task()` global that dispatches subagents from code. A task spanning many independent units (reviewing every file in a directory, triaging a batch of tickets) becomes a loop that fans the work out, so it runs deterministically instead of one model-chosen tool call at a time.
+
+### The `task()` global
+
+`task()` takes the prompt the subagent receives (`description`), which configured subagent to run (`subagentType`), and an optional `responseSchema` for structured output. It runs a full agentic loop and resolves to the subagent's result:
```
npentrel:
```suggestion
`task()` takes the following inputs:

 - `description`: the prompt for the subagent
 - `subagentType`: which configured subagent to run
 - `responseSchema` (optional): structured output
 
a `task()` runs a full agentic loop and resolves to the subagent's result:
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -201,51 +168,78 @@ import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
...
 
 ```javascript
```
npentrel, line 222: I think this needs js gates

src/oss/deepagents/interpreters.mdx
```diff
@@ -201,51 +168,78 @@ import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
...
 
-Recursive language models use an interpreter as a workspace for decomposition. The model keeps a large input or working set in runtime variables, writes code to inspect and split it, calls subagents or other model tools on smaller pieces, and then stitches the returned results together in code.
+### Working with large inputs
 
-This separates the variable space from the agent's context. The variable space is information stored in the interpreter, and the agent's context is what the model actually processes in the next model call. The model can decide which snippets become subagent tasks, which results need another pass, and what final synthesis should return to the main conversation.
+The same `task()` mechanism handles inputs too large to reason over in one pass, the approach described in the [Recursive Language Models paper](https://arxiv.org/abs/2512.24601). Keep the large input or working set in interpreter variables (the *variable space*, distinct from the model's context), then let the model inspect it, split it into slices, dispatch a subagent per slice, and stitch the returned results back together.
```
npentrel: note sure why `same`
```suggestion
The `task()` mechanism handles inputs too large to reason over in one pass, the approach described in the [Recursive Language Models paper](https://arxiv.org/abs/2512.24601). Keep the large input or working set in interpreter variables (the *variable space*, distinct from the model's context), then let the model inspect it, split it into slices, dispatch a subagent per slice, and stitch the returned results back together.
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -281,9 +275,7 @@ flowchart TB
     class Answer output
 ```
 
-For background on this pattern, see the [Recursive Language Models paper](https://arxiv.org/abs/2512.24601).
-
-In Deep Agents, the recursive call is often the `task` tool exposed through programmatic tool calling. The interpreter can call subagents over many slices, combine their answers, and return a single synthesized result:
+The recursive call is just `task()` over slices. The interpreter calls a subagent for each one, combines their answers, and returns a single synthesized result:
```
npentrel:
```suggestion
The recursive call runs `task()` over slices of the input. The interpreter calls a subagent for each one, combines their answers, and returns a single synthesized result:
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -306,11 +298,553 @@ const releaseSummary = riskReports
...
 ```
 
-## Interpreter skills
+### Orchestration patterns
+
+The agent picks a strategy from the shape of the task; these emerge from how it writes interpreter code, not from configuration, and the subagents you make available determine what it can do. Every pattern shares one model: hold work in JS variables, dispatch subagents with `task()`, and combine results in code. The diagrams below show the common shapes, each with a runnable example.
+
+#### Classify and act
+
+Items are classified first, then each is handled by a specialized subagent based on its classification. This lets you process mixed inputs where different items need different expertise.
```
npentrel:
```suggestion
Items are classified first, then each item is handled by a specialized subagent based on its classification. This lets you process mixed inputs where different items need different expertise.
```

src/oss/deepagents/interpreters.mdx
```diff
@@ -306,11 +298,553 @@ const releaseSummary = riskReports
 releaseSummary;
 ```
 
-## Interpreter skills
+### Orchestration patterns
```
npentrel: this is great though I wonder if it should be its own page - but merge this as is and we can consider that separately.

npentrel (approved):
A range of minor feedback, please accept what makes sense. Since all of this is nit/minor level I'll approve already and you can merge when you feel it's ready

PR #4428, opened 2026-06-12
docs(deepagents): fix LangSmithSandbox import path and constructor usage

moneshrallapalli:
## Overview

Fixes two errors in the Python snippet on the Sandbox page (`src/oss/deepagents/frontend/sandbox.mdx`) that make it fail immediately when copy-pasted:

1. **Wrong import path:** the snippet imports `from deepagents.sandbox import LangSmithSandbox`, but the `deepagents.sandbox` module does not exist, so the snippet raises `ModuleNotFoundError`. The correct path, already used on the Sandboxes page (`sandboxes.mdx`), is `from deepagents.backends.langsmith import LangSmithSandbox`.
2. **Nonexistent constructor parameter:** the snippet calls `LangSmithSandbox(resolve=lambda: ...)`, but the actual constructor is `LangSmithSandbox(sandbox: Sandbox)` — there is no `resolve` parameter, so after fixing the import the snippet raises `TypeError`. Replaced with the supported `BackendFactory` pattern: `create_deep_agent`'s `backend` parameter is typed `BackendProtocol | BackendFactory | None`, so a `lambda runtime: ...` that resolves the sandbox per thread is passed directly. This also matches the TypeScript example in the same file, which already uses the factory pattern correctly.

Both errors were verified against `deepagents==0.6.8` using `inspect.signature` on `LangSmithSandbox.__init__` and `create_deep_agent`.

## Type of change

## Related issues/PRs

- GitHub issue: N/A
- Feature PR: N/A

## Checklist

- [x] I have updated navigation in `src/docs.json` if needed (no nav changes required)

## Additional notes

`make lint_prose` passes on the changed file.


moneshrallapalli:
@npentrel @lnhsingh This is ready for review. Fixes a broken import path and a nonexistent constructor parameter in the LangSmithSandbox Python snippet, verified against deepagents 0.6.8.

npentrel (approved):
Thank you!

PR #4454, opened 2026-06-15
docs(streaming): fix incorrect type check in sub-agents example

### LINKED ISSUE #4447: [langchain]: Incorrect use of AIMessage instead of AIMessageChunk in streaming code example
### Type of issue

issue / bug

### Language

Python

## Description
In the `Streaming from sub‑agents` section of the streaming documentation, the code example that adds logic to track which agent is emitting tokens contains an incorrect type check.  

Specifically, the streaming loop uses `isinstance(token, AIMessage)` to process streamed chunks, but when `stream_mode=["messages"]` is used, each token is an **`AIMessageChunk`**, not an `AIMessage`. This check will always be `False`, preventing the streaming output from being rendered incrementally.

## Current (incorrect) code in the docs
```python
if isinstance(token, AIMessage):
    _render_message_chunk(token)
```

## Expected (correct) code
```python
if isinstance(token, AIMessageChunk):
    _render_message_chunk(token)
```

## Steps to reproduce the issue
1. Go to the documentation page: [Streaming from sub‑agents](https://docs.langchain.com/oss/python/langchain/streaming#streaming-from-sub-agents)
2. Look for the code block following the sentence:  
   *“Next, we add logic to the streaming loop to report which agent is emitting tokens:”*
3. Observe the `isinstance(token, AIMessage)` check, which should be `AIMessageChunk`.

## Why this matters
- The incorrect check breaks the streaming example: users who copy it will see no incremental token output.
- The same page already correctly uses `AIMessageChunk` in other examples, so this is an inconsistency that should be fixed for clarity.

## Suggested fix
Change the line `if isinstance(token, AIMessage):` to `if isinstance(token, AIMessageChunk):` in the affected code block.

maycode0:
## Description
This PR fixes an incorrect type check in the streaming documentation example under the "Streaming from sub‑agents" section.

**Problem:**  
The code uses `isinstance(token, AIMessage)`, but when `stream_mode=["messages"]` is used, each token is an `AIMessageChunk`, not an `AIMessage`. This prevents the streaming output from rendering.

**Solution:**  
Changed the check to `isinstance(token, AIMessageChunk)`.

## Related Issue
Closes #4447


npentrel (approved):
Thank you!

PR #4457, opened 2026-06-15
docs: add TokenMix provider integration page

TokenMixAi:
Adds a provider integration page for **TokenMix**, an OpenAI-compatible API gateway that exposes DeepSeek, Qwen, Kimi, GLM, MiniMax and other models through a single endpoint (`https://api.tokenmix.ai/v1`) and one API key.

Because the gateway is OpenAI-compatible, the page documents usage through the existing `langchain-openai` package by pointing `ChatOpenAI` at the TokenMix base URL — no new package is required. It mirrors the structure of the existing Together AI and xAI provider pages (Installation / Environment / Example) and adds short tool-calling and model-catalog notes.

New file:
- `src/oss/python/integrations/providers/tokenmix.mdx`

The provider folder is auto-globbed by Mintlify, so no `docs.json` change is needed (consistent with the existing `together.mdx` / `openrouter.mdx` pages, which also have no explicit nav entry).

Local checks:

```
$ codespell src/oss/python/integrations/providers/tokenmix.mdx
(no issues)
```

All links are external (tokenmix.ai) except the standard models concept reference. No `@[...]` API symbols are used, since TokenMix is consumed via `langchain-openai`.


npentrel (approved):

PR #4480, opened 2026-06-15
docs: add Salesforce OAuth provider setup for self-hosted Fleet

jfglanc:
## What

Adds a Salesforce entry to the self-hosted Fleet "Available providers" docs: a table row, the `salesforceOAuthProvider` config key, and a full setup-guide accordion covering both the Salesforce-side External Client App creation and the LangSmith-side provider configuration.

## Why

Self-hosted customers had no documented way to enable the Salesforce integration (Pylon #24496 / AB-2687). The Helm wiring landed in langchain-ai/helm#769 (`SALESFORCE_OAUTH_PROVIDER`); this is the customer-facing companion. Unlike the other providers, Salesforce requires the customer to create their own app (they are not using a LangChain-managed connected app), and Salesforce now blocks new Connected App creation, so the guide uses an External Client App throughout.

## How

- Table row + `salesforceOAuthProvider` YAML key, mirroring the existing providers.
- New accordion structured like the Microsoft entry: create the External Client App (Settings then Policies), copy the consumer key/secret, then configure the OAuth provider in LangSmith and apply via `langsmith_config.yaml`.
- The Salesforce-side steps (My Domain authorization URL, `api` + `refresh_token` scopes, the Settings-vs-Policies split, up-to-30-minute propagation, callback exact-match) are verified against official Salesforce docs.

## Testing

Verified MDX tag balance (Accordion / Steps / Step / Note / Warning all matched), the table anchor resolves to the new accordion id (`#salesforce-oauth-provider`), and the internal `langsmith_config.yaml` link reuses the same known-good target as every other accordion. No automated docs build exists in this repo.

## Anything Else

Opened as a **draft**: three LangSmith-side values are marked inline with `{/* TODO confirm */}` and must be resolved before merge:

- **Provider Type** value for Salesforce in the Fleet provider registry (drafted as `salesforce`)
- **Token endpoint auth method** (drafted as `client_secret_post`)
- Whether Fleet auto-requests the OAuth scopes vs relying on the app's selected scopes (and whether it sends PKCE)

A couple of release-dependent Salesforce UI labels (the org-level "allow External Client App creation" toggle, and "Enable OAuth" vs "Enable OAuth Settings") are worth a final check against a live org before publish.

Linear: AB-2687


src/langsmith/deploy-self-hosted-full-platform.mdx
```diff
@@ -520,6 +522,99 @@ To enable LinkedIn OAuth for Fleet, create a LinkedIn OAuth app and configure it
...
+  </Step>
+
+  <Step title="Enable OAuth and configure the OAuth settings">
+    Expand **API (Enable OAuth Settings)** and select **Enable OAuth**. Then configure:
+
+    - **Callback URL**, replacing `<hostname>` with your LangSmith hostname and `<provider-id>` with your provider ID:
+
+    ```
+    https://<hostname>/host-oauth-callback/<provider-id>
+    ```
```
bracesproul, line 550: i would make it something like this (but ideally all non-default ones can be nested inside a dropdown):
```suggestion
    Default:
    ```
    https://smith.langchain.com/host-oauth-callback/<provider-id>
    ```
    EU:
    ```
    https://eu.smith.langchain.com/host-oauth-callback/<provider-id>
    ```
    APAC:
    ```
    https://apac.smith.langchain.com/host-oauth-callback/<provider-id>
    ```
    AWS:
    ```
    https://aws.smith.langchain.com/host-oauth-callback/<provider-id>
    ```
    Self hosted:
    ```
    https://<your hostname>/host-oauth-callback/<provider-id>
    ```
```

src/langsmith/deploy-self-hosted-full-platform.mdx
```diff
@@ -520,6 +522,99 @@ To enable LinkedIn OAuth for Fleet, create a LinkedIn OAuth app and configure it
...
+
+    <Note>
+    After you create the app, allow up to 30 minutes for it to propagate before the first connection attempt.
+    </Note>
+  </Step>
+
+  <Step title="Configure OAuth provider in LangSmith">
+    In LangSmith, go to **Settings > OAuth Providers** and add a new provider:
+    - **Name**: For example, `Salesforce`
+    - **Provider ID**: Unique string, for example: `salesforce`
```
bracesproul: nit, i think this is the pattern we normally suggest
```suggestion
    - **Provider ID**: Unique string, for example: `salesforce-oauth-provider`
```

src/langsmith/deploy-self-hosted-full-platform.mdx
```diff
@@ -520,6 +522,99 @@ To enable LinkedIn OAuth for Fleet, create a LinkedIn OAuth app and configure it
...
+    - **Client ID**: Consumer Key from Salesforce
+    - **Client Secret**: Consumer Secret from Salesforce
+    - **Authorization URL**: `https://<MyDomain>.my.salesforce.com/services/oauth2/authorize`
+    - **Token URL**: `https://<MyDomain>.my.salesforce.com/services/oauth2/token`
+    - **Provider Type**: `salesforce` {/* TODO confirm exact value in Fleet provider registry */}
+    - **Token endpoint auth method**: `client_secret_post` {/* TODO confirm */}
```
bracesproul: dont think we expose this in the UI?

src/langsmith/deploy-self-hosted-full-platform.mdx
```diff
@@ -520,6 +522,99 @@ To enable LinkedIn OAuth for Fleet, create a LinkedIn OAuth app and configure it
...
+    </Note>
+  </Step>
+
+  <Step title="Apply the changes">
+    Add the following to your [`langsmith_config.yaml`](/langsmith/kubernetes#configure-your-helm-charts) and deploy:
+
+    ```yaml
+    fleet:
+      oauth:
+        providerOrgId: "<your-org-id>"
```
bracesproul:
```suggestion
        providerOrgId: "<your-langsmith-org-id>"
```

src/langsmith/deploy-self-hosted-full-platform.mdx
jfglanc: @bracesproul this is specifically for self-hosted documentation, not SaaS. All the other providers (Google, Linear, LinkedIn, Microsoft, Slack) use the same `<hostname>` form. Happy to add them here if useful, though

src/langsmith/deploy-self-hosted-full-platform.mdx
jfglanc: @bracesproul  I checked and Fleet already handles the provider type creation automatically, so the code is correct, no need to show in the UI at all. I updated the docs to reflect the simpler setup process.

src/langsmith/deploy-self-hosted-full-platform.mdx
bracesproul: ah lol youre right. good point!

bracesproul (approved):
sensational! ty

PR #4486, opened 2026-06-16
docs(otel): document linking OTel spans to a LangSmith SDK trace

QuentinBrosse:
## What

Documents how to link an OpenTelemetry span to an **existing LangSmith SDK trace** (nesting it under a run created via the LangChain/LangSmith SDK) using the `langsmith.*` span attributes.

- Adds the identity/linking attributes to the **Core LangSmith attributes** table — they were previously undocumented: `langsmith.trace.id`, `langsmith.span.id`, `langsmith.span.parent_id`, `langsmith.span.dotted_order` (with a link to the `dotted_order` reference).
- Adds a worked example under **Implementation examples**: "Link OpenTelemetry spans to a LangSmith SDK trace."

## Why

Native OTLP `parentSpanId` can't reference an existing LangSmith run (an OTLP span ID is 8 bytes; run IDs are full UUIDs). The supported path is to set the `langsmith.*` attributes with the parent's full UUIDs, which override the IDs derived from the native OTLP span. This was only known via support hand-offs; this documents it.

## Validation

Verified against **prod** LangSmith: created a LangChain-SDK run, emitted an OTel span with these attributes pointing at it, and confirmed the span materialized and nested under the SDK run in the same trace (parent had a normal full UUID — the case native `parentSpanId` cannot address).

## Related

- LIN-240 (OTLP spans with orphan parentSpanId)
- Companion docs PR: #4485 (documents the orphan-parent drop behavior)

### DIFF
--- a/src/langsmith/trace-with-opentelemetry.mdx
+++ b/src/langsmith/trace-with-opentelemetry.mdx
@@ -465,7 +465,7 @@ Here is an [example](https://smith.langchain.com/public/d6d47eeb-511e-4fda-ad17-
 
 ### Link OpenTelemetry spans to a LangSmith SDK trace
 
-Native OTLP `parentSpanId` can't reference an existing LangSmith run: an OTLP span ID is 8 bytes, while LangSmith run IDs are full UUIDs. To attach an OpenTelemetry span to a run created elsewhere (for example, a LangChain-SDK run), set the `langsmith.*` attributes with the parent's full UUIDs — they override the IDs derived from the native OTLP span, so the span nests under the existing run in the same trace.
+Native OTLP `parentSpanId` can't reference an existing LangSmith run: an OTLP span ID is 8 bytes, while LangSmith run IDs are full UUIDs. To attach an OpenTelemetry span to a run created elsewhere (for example, a LangChain-SDK run), set the `langsmith.*` attributes with the parent's full UUIDs. They override the IDs derived from the native OTLP span, so the span nests under the existing run in the same trace.
 
 ```python
 import uuid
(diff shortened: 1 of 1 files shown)


KiewanVillatel (approved):

src/langsmith/trace-with-opentelemetry.mdx
```diff
@@ -459,6 +463,46 @@ You do not need to set OTEL environment variables or exporters. `configure()` wi
 
 Here is an [example](https://smith.langchain.com/public/d6d47eeb-511e-4fda-ad17-2caa7bd7150b/r) of what the resulting trace looks like in LangSmith.
 
+### Link OpenTelemetry spans to a LangSmith SDK trace
+
+Native OTLP `parentSpanId` can't reference an existing LangSmith run: an OTLP span ID is 8 bytes, while LangSmith run IDs are full UUIDs. To attach an OpenTelemetry span to a run created elsewhere (for example, a LangChain-SDK run), set the `langsmith.*` attributes with the parent's full UUIDs — they override the IDs derived from the native OTLP span, so the span nests under the existing run in the same trace.
```
katmayb:
```suggestion
Native OTLP `parentSpanId` can't reference an existing LangSmith run: an OTLP span ID is 8 bytes, while LangSmith run IDs are full UUIDs. To attach an OpenTelemetry span to a run created elsewhere (for example, a LangChain-SDK run), set the `langsmith.*` attributes with the parent's full UUIDs. They override the IDs derived from the native OTLP span, so the span nests under the existing run in the same trace.
```

katmayb (approved):

PR #4604, opened 2026-06-24
feat(cursor): adding docs for new langsmith plugin [DOC-1295]

katmayb:


src/langsmith/trace-with-cursor.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+## Prerequisites
+
+Before setting up tracing, ensure you have:
+
+- [Cursor](https://cursor.com/) installed.
+- [Node.js](https://nodejs.org/) 22.13 or later. The hooks use the built-in `node:sqlite` module to recover attachments from Cursor's local database.
+- A [LangSmith API key](/langsmith/create-account-api-key).
+
+## Install and enable the plugin
```
katmayb, line 19: It looks like `pnpm install && pnpm build` in the README is not required for users just capturing cursor traces.

I think a fresh clone + `node scripts/install.mjs` would be enough for most users? (Rather than those contributing who need to build).

Let me know if I should add them back in

harisaiharish (approved):
fixed installation steps

PR #4609, opened 2026-06-24
fix: update sandbox private registry docs

langchain-infra:
## Description
Update the sandbox snapshot private registry examples to create a registry with SDK helpers, then use the returned registry ID for snapshot creation instead of unsupported inline snapshot credentials.

## Test Plan
- [x] `git diff --check`
- [ ] `make lint_prose FILES=src/langsmith/sandbox-snapshots.mdx` was not run because Vale is not installed in the sandbox.

Made by [Open SWE](https://openswe.vercel.app/agents/d22cdfb4-6120-dde5-1c66-307e27c4d982)

### DIFF
--- a/src/langsmith/sandbox-snapshots.mdx
+++ b/src/langsmith/sandbox-snapshots.mdx
@@ -48,20 +48,16 @@ console.log(snapshot.id);
 
 ### Private registries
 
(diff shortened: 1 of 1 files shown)


npentrel:
this is no longer correct / already in main - sorry for the delay

PR #4704, opened 2026-07-02
docs: add pdfmuse document loader integration

casperkwok:
Adds an integration docs page for `langchain-pdfmuse`, a document loader for [pdfmuse](https://github.com/casperkwok/pdfmuse) — a deterministic PDF/DOCX parser aimed at RAG.

- Runs locally — no API, and no ML dependencies in the core path
- Extracts text with exact coordinates + tables
- In `elements` mode, emits one `Document` per block with `page` / `heading_path` / `bbox` metadata
- Deterministic: the same file always yields identical `Document`s, so the index doesn't drift between runs

Package on PyPI: https://pypi.org/project/langchain-pdfmuse/

**Changes**
- new page: `src/oss/python/integrations/document_loaders/pdfmuse.mdx`
- listed in the PDFs table in `document_loaders/index.mdx`


npentrel (approved):

PR #4741, opened 2026-07-06
update sidebar organization deep agents

npentrel:
Screenshot 2026-07-06 at 14 58 24

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -196,6 +196,7 @@
                         "pages": [
                           "oss/python/deepagents/quickstart",
                           "oss/python/deepagents/customization",
+                          "oss/python/deepagents/models",
                           "oss/python/deepagents/comparison",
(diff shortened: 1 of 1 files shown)


mdrxy (approved):

src/docs.json
```diff
@@ -232,9 +227,33 @@
                               "oss/python/deepagents/dynamic-subagents"
                             ]
                           },
-                          "oss/python/deepagents/profiles",
                           "oss/python/deepagents/event-streaming",
-                          "oss/python/deepagents/streaming",
+                          "oss/python/deepagents/streaming"
```
ccurme, line 231: does streaming make sense in "execution environment"?

ccurme (approved):

src/docs.json
npentrel: that is where sydney put it - we can check this with her when she's back

PR #4751, opened 2026-07-06
chore: add experiment-runs migration guide [closes LIN-434]

GowriH-1:
## Description
Adds the SmithDB migration guide entry for the dataset experiment-runs query endpoint, covering method names, parameter and response changes, examples, and self-hosted feature flag guidance.

Addresses review feedback from Quentin:
- Clarified the intro wording and the "preview select fields" explanation, which was vague across all 5 language tabs
- Folded the experiment-ID lookup directly into each example (`client.read_project(project_name=...).id` / equivalent) instead of a separate section, matching the sibling `runs-query.mdx` guide's inline pattern
- Added missing Before/After example pairs for pagination and sort (previously cURL-only; other languages showed After only)
- Replaced the manual `next_cursor`-checking pagination example with each SDK's built-in auto-pagination helper (Python `async for`, TS `for await...of`, Kotlin `.autoPager()`); Go paginates via the request body's `Cursor` field directly after a real bug in its `QueryAutoPaging` helper turned up (see Testing setup)
- Simplified the Response fields section: instead of duplicating per-run field renames in every tab, it links to `runs-query.mdx`'s own section for the nested `runs` array (verified identical `Run` type across all four SDKs) and documents only what's unique here — the dataset-example fields alongside `runs`

Addresses review feedback from Kiewan:
- Split the conflated `limit`/`page_size` table row into two entries (semantically different — total cap vs. per-page count), matching `runs-query.mdx`'s precedent
- Removed two dead `*(not exposed)* -> *(removed)*` table rows (`format`, `include_annotator_detail`) that carried no migration information
- Rewrote the Sort section's framing so "renamed from `sort_params`" only applies where true (Go/Java); Python/TS state plainly they never supported it, backed by a real working After example instead of a fabricated Before

**All samples execute for real instead of being permanently dead code.** Every sample was wrapped in `if False:`/`if (false) { }`, so a passing test only proved the file parses, not that the API calls were correct — how an earlier TS `.create()`→`.query()` rename bug slipped past a green test in this same PR.

**Testing setup**: one dataset (`docs-experiment-runs-query-fixture`, 3 arithmetic examples, one deliberately wrong so grading is genuinely mixed) and **one experiment** (`docs-experiment-runs-query-fixture-experiment`), both created idempotently and shared across every topic (basic/pagination/sort) and language. An earlier version had each sample create its own fresh experiment — ~18 real creations sequentially in one CI job, overloading the test org badly enough to hang for 20+ minutes. Consolidating to one shared, name-checked experiment cut that to one real creation, ever; everything else just reads it.
- Python/TypeScript/Kotlin build the fixture with manual run + feedback calls rather than `evaluate()`, which only produces an auto-suffixed name — incompatible with a shared, checkable one. Go already built its fixture manually (no `evaluate()` in that SDK); only the name-check gating changed
- Found a real SDK bug: `langsmith-go`'s `QueryAutoPaging` sends the cursor as a URL query param, but `experiment-runs.query` is a POST expecting `cursor` in the JSON body — the backend silently ignores it and the helper loops on page 1 forever. Fixed by setting `Cursor` on the request body manually instead
- Fixture-creation code is hidden from rendered docs; the experiment-ID lookup stays visible since real users need that pattern
- `test-code-samples` is green in CI with the real `LANGSMITH_API_KEY` secret (2m58s)

Also aligns samples with current SDK naming: both the legacy and new v2 endpoints already existed in the pinned harness versions (`langsmith-go` v0.17.0, `langsmith-java` 0.1.0-beta.11) — those versions just predate a `create`→`query` method rename across all three SDKs within about two weeks of each other in early July. Python's `get_experiment_results()` is a hand-written convenience method, never renamed. Bumped the pinned Go/Java versions (`v0.18.2`/`beta.13`) and adopted `main`'s TS `langsmith@^0.8.1` bump so samples and harness both reflect current naming.

## SDK verification
Checked against the actual SDK source (cloned `langchain-ai/langsmith-sdk`, `langsmith-go`, `langsmith-java`, unshallowed for true history), not just re-read from the PR text: `Client.evaluate()`/`.has_dataset()`/`.has_project()`/`.create_run()`/`.create_feedback()` (Python `client.py`); `hasDataset`/`hasProject`/`createRun`/`createFeedback`/`listExamples` (TS `client.ts`); `Datasets`/`Examples`/`Sessions`/`Runs`/`Feedback` services and the `ItemsCursorPostPagination.GetNextPage()` cursor bug confirmed directly in the pinned `v0.18.2` module (Go — no `evaluate()` helper exists there); `SessionService`/`RunService`/`FeedbackService` builders (`RunIngest`, `FeedbackCreateSchema`, `SessionCreateParams`) (Kotlin/Java).

## Test Plan
- [x] `make test-code-samples` — green in CI with the real `LANGSMITH_API_KEY` secret; locally (no credentials) every sample fails only on `401 Unauthorized`, confirming it reaches a genuine network call
- [x] `make code-snippets` — regenerated extracted snippets; hidden fixture-creation code doesn't leak into rendered output
- [x] `make lint_prose` (Vale) — clean
- [x] `go build` / Kotlin `jbang` compile checks pass for every rewritten file
- [x] Merge conflicts with `main` resolved (adopted `main`'s independent `langsmith@^0.8.1` bump + `.npmrc` fix, merged `go.mod`'s `google/uuid` requirement)

Made by [Open SWE](https://openswe.vercel.app/agents/8e4eb612-0fb7-4d76-2598-4bafaa2a2208); review feedback, SDK verification, and real-execution rewrite by Claude Code.


src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,308 @@
...
+  </Tab>
+  <Tab title="After">
+    <SmithdbExperimentRunsQueryBasicAfterSh />
+  </Tab>
+</Tabs>
+
+  </Tab>
+</Tabs>
+
+#### Page with a cursor
```
QuentinBrosse: Should we add SDKs examples here?

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,308 @@
...
+<Tabs sync={false}>
+  <Tab title="Before">
+    <SmithdbExperimentRunsQueryPaginationBeforeSh />
+  </Tab>
+  <Tab title="After">
+    <SmithdbExperimentRunsQueryPaginationAfterSh />
+  </Tab>
+</Tabs>
+
+#### Sort by feedback score
```
QuentinBrosse, line 389: Should we add SDKs examples here?

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,308 @@
...
+    <SmithdbExperimentRunsQueryBasicBeforePy />
+  </Tab>
+  <Tab title="After">
+    <SmithdbExperimentRunsQueryBasicAfterPy />
+  </Tab>
+</Tabs>
+
+  </Tab>
+  <Tab title="TypeScript">
+    The legacy endpoint was not exposed on the public TypeScript Client. Use explicit `selects` in the v2 SDK method to keep the lightweight response shape.
```
QuentinBrosse: Should we just say that we added the method? I feel the `selects` thing doesn't make sense in this context?

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,308 @@
...
+<Tabs>
+  <Tab title="Python">
+    | Before | After |
+    |--------|-------|
+    | `client.get_experiment_results()` | `client.datasets.experiment_runs.query()` |
+  </Tab>
+  <Tab title="TypeScript">
+    | Before | After |
+    |--------|-------|
+    | *(no public v1 Client method)* | `client.datasets.experimentRuns.query()` |
```
QuentinBrosse: We use `v1` in some places, and `legacy` in others. Should we align on legacy?

We don't really have v1/v2 clients, all our SDK are v0. :(

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,466 @@
...
+<Tabs>
+  <Tab title="Python">
+    <Warning>
+    `experiment_ids` is required and replaces `session_ids`. Values are still experiment tracing-project UUIDs.
+    </Warning>
+
+    | Before (`get_experiment_results`) | After (`datasets.experiment_runs.query`) | Notes |
+    |---|---|---|
+    | Experiment project ID | `experiment_ids` | `get_experiment_results` accepted one project/experiment; v2 accepts a required non-empty list |
+    | `limit` | `page_size` | `page_size` defaults to 20, max 100 |
```
KiewanVillatel: If I'm not mistaken the two are semantically different. One is to define the max number of items returned (limit), the other one is to control pagination and number of items returned by page. As it's currently written, it seems like it's only a renaming, which is not accurate. Should we do the same as for [query runs](https://docs.langchain.com/langsmith/smithdb-sdk-migration#query-parameters) and extract two separate entries in the table to avoid confusion?

src/code-samples/langsmith/smithdb-migration/experiment-runs-query-basic.py
```diff
@@ -0,0 +1,36 @@
...
+    # :codegroup-tab: After
+    from langsmith import Client
+
+
+    async def main():
+        client = Client()
+        page = await client.datasets.experiment_runs.query(
+            dataset_id,
+            experiment_ids=[experiment_id],
+            page_size=20,
```
KiewanVillatel: In this example it looks like we are using `page_size` as a limit, which can be misleading and would not work in the general case (eg. if we want to retrieve 1k items which wouldn't be supported by page_size where max value is 100).

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,466 @@
...
+    | *(handled internally)* | `cursor` | Pass the previous page's `next_cursor` to fetch the next page |
+    | `preview` | `selects` | Omitted `selects` returns only run IDs; use `INPUTS_PREVIEW` and `OUTPUTS_PREVIEW` for previews, or `INPUTS` and `OUTPUTS` for full payloads |
+    | *(not exposed)* | `sort` | Use `{by, order}` for feedback-score sorting |
+    | `filters` | `filters` | Unchanged; maps experiment UUID strings to filter expressions |
+    | `comparative_experiment_id` | `comparative_experiment_id` | Unchanged |
+    | *(not exposed)* | `example_ids` | Optional example UUID filter, max 1000 |
+    | *(not exposed)* | *(removed)* | The v2 endpoint returns JSON only |
```
KiewanVillatel: Should we remove those 2 lines? Same for other languages.

src/snippets/langsmith/smithdb-migration/experiment-runs-query.mdx
```diff
@@ -0,0 +1,466 @@
...
+    <SmithdbExperimentRunsQueryPaginationAfterSh />
+  </Tab>
+</Tabs>
+
+  </Tab>
+</Tabs>
+
+#### Sort by feedback score
+
+The legacy endpoint used `sort_params.sort_by` and `sort_params.sort_order`. The new endpoint uses `sort.by` and `sort.order`, and sorting is supported only when you query one experiment.
```
KiewanVillatel: I find this example a bit confusing. We are mentioning that `sort_params.sort_by` and `sort_params.sort_order` were replaced by `sort.by` and `sort.order`, but then we don't provide example and say that `get_experiment_results` did not support sorting by feedback score.

If the example is about showing how to filter by feedback score, I'd remove the phrase about sort by renaming. If it's about showing how to sort in general, I'd provide en example that is possible in both Before and After to show how to translate from V1 to V2. We might need both.

KiewanVillatel (approved):

PR #4759, opened 2026-07-07
docs: add recursion_limit formula for cyclical graphs

harshitayadavv:
The current page explains how to raise `recursion_limit` via the config object, but doesn't explain that the limit counts individual node executions rather than "rounds" or "iterations" of a cycle, and gives no way to calculate an appropriate value for a given graph.

This is a common source of confusion for anyone building cyclical/multi-round agent graphs,since it's easy to underestimate how many steps a multi-round loop actually needs.

This PR adds a short formula (`nodes_per_cycle × max_iterations + buffer`) with a worked example directly on the `GRAPH_RECURSION_LIMIT` error page, since that's where users land when they hit this error. Added to both the Python and JS tabs for parity.

### DIFF
--- a/src/oss/langgraph/errors/GRAPH_RECURSION_LIMIT.mdx
+++ b/src/oss/langgraph/errors/GRAPH_RECURSION_LIMIT.mdx
@@ -54,13 +54,13 @@ However, complex graphs may hit the default limit naturally.
 graph.invoke({...}, {"recursion_limit": 1000})
 ```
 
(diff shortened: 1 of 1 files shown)


harshitayadavv:
Hi @npentrel , @lnhsingh  this is a small docs-only addition (formula for calculating recursion_limit in cyclical graphs). Happy to adjust wording/format if needed!

PR #4775, opened 2026-07-07
docs: update evaluator SDK Python async examples

sineha-mani:
## Summary
- Update Python evaluator SDK examples to use `await` with `Client().evaluators` methods.
- Add notes clarifying that current Python evaluator methods are async and should run inside an async function.
- Adjust list examples to await a page or async-iterate the paginator.

## Test plan
- `git diff --check`

Made with [Cursor](https://cursor.com)


fjmorris (approved):
Approved pending following change:

In the Prerequisites section, after:
```
In the following examples, replace placeholders such as <evaluator-uuid> with the corresponding information from LangSmith.
```

add something like:
```
All python async examples on this page assume they run inside `async def main(): ... asyncio.run(main())`, as shown in the [create a code evaluator example](#code-evaluator).
```

PR #4805, opened 2026-07-09
docs: add exceptions table and not-found example to SmithDB migration guide

KiewanVillatel:
## Summary
- Add a dedicated **Exceptions** section (Python/TypeScript) to the SmithDB SDK migration guide (`src/langsmith/smithdb-sdk-migration.mdx`), mapping the legacy `langsmith.utils` exception classes to the new Stainless-generated ones.
- Add a **"Handle a run that is not found"** example to `runs-retrieve.mdx`
- Add note on the fact that python methods are now async
- Fix some code samples

## Notable implementation details
- Java/Kotlin and Go are unaffected by the exceptions rename — both SDKs were already Stainless-generated before SmithDB, so `NotFoundException` / `*langsmith.Error` (checked via `errors.As`) are unchanged between the old and new methods.


src/code-samples/langsmith/smithdb-migration/runs-retrieve-not-found-after.go
```diff
@@ -0,0 +1,58 @@
...
+if sessErr != nil {
+	panic(sessErr.Error())
+}
+projectID = sessions.Items[0].ID
+
+b := make([]byte, 16)
+_, _ = rand.Read(b)
+b[6] = (b[6] & 0x0f) | 0x40
+b[8] = (b[8] & 0x3f) | 0x80
+runID = fmt.Sprintf("%x-%x-%x-%x-%x", b[0:4], b[4:6], b[6:8], b[8:10], b[10:16])
```
QuentinBrosse: nit: should we use https://pkg.go.dev/uuid ?

src/langsmith/smithdb-sdk-migration.mdx
```diff
@@ -55,6 +55,52 @@ If a call site or parameter is not covered by the guide, stop and ask rather
 than guessing.
 ```
 
+## Exceptions
+
+For Python and TypeScript, the SmithDB-backed methods raise Stainless-generated exceptions instead of the legacy `langsmith.utils` exception classes. Java and Go are unaffected—both SDKs were already Stainless-generated before SmithDB.
```
QuentinBrosse: Should we also include Java (raising `NotFoundException` and maybe more).

Also, I wonder if we could just provide a link to the reference: https://reference.langchain.com?

src/langsmith/smithdb-sdk-migration.mdx
```diff
@@ -55,6 +55,52 @@ If a call site or parameter is not covered by the guide, stop and ask rather
...
+
+<Tabs>
+  <Tab title="Python">
+    | Before (`langsmith.utils`) | After (`langsmith._openapi_client`) | Notes |
```
QuentinBrosse: Should we export `langsmith._openapi_client` I think it's already done, but wanted to double-check

src/code-samples/langsmith/smithdb-migration/runs-retrieve-not-found.py
```diff
@@ -0,0 +1,48 @@
+import uuid
+
+# :snippet-start: runs-retrieve-not-found-before-py
+# :codegroup-tab: Before
```
QuentinBrosse, line 4: Q: Why do we have a single file for Python but 2 before/after files for the other languages? I like one file format! I guess it's not possible for every language (e.g. you cannot put imports in the middle of the file in Java/Go?)

src/snippets/langsmith/smithdb-migration/runs-retrieve.mdx
```diff
@@ -642,3 +656,68 @@ asyncio.run(main())
   </Tab>
 </Tabs>
 
+#### Handle a run that is not found
```
QuentinBrosse: nit:


```suggestion
#### Handle a not-found run
```

src/langsmith/smithdb-sdk-migration.mdx
KiewanVillatel: Good catch, the doc is outdated, now errors are directly exported and available under `langsmith` (not `langsmith._openapi_client`). Will fix the doc.

src/langsmith/smithdb-sdk-migration.mdx
KiewanVillatel: For others nothing changed. But yes let me rework this part and move the comment under language selection.

src/code-samples/langsmith/smithdb-migration/runs-retrieve-not-found.py
KiewanVillatel: Good catch, and it's not new to this PR: every example in this migration guide already follows this split (`runs-query-*`, `runs-retrieve-basic`, `runs-retrieve-by-id`, etc.) — Python is always one file, Go/Kotlin/TS/sh always get separate before/after files.

The reason is how `make test-code-samples` executes each sample: it runs the *whole file* as a real program (`go run` for Go, JBang for Kotlin), so each file needs exactly one `package main`/one `fun main`/one import block. Two versions can't coexist in a single Go or Kotlin file. Python is just executed top-to-bottom as a script, so both before and after sections can share one file without conflicting.

TS does not have that hard constraint, but we keep it in two files to match the pattern used across the rest of the guide (and to avoid the before/after versions clashing on top-level `const`/`let` names like `client` or `runId`).

src/code-samples/langsmith/smithdb-migration/runs-retrieve-not-found-after.go
KiewanVillatel: Done

src/code-samples/langsmith/smithdb-migration/runs-retrieve-not-found.py
QuentinBrosse: ok, it's nit anyway, i was just curious. thanks!

src/langsmith/smithdb-sdk-migration.mdx
KiewanVillatel: Done

src/langsmith/smithdb-sdk-migration.mdx
KiewanVillatel: Done

QuentinBrosse (approved):

PR #4813, opened 2026-07-09
docs(livekit): document instrument_session for realtime user transcript

carolinedivittorio:
Add a section on capturing the realtime (speech-to-speech) user transcript via `processor.instrument_session`, and note it must not be used with the STT/LLM/TTS cascade (where it would double-count the user's turns).

### DIFF
--- a/src/langsmith/trace-with-livekit.mdx
+++ b/src/langsmith/trace-with-livekit.mdx
@@ -70,7 +70,7 @@ async def my_agent(ctx: agents.JobContext):
     await session.start(room=ctx.room, agent=Agent(instructions="You are a helpful assistant."))
 ```
 
-This works for both the STT/LLM/TTS cascade and speech-to-speech models: if you choose to build the `AgentSession` with a realtime model (for example, `lk_openai.realtime.RealtimeModel(...)`), the tracing setup is unchanged.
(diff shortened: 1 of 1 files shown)


src/langsmith/trace-with-livekit.mdx
```diff
@@ -70,7 +70,7 @@ async def my_agent(ctx: agents.JobContext):
     await session.start(room=ctx.room, agent=Agent(instructions="You are a helpful assistant."))
 ```
 
-This works for both the STT/LLM/TTS cascade and speech-to-speech models: if you choose to build the `AgentSession` with a realtime model (for example, `lk_openai.realtime.RealtimeModel(...)`), the tracing setup is unchanged.
+This works for both the STT/LLM/TTS cascade and speech-to-speech (realtime) models. Realtime models (for example, `lk_openai.realtime.RealtimeModel(...)`) need one extra call to capture the user's transcript — see [When using LiveKit with a realtime model](#when-using-livekit-with-a-realtime-model).
```
katmayb:
```suggestion
This works for both the STT/LLM/TTS cascade and speech-to-speech (realtime) models. Realtime models (for example, `lk_openai.realtime.RealtimeModel(...)`) need one extra call to capture the user's transcript. Refer to [When using LiveKit with a realtime model](#when-using-livekit-with-a-realtime-model).
```

katmayb (commented):
LGTM, just a small change for the linter.

katmayb (approved):

PR #4837, opened 2026-07-10
docs: clarify trace retention behavior

sineha-mani:
## Summary
- Clarify when feedback, evaluators, automation rules, and annotation queue operations can extend trace retention.
- Update stale docs that said all annotation queue adds and multi-turn evaluators automatically upgrade traces.
- Document that manual UI feedback/notes and manual annotation queue adds do not change retention by default, while SDK/API feedback may still extend retention unless opted out.

## Test Plan
- [x] Docs-only change; reviewed rendered Markdown diff locally.


fjmorris (approved):
pushed a commit with my docs revisions. my notes:
[docs-review.md](https://github.com/user-attachments/files/30239657/docs-review.md)

PR #4958, opened 2026-07-18
Add ShannonBase vector store integration

apoorva-01:
Hey! This adds `langchain-shannonbase`, a vector store built on MySQL 9's native `VECTOR` type. The same code works on ShannonBase, plain self-hosted MySQL 9, and HeatWave, since they all share the `VECTOR` / `STRING_TO_VECTOR` / `DISTANCE` functions.

Why bother? 
Right now the only MySQL vector store around is locked to Google Cloud SQL, so if your data's in plain MySQL 9 or ShannonBase there just wasn't one. ShannonBase even has "LangChain integration" sitting on their wishlist, so this scratches that itch.

A few notes:
- it's on PyPI: https://pypi.org/project/langchain-shannonbase/
- passes the standard `VectorStoreIntegrationTests` (runs offline in CI on 3.9 / 3.11 / 3.12)
- does metadata filtering, MMR, and `get_by_ids`

The PR just adds the doc page (`src/oss/python/integrations/vectorstores/shannonbase.mdx`) and a `packages.yml` entry. If it qualifies for `highlight` as a new integration that'd be lovely, but totally fine if not.

Thanks for taking a look!


apoorva-01:
@mdrxy - It's ready for review whenever you get a chance. Thanks!

npentrel (approved):

PR #4960, opened 2026-07-18
docs(deepagents): document tool exclusion as a context compression technique

Engr-FaizanAli:
## Overview

Documents `HarnessProfile(excluded_tools=...)` as a third context-compression technique alongside offloading and summarization, since it does the same job (reducing what's sent to the model per turn) but wasn't cross-referenced from that section.

The "Context compression" section on the context-engineering page only listed offloading and summarization as built-in mechanisms for reducing context/token usage, even though `HarnessProfile`'s `excluded_tools` does the same job (removing tool schemas from every request) and was previously only documented as a "tool visibility" feature in the Tool prompts section with no cross-reference here.

Adds a third "Tool exclusion" card and subsection alongside Offloading and Summarization, showing the `register_harness_profile` usage and noting it's a static reduction (decided ahead of time) rather than a dynamic, threshold-triggered one like the other two.

## Type of change

## Related issues/PRs

- GitHub issue: langchain-ai/deepagents#616

## Checklist

## Additional notes

No `docs.json` changes needed — this adds a subsection/anchor to an existing page, not a new page. Verified with `uv run python scripts/check_cross_refs.py` — no new unresolved references introduced (the one pre-existing failure, in an unrelated JS provider page, is untouched by this change).

The code example mirrors a pattern verified in real use: excluding 4 unused built-in tools (`write_file`, `edit_file`, `execute`, `write_todos`) from a read-only search agent measured a ~20-30% reduction in total tokens on a real multi-turn task, with no change in agent behavior since those tools were never being called.

### DIFF
--- a/src/oss/deepagents/context-engineering.mdx
+++ b/src/oss/deepagents/context-engineering.mdx
@@ -233,6 +233,9 @@ The following techniques are the built-in mechanisms to ensure the context passe
   <Card title="Summarization" icon="article" href="#summarization">
     Old messages are compressed into an LLM-generated summary when limits are approached.
   </Card>
+  <Card title="Tool exclusion" icon="filter" href="#tool-exclusion">
+    Drop unused built-in tools so their schemas aren't sent to the model on every turn.
+  </Card>
 </CardGroup>
 
 ### Offloading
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thank you! I've moved the content slightly as where you'd put it it was next to things that happen automatically.

PR #5042, opened 2026-07-22
docs: clarify N_JOBS_PER_WORKER governs run concurrency, not request concurrency

victorm-lc:
## Why

On a customer call it came up that users believe their deployments are limited in **how many requests they can serve concurrently**, because `N_JOBS_PER_WORKER` is the only setting our docs tie to the word "concurrency," and the scaling docs blur "runs" and "requests." This PR makes the distinction explicit.

## What actually happens

Verified against the `langgraph-api` source:

- `N_JOBS_PER_WORKER` (default 10) is a token/semaphore over the background run-execution loop. It bounds how many **runs a single queue worker executes concurrently** — nothing more.
- It does **not** limit API request concurrency. The API server is a single async `uvicorn` process with no `--workers` flag and no semaphore keyed on `N_JOBS_PER_WORKER` on the request path. Request capacity scales with API-server replicas and is bounded by the async event loop and the Postgres pool (`LANGGRAPH_POSTGRES_POOL_MAX_SIZE`, default 150).
- `POST /threads/{id}/runs` persists a `pending` run and returns immediately, without consuming a worker token. Excess runs wait in the queue until a run slot frees.
- In production, API servers (`N_JOBS_PER_WORKER=0`) and queue workers (`N_JOBS_PER_WORKER>0`) are separate, independently scalable container pools.

## Changes

- **`agent-server-scale.mdx`**: new `## Request vs. run concurrency` section up top; clarified the throughput math sizes run-execution capacity (queue workers); added a note to the example-configurations table.
- **`agent-server.mdx`**: reworded the `N_JOBS_PER_WORKER` line so it clearly means run execution, not request handling; fixed the cross-link (`env-var-cloud` → `env-var-self-hosted`, matching the scaling page).
- **`snippets/langsmith/env-vars/shared.mdx`**: expanded the terse `N_JOBS_PER_WORKER` definition with the runs-vs-requests distinction and a link to the scaling guide (renders on both cloud and self-hosted env-var pages).
- **`scalability-and-resilience.mdx`**: named the queue worker and `N_JOBS_PER_WORKER` explicitly so "concurrent runs" is not read as request capacity.

## Review focus

The example-configurations table in `agent-server-scale.mdx` keeps its "requests per second" framing; I added a lead-in sentence rather than restructuring it. Worth a careful read to confirm the wording lands.

## Verification

- `make lint_prose` on all four files: clean.
- `make broken-links`: no broken links (new same-page anchor `#request-vs-run-concurrency` and the updated cross-links resolve).

## AI involvement

Drafted with Claude Code (research against `langgraph-api` source + docs edits); reviewed by @victorm-lc.


src/langsmith/agent-server-scale.mdx
```diff
@@ -13,6 +13,15 @@ If you're not yet familiar with how API servers and queue workers operate at the
 
 For [Cloud](/langsmith/cloud-platform-features#scaling), the platform autoscales automatically and the Helm configurations below do not apply.
 
+## Request vs. run concurrency
+
+Two independent kinds of concurrency determine how the Agent Server scales, and they are controlled separately:
+
+- **Request concurrency** is how many API requests (creating runs, reading thread state, streaming results) the deployment serves at once. API servers handle requests asynchronously and scale horizontally with the number of API server replicas. Request-serving capacity is not capped by [`N_JOBS_PER_WORKER`](/langsmith/env-var-self-hosted).
```
jdrogers940: nit: Maybe just leave N_JOBS_PER_WORKER off here? Makes sense in the context of this PR, but I think to new eyes, its cleaner to just focus on what controls this instead of what doesn't.
```suggestion
- **Request concurrency** is how many API requests (creating runs, reading thread state, streaming results) the deployment serves at once. API servers handle requests asynchronously and the request concurrency scales horizontally with the number of API server replicas.
```

src/langsmith/agent-server-scale.mdx
```diff
@@ -13,6 +13,15 @@ If you're not yet familiar with how API servers and queue workers operate at the
...
+- **Run concurrency** is how many runs execute at once. A single queue worker executes up to `N_JOBS_PER_WORKER` runs concurrently (default 10), and total run capacity scales with the number of queue workers multiplied by `N_JOBS_PER_WORKER`.
```
jdrogers940: I think we can make this more explicit because this is a hard cap.
```suggestion
- **Run concurrency** is how many runs execute at once. A single queue worker executes up to `N_JOBS_PER_WORKER` runs concurrently (default 10). Run concurrency is capped at the number of queue workers multiplied by `N_JOBS_PER_WORKER`.
```

jdrogers940: Maybe we can link out to the N_JOBS_PER_WORKER env var here too?

src/langsmith/agent-server-scale.mdx
```diff
@@ -13,6 +13,15 @@ If you're not yet familiar with how API servers and queue workers operate at the
...
+
+Creating a run is a fast write request: the API server persists a pending run and returns immediately, without waiting for the run to execute. If every run slot is busy, additional runs wait in the queue until a slot frees. Raising `N_JOBS_PER_WORKER` or adding queue workers increases run throughput; it does not change how many requests the deployment can serve concurrently.
```
jdrogers940: Do we think there's a gap in the docs around the queuing section explaining this? Maybe good to link out to the architecture there.

jdrogers940 (approved):

src/langsmith/agent-server-scale.mdx
victorm-lc: Done. Dropped the `N_JOBS_PER_WORKER` clause from this bullet and applied your suggested wording. The point that it does not cap request concurrency still lives in the closing paragraph, so nothing is lost.

src/langsmith/agent-server-scale.mdx
victorm-lc: Done. Switched to hard-cap wording (Run concurrency is capped at the number of queue workers multiplied by `N_JOBS_PER_WORKER`) and moved the `N_JOBS_PER_WORKER` env var link onto this bullet, which is now its first mention.

src/langsmith/agent-server-scale.mdx
victorm-lc: Linked `queue` out to the Run execution lifecycle section on the Agent Server page, which documents enqueue, pending, pickup, and slot release. Kept this to a link rather than expanding the queue docs for now.

PR #700, opened 2025-09-30
updates to langchain/messages guide

ccurme:

PR #5056, opened 2026-07-23
docs: revert SystemPromptConfig docs [langchain-ai/deepagents#4969]

imnishitha:
## Description
`langchain-ai/deepagents#4969` reverted the alpha `SystemPromptConfig` (`prefix`/`base`/`suffix` dict) feature, so `system_prompt=` is a plain `str | SystemMessage` again. This drops the `SystemPromptConfig` guidance from the Python-side "System prompt" section entirely and fixes the section intro, which incorrectly claimed a built-in base prompt ships by default. Also resyncs the Python function-signature snippet (stale `SystemPromptConfig` reference) and drops the now-stale `SystemPromptConfig` link-map entry. No changelog changes.

`langchain-ai/deepagentsjs` mirrored the revert (#700, #703), but kept `SystemPromptConfig` in its type union as a `@deprecated` compatibility shim slated for removal in the next major release. Dropped it from the JS signature snippet to match. **Note for reviewers:** that snippet is normally auto-generated by `scripts/sync_deepagents_signatures.py` from the live API reference site, which still lists the deprecated type — don't blindly re-run that script against this one file until deepagentsjs's next major release actually removes the shim, or it'll reintroduce `SystemPromptConfig`.

## Test Plan
- [x] `make check-cross-refs`
- [x] `make lint_prose FILES=...` on changed `.mdx` files

Opened collaboratively by Nishitha Madhu and open-swe


src/oss/deepagents/customization.mdx
```diff
@@ -177,7 +177,13 @@ For detailed configuration options including stdio servers, OAuth authentication
 
 ## System prompt
 
-Deep Agents ship with a built-in base system prompt that teaches the agent how to use the harness scaffolding (planning, filesystem tools, subagents). Pass `system_prompt=` to prepend your own instructions before that base prompt:
+:::python
+Deep Agents does not add an authored base system prompt by default: `create_deep_agent` starts from an empty base, so a deep agent ships with no persona or task guidance out of the box. Pass `system_prompt=` to add your own instructions, or pass `system_prompt=BASE_AGENT_PROMPT` to restore the SDK's built-in persona and task-guidance prompt, or set it on a [profile](/oss/deepagents/profiles#harness-profiles) as `base_system_prompt`:
```
ccurme: (nit) I don't think we should start with a negative here ("we don't do X"). Just say that you can use `system_prompt=...` to specify the system prompt.

ccurme: I wouldn't talk about "restoring" here, that context is only relevant for the transition from 0.6 to 0.7. Managing 0.7 changes should go in changelog.

src/oss/deepagents/customization.mdx
```diff
@@ -193,31 +199,44 @@ Besides a string, the main agent also accepts a @[`SystemMessage`] with structur
...
+### Prompt assembly
+
 :::python
-<Note>
-`SystemPromptConfig` requires `deepagents>=0.7.0a6`.
-</Note>
+Deep Agents builds the system prompt from up to three named parts so that caller-supplied instructions and any model-specific [profile](/oss/deepagents/profiles) overrides can coexist with predictable precedence. Without this layering, a profile suffix tuned for Claude (for example) could overwrite or be overwritten by your `system_prompt=` argument depending on call order; the named slots make the ordering explicit and stable.
 
-For full control over prompt assembly, pass a @[`SystemPromptConfig`] dict with `prefix`, `base`, and `suffix` keys:
+The three named parts (each may be absent):
```
ccurme: The PR you're more or less reverting here did two things:
1. Add docs on system prompt configurability
2. Simplify this section on the system prompt. This section focused on harness profiles. It was concept-heavy, without snippets for demonstrating how to use them to alter the prompt. So I moved this to the harness profiles page.

Your PR is un-doing both of these things, we should instead just update the guidance on configuration.

npentrel (approved):

src/oss/deepagents/customization.mdx
```diff
@@ -188,38 +188,11 @@ Deep Agents ship with a built-in base system prompt that teaches the agent how t
 :::
 
 <Note>
-Besides a string, the main agent also accepts a @[`SystemMessage`] with structured [content blocks](/oss/langchain/messages#standard-content-blocks); Deep Agents preserve those blocks and append the built-in base prompt ([subagent](/oss/deepagents/subagents) dictionary specs remain strings).
+Besides a string, the main agent also accepts a @[`SystemMessage`] with structured [content blocks](/oss/langchain/messages#standard-content-blocks); Deep Agents preserve those blocks ([subagent](/oss/deepagents/subagents) dictionary specs remain strings).
 </Note>
 
 When middleware adds special tools, like the filesystem tools, it appends its own guidance to the system prompt at runtime.
```
ccurme: Is this true?

sydney-runkle: yup

ccurme: are you sure? it looks like for StateBackend for example FilesystemMiddleware appends nothing to the system prompt https://github.com/langchain-ai/deepagents/blob/f1727c785c44809e7b1b7ed8c5d0b31c560ae3ad/libs/deepagents/deepagents/graph.py#L629-L636

imnishitha: verified - by default mw appends nothing (no custom system_prompt and no execution support)

ccurme (approved):

PR #5060, opened 2026-07-23
docs: update self-hosted SBOM verification (CycloneDX + drop version qualifier)

langchain-infra:
Updates the self-hosted **Mirroring images** guide's SBOM section (`self-host-mirroring-images.mdx`).

### Switch SBOM verification to CycloneDX
Released images now carry **CycloneDX** SBOM attestations instead of SPDX. The SPDX predicate for the consolidated `langsmith-backend` image exceeded Rekor's transparency-log entry size limit, so signing was failing; CycloneDX is roughly half the size and keeps the attestation in the tlog. See langchainplus signing change (`sign_dockerhub_release_image.sh` → `--type cyclonedx`).

Doc changes:
- `cosign verify-attestation --type spdxjson` → `--type cyclonedx` (both the tag-level and per-digest examples)
- Prose "SPDX" / "SPDX 2.3 document" → "CycloneDX"
- Extracted file `langsmith-backend.spdx.json` → `.cdx.json` (Grype/Trivy auto-detect CycloneDX by content, so the scanner commands are unchanged otherwise)

### Drop version qualifier (pre-existing commit)
Removes the `0.15.12` version qualifier from the tag-level verify example.

> Note: this documents the format used by releases that carry the CycloneDX signing change. Older/current stable images still ship SPDX until that change reaches a stable release.

### DIFF
--- a/src/langsmith/self-host-mirroring-images.mdx
+++ b/src/langsmith/self-host-mirroring-images.mdx
@@ -237,7 +237,7 @@ cosign download attestation docker.io/langchain/langsmith-backend:<tag>
 
 Released images also carry signed SPDX software bill of materials (SBOM) attestations, one per architecture in the image index.
 
-On `0.15.12` and later, the per-architecture SBOMs are also attached to the multi-architecture index digest, so you can verify against a bare tag directly:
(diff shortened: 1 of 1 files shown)


saiannam-dev (approved):
LG!

PR #5107, opened 2026-07-26
correct set_llm_cache import in 4 cache/provider docs

SpiliosDimakopoulos:
from langchain.globals import set_llm_cache raises ImportError on langchain>=1.0 - langchain.globals was removed entirely. Fixed all 7 occurrences across cratedb.mdx, astradb.mdx, redis.mdx, and redis_llm_caching.mdx.

Used langchain_core.globals (rather than langchain_classic.globals, which also works) to match the import path already used correctly elsewhere on the site - e.g. couchbase.mdx, microsoft.mdx, and mongodb_atlas.mdx - so the fix is consistent with existing convention rather than introducing a second valid-but-inconsistent path.


npentrel:
There don't appear to be changes in this PR - closing for now. Please feel free to reopen if needed

PR #5119, opened 2026-07-27
fix: docs mv does not update site-absolute links when moving files

SpiliosDimakopoulos:
_rewrite_links() (used by `docs mv` to keep links intact when a file is
relocated) joined every candidate URL with `md_file.parent / url`. But
joining a Path with an absolute string discards the left-hand side
(`Path("/a/b") / "/c/d" == Path("/c/d")`), so a site-absolute link like
`/oss/langchain/agents` -- the convention used throughout this docs
site -- was silently resolved against the filesystem root instead of
docs_root. It could never match the file being moved, so the link was
left untouched (and broken) after the move.

A repo-wide check found 7,872 links using this absolute `/oss/...` /
`/langsmith/...` style versus zero using relative `../file.mdx` style,
so this affected effectively every internal link in the site -- `docs
mv` was not actually keeping links intact for the link style the docs
actually use.

Added `_resolve_site_absolute_url()`, which maps a site-absolute URL
back to its source file the same way Mintlify serves it (strip the
leading slash, try `.mdx`/`.md` under docs_root), and use it whenever
a matched URL starts with `/`. Relative-link handling (`_rewrite_links`
for other cases, `_update_internal_links_in_moved_file`, `drop_suffix_
from_links`) is unchanged -- those already correctly skip absolute
links since their own resolution doesn't depend on the referencing
file's location.

Added three regression tests: an absolute link gets rewritten,
anchors on absolute links are preserved through the rewrite, and
absolute links that are unrelated to the moved file or don't resolve
to any file at all (e.g. Mintlify-generated API reference pages with
no corresponding source file) are left untouched rather than raising.

Verified: full test suite passes locally with no failures, plus ruff
format, ruff check, ty, and codespell all clean on the changed files.

### DIFF
--- a/commit-msg.txt
+++ b/commit-msg.txt
@@ -1,0 +1,23 @@
+fix: docs mv does not update site-absolute links when moving files
+
+_rewrite_links() joined every URL with `md_file.parent / url`, but
+joining a Path with an absolute string discards the left-hand side
+(`Path("/a/b") / "/c/d" == Path("/c/d")`). So for a site-absolute link
+like `/oss/langchain/agents` -- the convention used throughout this
+docs site -- the code silently resolved it against the filesystem
+root instead of docs_root, never matched the file being moved, and
+left the link untouched. A repo-wide check found 7,872 links using
+this absolute style and zero using relative `../` style, so this
+affected effectively every link in the site.
+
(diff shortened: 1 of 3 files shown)


npentrel:
we don't actually use this command I'm afraid, so I will close this PR

PR #5142, opened 2026-07-28
remove todolist middleware from default stack

npentrel:


src/code-samples/deepagents/data-analysis.py
```diff
@@ -87,6 +88,7 @@ def slack_send_message(text: str, file_path: str | None = None) -> str:
     tools=[slack_send_message],
     backend=backend,
     checkpointer=checkpointer,
+    middleware=[TodoListMiddleware()],
```
sydney-runkle, line 91: does it need?

sydney-runkle (approved):
i think we have some diagrams showing "planning" as a capability core to the harness?

src/code-samples/deepagents/data-analysis.py
npentrel: probably not necessarily but if we suggest people use it for complex tasks I think it's fair to add it here?

src/oss/deepagents/overview.mdx
```diff
@@ -293,15 +296,31 @@ For other providers, see [Middleware integrations](/oss/integrations/middleware#
...
 The delegation component enables agents to break large problems into smaller, parallelizable units of work. It has two layers:
 
-- **[Task planning](#task-planning)**: a built-in `write_todos` tool for structured task tracking
+- **[Task planning](#task-planning)**: an opt-in `write_todos` tool for structured task tracking
 - **[Subagents](#subagents)**: ephemeral child agents that handle isolated subtasks
 
 ### Task planning
 
-The harness provides a `write_todos` tool that lets agents maintain a structured task list during execution.
+Task planning is an opt-in harness capability that lets agents maintain a structured task list during execution. Planning is often useful for:
```
npentrel:
```suggestion
Task planning is an opt-in harness capability that lets agents maintain a structured task list during execution.

Starting in v0.7 task planning is opt-in only. In earlier versions, task planning middleware was included by default.

Planning is often useful for:
```

PR #5151, opened 2026-07-28
docs: swap filesystem system_prompt example for tools allowlist

imnishitha:
## Description
In the Deep Agents customization page, replaces the "Customize the filesystem instructions in the system prompt" `FilesystemMiddleware` override example with one demonstrating the `tools` allowlist param instead, and removes the "Customize the task tool description" `SubAgentMiddleware` example.

## Test Plan
- [ ] Confirmed no other docs pages link to the removed/renamed accordion anchors
- [ ] `markdownlint` on the changed file shows only pre-existing issues (unrelated to this change)

Opened collaboratively by Nishitha Madhu and open-swe


src/oss/deepagents/customization.mdx
```diff
@@ -476,8 +476,12 @@ Declarative subagents defined via `subagents=` do not inherit the main agent's m
...
         ```
     </Accordion>
-    <Accordion title="Customize the filesystem instructions in the system prompt" icon="file-text">
-        Override @[`FilesystemMiddleware`] with a `system_prompt` to replace the filesystem-specific instructions it appends to the system prompt in place of the dynamically generated default.
+    <Accordion title="Restrict the enabled filesystem tools" icon="filter">
+        <Note>
+        The `tools` allowlist on `FilesystemMiddleware` requires `deepagents>=0.7.0a4`.
+        </Note>
+
+        Override @[`FilesystemMiddleware`] with a `tools` allowlist to expose only a subset of the filesystem tools to the model, instead of the full default set. Any built-in filesystem tool left out of the list is removed from both the model's tool list and the middleware's dynamic system prompt section.
```
ccurme: Why are we talking about a dynamic system prompt here? Aren't we no longer doing this?

ccurme (approved):

PR #5166, opened 2026-07-28
fix: inline LangGraph Agent Server note to fix broken link

### LINKED ISSUE #5057: [langchain]: Fix broken Agent Server link in frontend pattern docs
### Type of issue

issue / bug

### Language

JavaScript

### Description

The **LangGraph Agent Server** link on [Branching chat](https://docs.langchain.com/oss/javascript/langchain/frontend/branching-chat) points to:

`/oss/javascript/langchain/langgraph/local-server`

That URL returns 404. The intended page is:

`/oss/javascript/langgraph/local-server`

The same shared note is used by the `join-rejoin`, `message-queues`, and `time-travel` frontend pattern pages, so those pages are susceptible to the same broken path in both language builds.

#### Root cause

`src/snippets/oss/requires-langgraph-server.mdx` uses `/oss/langgraph/local-server`. During snippet processing, `_process_snippet_markdown_file` converts that URL to `../langgraph/local-server`. Mintlify resolves the relative URL from nested consumers such as `/oss/javascript/langchain/frontend/branching-chat`, producing the invalid `/oss/javascript/langchain/langgraph/local-server` path.

#### Proposed fix

For the minimal safe fix, inline the requirement note in the four consuming frontend pages and keep the source URL as `/oss/langgraph/local-server`. The normal page build can then apply `_rewrite_oss_links` with the target language and produce `/oss/{python|javascript}/langgraph/local-server`. Remove the now-unused shared snippet.

Alternatively, make snippet processing language- and consumer-path-aware rather than converting every `/oss/` link to a fixed `../` relative path.

Add regression coverage that builds a nested frontend page for both targets and asserts that the rendered links are:

- `/oss/python/langgraph/local-server`
- `/oss/javascript/langgraph/local-server`

#### Acceptance criteria

- The Agent Server link works from all four frontend pattern pages in both Python and JavaScript docs.
- The build has regression coverage for an `/oss/` link inside a snippet imported by a nested page.

SpiliosDimakopoulos:
## Overview
Fixes #5057. The LangGraph Agent Server link on the branching-chat,
join-rejoin, message-queues, and time-travel frontend pattern pages
returned 404, e.g. /oss/javascript/langchain/langgraph/local-server
instead of /oss/javascript/langgraph/local-server.

Root cause (as diagnosed in #5057): the shared snippet
src/snippets/oss/requires-langgraph-server.mdx uses /oss/langgraph/local-server.
During snippet processing, that URL gets converted to a fixed ../
relative path, which Mintlify then resolves incorrectly from these
nested pages.

Fix: inline the note directly in the four consuming pages (removing
the RequiresLanggraphServer import/usage, keeping UseStreamTypeInference
untouched) and delete the now-unused shared snippet, so the standard
per-language link rewriting applies correctly - this is the "minimal
safe fix" proposed in the issue itself.

## Type of change

## Checklist
- [x] I have used root relative paths for internal links
- [ ] N/A - no navigation changes needed

### DIFF
--- a/pipeline/core/builder.py
+++ b/pipeline/core/builder.py
@@ -253,6 +253,36 @@ class DocumentationBuilder:
             # Return original content if there's an error
             return content
 
+    def _rewrite_snippet_imports_for_language(
+        self, content: str, target_language: str
+    ) -> str:
+        """Point MDX snippet imports at language-specific copies under /snippets/{lang}/.
+
+        Snippet markdown is emitted as absolute, language-prefixed /oss/ links in
+        ``build/snippets/{python|javascript}/...``. Versioned pages must import
+        those copies so nested consumers (e.g. langchain/frontend/*) resolve
+        correctly. Already-prefixed imports are left unchanged.
+
+        Args:
+            content: Markdown/MDX source that may contain snippet imports.
+            target_language: Target language ("python" or "js").
+
+        Returns:
+            Content with rewritten snippet import paths.
+        """
+        lang_name = self.language_url_names[target_language]
+        pattern = r"""(from\s+)(['"])(/snippets/[^'"]+\.mdx?)\2"""
+
+        def rewrite_import(match: re.Match) -> str:
+            """Rewrite a single snippet import if it is not already language-scoped."""
+            prefix, quote, path = match.group(1), match.group(2), match.group(3)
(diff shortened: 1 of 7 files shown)


npentrel (approved):
Thanks for reporting - I fixed the underlying issue instead

PR #5176, opened 2026-07-29
docs(langsmith): document the child_runs replacement in the SmithDB migration guide

QuentinBrosse:
## Why

The SmithDB migration guide currently tells readers that `client.read_run(run_id, load_child_runs=True)`, `run.child_runs`, and `run.child_run_ids` have **No equivalent** in v2. They do have one: fetch every run in the trace with `traces.list_runs`, then filter on `parent_run_ids`. Anyone reading that table today concludes the capability was dropped and has no path forward.

## What changed

A new `#### Load a run's child runs` example in the runs-retrieve section, with tested before/after samples in Python and TypeScript, plus the three table rows now pointing at it instead of saying "No equivalent".

`parent_run_ids` is the full ancestor chain, so filtering on it yields every descendant at any depth (what `child_run_ids` held). Grouping by the last entry recovers the direct children (what `child_runs` held). Both shapes are shown.

Java, Go, and cURL get a tab explaining there is nothing to migrate, because loading children in one call was only ever a client-side fan-out in the Python and TypeScript SDKs.

## Areas needing careful review

**Two existing inaccuracies corrected.** Both were found while verifying this change, and both are the same underlying fact:

1. The cURL query-parameter table listed `load_child_runs` as a v1 param. It never was one: `GET /api/v1/runs/{run_id}` accepts `run_id, session_id, start_time, exclude_s3_stored_attributes, exclude_serialized, include_messages`, and `load_child_runs` appears zero times in `langsmith-platform-openapi.json`. The Python client sends no params on that GET and calls `_load_child_runs` separately (`langsmith/client.py:4017`). Row removed.
2. The TypeScript response table named the v1 fields `run.childRunIds` / `run.childRuns`. The actual properties are snake_case (`child_run_ids`, `child_runs`). Corrected.

**The samples seed their own data.** The `default` project holds no nested traces, so rather than depend on data they do not control, all three samples trace a small nested call tree in a hidden `:remove-start:` block and poll until it is readable. The Python sample then asserts the v2 result equals the v1 result on that same trace, so the documented replacement is verified equivalent rather than assumed.

**Placement.** The example sits after "Handle a not-found run". It documents a `runs.retrieve` migration but the replacement calls `traces.list_runs`, so an argument exists for putting it in the Traces section instead.

## Verification

- `make test-code-samples` on all three samples: 3/3 passed. Python run repeatedly to check for ingestion flakiness: 3/3 consecutive passes.
- Equivalence confirmed on the seeded trace: v1 reports 2 direct children and 3 descendants; v2 filtering reports the same 3 descendants and the same 2 direct children, matched by run ID.
- `make lint`: passed (ruff format, ruff check, ty, codespell).
- `make lint_prose`: 0 errors, 0 warnings, 0 suggestions.
- `make broken-links`: no broken links, including the new same-page anchors.
- `make lint_md`: no new findings (remaining ones in this file are pre-existing).

## AI disclosure

Written by Claude Opus 5 via Claude Code, under human direction and review.

### DIFF
--- a/src/code-samples/langsmith/smithdb-migration/runs-retrieve-child-runs.py
+++ b/src/code-samples/langsmith/smithdb-migration/runs-retrieve-child-runs.py
@@ -1,0 +1,153 @@
+# :remove-start:
+import time
+
+from langsmith import Client as _SeedClient
+from langsmith import traceable
+from langsmith.run_helpers import get_current_run_tree, tracing_context
+
+_SEEDED: dict[str, str] = {}
+
+
+@traceable(run_type="llm")
+def _leaf(index: int) -> str:
+    return f"leaf {index}"
+
+
+@traceable
+def _branch() -> str:
+    _leaf(0)
+    return "branch"
+
+
+@traceable(name="docs-child-runs-example")
+def _root() -> str:
+    run_tree = get_current_run_tree()
+    assert run_tree is not None, "tracing is not enabled"
+    _SEEDED["run_id"] = str(run_tree.id)
+    _SEEDED["trace_id"] = str(run_tree.trace_id)
+    _leaf(1)
+    _branch()
+    return "root"
+
+
+def _seed_trace() -> tuple[str, str]:
+    """Trace a small nested call tree and wait until it is readable.
+
+    The `default` project holds no nested traces of its own, so the sample
+    creates one instead of depending on data it does not control. The v1 and v2
+    read paths become consistent at slightly different times, so poll until the
+    v1 path returns the full child tree.
+    """
+    client = _SeedClient()
+    with tracing_context(project_name="default", enabled=True):
+        _root()
+    client.flush()
+
+    run_id, trace_id = _SEEDED["run_id"], _SEEDED["trace_id"]
+    for _ in range(30):
+        try:
+            run = client.read_run(run_id, load_child_runs=True)
+        except Exception:  # noqa: BLE001 - not ingested yet
+            run = None
+        if run is not None and len(run.child_runs or []) == 2:
+            return run_id, trace_id
+        time.sleep(2)
+    raise AssertionError(f"seeded run {run_id} never became readable with children")
+
+
+_RUN_ID, _TRACE_ID = _seed_trace()
+# :remove-end:
+
+# :snippet-start: runs-retrieve-child-runs-before-py
+# :codegroup-tab: Before
+from langsmith import Client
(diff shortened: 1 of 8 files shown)


src/snippets/langsmith/smithdb-migration/runs-retrieve.mdx
```diff
@@ -739,3 +742,45 @@ curl "https://api.smith.langchain.com/api/v1/runs/$RUN_ID" \
...
+  </Tab>
+  <Tab title="Java">
+    Nothing to migrate. Loading a run's children in one call was a convenience implemented in the Python and TypeScript SDKs, which issued a second query on the caller's behalf. The Java SDK never exposed it, and `client.runs().retrieve()` never returned nested child runs. To walk a trace's runs, use `client.traces().listRuns()`.
+  </Tab>
+  <Tab title="Go">
+    Nothing to migrate. Loading a run's children in one call was a convenience implemented in the Python and TypeScript SDKs, which issued a second query on the caller's behalf. The Go SDK never exposed it, and `client.Runs.Get()` never returned nested child runs. To walk a trace's runs, use `client.Traces.ListRuns()`.
+  </Tab>
+  <Tab title="cURL">
+    Nothing to migrate. Loading a run's children in one call was a convenience implemented in the Python and TypeScript SDKs, which issued a second query on the caller's behalf. `GET /api/v1/runs/{run_id}` never accepted a flag for it and never returned nested child runs. To walk a trace's runs, use `GET /v2/traces/{trace_id}/runs`.
+  </Tab>
```
QuentinBrosse, line 782: Be much more concise here. 1 sentence by lang only

src/code-samples/langsmith/smithdb-migration/runs-retrieve-child-runs.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+async def main():
+    client = Client()
+    project = await client.aread_project(project_name="default")
+    run_id = "<run-id>"
+    # `trace_id` is the run's trace ID. A root run is its own trace, so
+    # `trace_id` equals `run_id`. For any other run, read `trace_id` from
+    # `client.runs.retrieve()` first.
+    trace_id = "<trace-id>"
```
QuentinBrosse: Simplify by using only trace_id, don't give all those details, same for JS

KiewanVillatel (approved):

PR #5216, opened 2026-07-30
docs: add session_id in feedback creation

emil-lc:
## Overview

Creating feedback for a run now requires `session_id` / `sessionId`, the UUID of the tracing project (session) that owns the run. #5048 documented this in the SmithDB migration guide, but every feedback example in the regular docs still omits it, so anyone copying one gets a deprecation warning today and a hard error on SmithDB-only deployments.

This updates the six pages that create run-level feedback, plus the feedback field reference.

## Type of change

## Related issues/PRs

- Feature PR: #5048 (documented the same requirement in the SmithDB migration guide)
- SDK: both the warning/error and the object-form `createFeedback` overload used here come from `langsmith-sdk` branch `emil/create-feedback-missing-session-warning-error`, which is **not released yet**. See "Release ordering" below

- Linear issue:
- Slack thread:

## Checklist

- [x] All code examples have been tested and work correctly — every example was executed against a live workspace; see Verification
- [x] I have updated navigation in `src/docs.json` if needed — no new pages, so no nav change

### How each page gets the project UUID

Two cases, and the split matters for review:

`RunTree.session_id` is declared `Optional[UUID] = None` (aliased to `project_id`) and nothing populates it client-side, and the JS `RunTree` has no such field at all. So a run traced by the app knows only its **project name** and has to resolve the UUID. Runs returned by the API do carry `session_id`, so those pages read it off the run instead.

Name-based resolution uses `create_project(..., upsert=True)` / `createProject({ upsert: true })` rather than `read_project()`. A project is created on trace ingestion, so resolving it right after the first traced call can 404; `upsert` removes that race without a retry loop. Verified idempotent, and it returns the same UUID `read_project()` would for a project that already exists.

| Page | How `session_id` is obtained |
|---|---|
| `attach-user-feedback.mdx` | `create_project(project_name=root_run.session_name, upsert=True).id`; TS resolves via `createProject()` |
| `log-traces-to-project.mdx` | One resolution per replica, each against its own client, because the two replicas write to different projects under different API keys |
| `observability-llm-tutorial.mdx` | Python reads `LANGSMITH_PROJECT` with a `"default"` fallback, matching `get_tracer_project()`; TS reads `runTree.project_name` |
| `optimize-classifier.mdx` | The page already pins the project to `"classifier"`, so it resolves that by name |
| `composite-evaluators-sdk.mdx` | `run.session_id`, straight off the run objects from `get_experiment_results()` |
| `run-evals-api-only.mdx` | Adds `session_id` to the `select` list on both `/runs/query` calls, then passes `run["session_id"]` in the REST body |

### All TypeScript examples use the object-form overload

Every `createFeedback` call in these docs (one on `attach-user-feedback.mdx`, two on `log-traces-to-project.mdx`, one on `observability-llm-tutorial.mdx`) now passes a single object:

```typescript
await client.createFeedback({ runId, sessionId, key: "feedback-key", score: 1.0 });
```

Two reasons, in order of importance:

1. `CreateFeedbackParams` declares `sessionId: string` as **required** when `runId` is present, so omitting it is a compile error. The three-argument overload declares `sessionId?: string`, where omitting it compiles cleanly and fails only at runtime. The type does the teaching.
2. The three-argument overload carries an `@deprecated` tag. `tsc` ignores that, but the TS language service does not, so a reader pasting a three-argument snippet into their editor sees `createFeedback` struck through. Shipping examples that an editor marks deprecated is not a good look.

Property order is consistent across all four: `runId`, `sessionId`, then `key`, then the payload.

### Release ordering

This is the one thing to get right before merging. The object-form overload does not exist in the published `langsmith` JS package yet, so until the SDK branch ships, these three TypeScript snippets do not compile against the released package. The Python snippets are fine either way, since `session_id` has always been an accepted keyword argument there. Merge after the SDK release, or accept a short window where the TS examples are ahead of the package.

### Other decisions worth a look

- **`start_time` is not in any example.** It is accepted and improves performance, but it is not required, and threading it through six pages added noise for no teaching value. It stays documented as a row in `feedback-data-fields.mdx`.
- **Pages deliberately untouched.** `presigned-feedback-tokens.mdx` and `agent-server-feedback.mdx` use the token endpoint, whose schema has no `session_id`; `annotation-queues-sdk.mdx` calls `create_feedback_config`, a different method; `audit-logs.mdx` only lists action names.

### Upstream follow-ups this PR cannot fix

1. The published `create_feedback` reference still describes `session_id` as optional ("Used to optimize feedback ingestion"), and its docstring example omits it. That example is what `attach-user-feedback.mdx` mirrors, so the two should stay in sync.
2. `langsmith-platform-openapi.json` still has `FeedbackCreateSchema.session_id` nullable, and describes the `feedback.<run_id>` multipart part as needing only `trace_id`. That file is auto-refreshed from the platform repo, so it needs an upstream change.

### Verification

Every example was **executed against a live LangSmith workspace**, not just read. Each run asserted that the feedback row came back with `session_id` equal to the resolved project UUID.

| Example | Result |
|---|---|
| `attach-user-feedback.mdx` (Python) | Both calls land; root-run and child-run feedback carry the resolved `session_id` |
| `attach-user-feedback.mdx` (TypeScript) | Ran against a locally built SDK; feedback landed with the resolved `sessionId` |
| `log-traces-to-project.mdx` | Two replicas resolved to two distinct project UUIDs; each feedback filed under its own project |
| `observability-llm-tutorial.mdx` | Ran end to end including the OpenAI call, with a non-default `LANGSMITH_PROJECT` |
| `optimize-classifier.mdx` | Ran end to end verbatim including the OpenAI call |
| `composite-evaluators-sdk.mdx` | Full flow: dataset, `evaluate()` with 3 judges, then composite feedback on both runs via `run.session_id` |
| `run-evals-api-only.mdx` | Both flows: single experiment (6 `correctness` rows) and pairwise (12 `ranked_preference` rows) |

The assumption flagged in the previous revision is now **closed**: `get_experiment_results()` does return runs with `session_id` populated, matching the experiment's project UUID. Verified on three pre-existing experiments plus the one this PR's own run created.

The TypeScript type-level claim is **verified** rather than asserted. All four documented snippets typecheck clean, and omitting `sessionId` fails to compile:

```
error TS2345: Property 'sessionId' is missing in type '{ runId: string; key: string; score: number; }'
  but required in type '{ runId: string; sessionId: string; projectId?: undefined; }'
```

Also verified: `session_id` comes back from `POST /runs/query`. That endpoint returned every field regardless of `select`, so adding `session_id` to the select list is explicitness rather than a strict requirement.

Race fix re-verified after the switch to `upsert`: all four affected examples were re-run against **brand-new** projects that had never been traced to, with no retry loop anywhere. The replica example is the sharpest case, since it resolves two projects that the run itself has just created; it previously needed a 24-attempt retry loop in the harness and now succeeds first try.

Rendering: `docs dev` (pipeline build plus `mint dev`) serves all six pages and `feedback-data-format` with HTTP 200. Spot-checked in a browser that the new `<Note>` renders as a callout, the Python tab shows the resolution line, and the imported field table shows the updated `session_id` row.

Vale: 0 errors, 0 warnings, 0 suggestions across all changed files.

### TODO: pre-existing bugs found while executing (not fixed here)

Both predate this PR, are unrelated to `session_id`, and reproduce on `main`. Need more investigation and fixing.

-  **The child-run example raises `IndexError`.** [`src/langsmith/attach-user-feedback.mdx:62`](https://github.com/langchain-ai/docs/blob/main/src/langsmith/attach-user-feedback.mdx#L62) — `foo_run_id = [run for run in child_runs if run.name == "foo"][0].id` indexes an empty list, because `child_runs` at [line 47](https://github.com/langchain-ai/docs/blob/main/src/langsmith/attach-user-feedback.mdx#L47) is always `[]`. `RunTree.child_runs` is only appended by `create_child()` (`langsmith-sdk` `python/langsmith/run_trees.py:658`), which `@traceable` no longer routes through. Confirmed by running the pre-change example straight off `main`. The same broken example is duplicated in the SDK docstring at `python/langsmith/client.py:8091` and `:8101`, so both should be fixed together. A working replacement is to fetch the child with `client.list_runs(project_name=..., trace_id=...)` after the trace is ingested.

- **`from langsmith import uuid7` sits outside its code fence.** [`src/langsmith/run-evals-api-only.mdx:88`](https://github.com/langchain-ai/docs/blob/main/src/langsmith/run-evals-api-only.mdx#L88) — the fence closes at line 87 and the import lands in body prose, so it renders as a stray paragraph. A reader copying the code blocks in order hits `NameError: uuid7`, since `uuid7()` is called later at lines 154 and 158. Either move it into the block ending at line 87 or into the block that uses it.

Written by Claude Opus 5 via Claude Code, under human direction and review.


src/langsmith/attach-user-feedback.mdx
```diff
@@ -21,6 +21,10 @@ The Python client will automatically background feedback creation if you pass `t
 This is essential for low-latency environments, where you want to make sure your application isn't blocked on feedback creation.
 </Tip>
 
+<Note>
```
QuentinBrosse: Do we need that, or is the example enough?

QuentinBrosse (approved):

src/langsmith/attach-user-feedback.mdx
emil-lc: Yeah, I agree it's not absolutely necessary. Will cut it out.

PR #5254, opened 2026-08-02
docs: list eucompliance.tools as an external callback integration

PatrickPi1312:
Lists `eucompliance-ai-act` as an external callback integration in `scripts/data/integration_external_docs.yaml`.

### Correction

I originally opened this as a hosted provider page plus an `all_providers` card. That was the wrong route — the triage bot pointed me at the [process introduced on 28 July 2026](https://docs.langchain.com/oss/python/contributing/publish-langchain#make-your-integration-discoverable), which is a YAML listing for integrations under the download threshold. I have taken the page and the card back out; this PR is now three lines in the YAML file and nothing else. Sorry for the noise.

### What it is

`eucompliance-ai-act` ([PyPI](https://pypi.org/project/eucompliance-ai-act/)) provides a `BaseCallbackHandler` that marks model output under **Article 50 of the EU AI Act**, which has applied since 2 August 2026. Article 50 places the transparency obligation on the deployer — whoever puts a model into a product — not on the model provider.

```python
from eucompliance_ai_act.integrations import langchain_callback

disclosure = langchain_callback(deployer="Your Company GmbH")
chain.invoke(input, config={"callbacks": [disclosure]})
```

Three properties were deliberate, because a compliance add-on that disrupts production gets removed after the first incident:

- **It does not block.** The disclosure runs on a background thread; the chain never waits for it.
- **It does not raise.** `raise_error` is `False` and exceptions are swallowed inside the handler.
- **Run content is not transmitted.** The SHA-256 is computed locally and only that hash is sent.


npentrel (approved):
Thanks!

PR #5262, opened 2026-08-03
Updates of MDA doc set before Public Beta

fjmorris:
Reconcile the Managed Deep Agents doc set to `managed-deepagents==0.5.0.dev84` for public beta. Supersedes #5035. All API facts verified against the installed SDK.

Provenance note: commits 1–3 were originally authored and verified against `0.5.0.dev80` (see their commit messages). The target has since moved to `0.5.0.dev84`; every commit 1–3 fact was re-verified against dev84 and still holds unchanged (identity headers, evals fixture schema, the `connectors.*` API surface, `connectors.github(events=[...])` routes/secrets, the LangSmith capability model, and the tool server integrations catalog). dev84 only adds capabilities (marketplace MCP connectors and additional identity connect providers), which are out of scope for commits 1–3. Commit 4 onward is authored against dev84.

**Commit 1 — terminology.** Rename `actor`→`user`, `tenant`→`organization`, `ingress`→`auth`, presets→`define_identity()`/`defineIdentity()` (scope-based), `LANGSMITH_TENANT_ID`→`LANGCHAIN_WORKSPACE_ID`, and the `X-MDA-Actor-Id`/`X-MDA-Tenant-Id` headers across identity, memory, how-it-works, tools, evals, and quickstart. `X-MDA-Ingress-Secret` and `MDA_INGRESS_SECRET` intentionally kept literal.

**Commit 2 — connectors API.** Move connector pages to the `connectors.*` namespace (`connectors.mcp`, `connectors.github`, `connectors.langsmith(capabilities=[...])`) with `export const connector`; add the GitHub tools/`installCLI` rule and a new tool server integrations catalog page.

**Commit 3 — GitHub webhooks fold.** Fold the GitHub channel into `connectors.github(events=[...])` on the connector page, delete the channels/github page, add a redirect, and repoint inbound links.

[...]

PR #5269, opened 2026-08-03
docs: clarify identity scope as three per-resource axes

fjmorris:
Docs-review pass on the Managed Deep Agents identity page. Stacked on #5262 (base `fjmorris/DOC-1403`), so this diff shows only the review change; it retargets to `main` automatically once #5262 merges.

Reworks the "Choose a scope" section so the three scopable resources and their distinct value sets are explicit (review feedback):

- Names the three resources up front — **threads, memory, credentials** — and states each accepts its own set of values.
- Splits the single merged value table into per-resource tables (Threads, Memory, Credentials), each listing only that resource's legal values with a "How the agent acts" column. Makes clear that `conversation` is threads-only and `custom` is credentials-only.
- Adds a "Set one value for all three" table showing how a single `scope` value expands per resource; keeps the product-shape combinations table.
- **Accuracy fix:** `scope: "organization"` resolves credentials to `none`, not `agent` (verified against dev84 `_SCOPE_AXES`). Flagging in case `agent` was the intended product behavior rather than a doc error.

Values and per-resource behavior sourced from `managed-deepagents==0.5.0.dev84` (`identity/__init__.py` type aliases and `identity/options.py`).

_Drafted with AI agent (dcode) assistance._

### DIFF
--- a/src/langsmith/managed-deep-agents-identity.mdx
+++ b/src/langsmith/managed-deep-agents-identity.mdx
@@ -81,29 +81,63 @@ flowchart LR
 
 ## Choose a scope
 
-`scope` is the isolation boundary for the deployment. One value covers the common product shapes; per-axis overrides handle the exceptions.
+`scope` sets the isolation boundary for the deployment across three resources:
 
-The scope values:
+- **Threads**: who can open or resume a conversation.
+- **Memory**: which durable memory slice a run may read and write.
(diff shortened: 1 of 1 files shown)


fjmorris:
Closed in favor of https://github.com/langchain-ai/docs/pull/5272

PR #5304, opened 2026-08-05
Add Fidacy action firewall to external tool integrations

lucaslubi:
Adds Fidacy to the external tool rows, Python and JS. Fidacy wraps a LangChain tool so the call is judged before it runs: a denied call never executes, and every allow and deny is Ed25519-signed so a third party can verify the decision offline. Works with any model driving the graph. Packages: langchain-fidacy on PyPI and @fidacy/langchain on npm, both published and tested from clean environments. I build and maintain it at https://github.com/fidacy/fidacy-open.

### DIFF
--- a/scripts/data/integration_external_docs.yaml
+++ b/scripts/data/integration_external_docs.yaml
@@ -538,6 +538,9 @@ python:
   - name: Toolstem
     pypi: langchain-toolstem
     docs_url: https://toolstem.com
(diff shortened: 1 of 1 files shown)


lucaslubi:
@lnhsingh ready for review. Two rows in integration_external_docs.yaml, Python and JS, both packages published and installable.

npentrel (approved):
Thank you!

PR #4649, opened 2026-06-29
Add SafePrompt integration provider page (JS)

ianreboot:
## Add SafePrompt integration provider page

[SafePrompt](https://safeprompt.dev) is a prompt-injection detection API. The [`@safeprompt.dev/langchain`](https://www.npmjs.com/package/@safeprompt.dev/langchain) package provides a callback handler that validates every prompt flowing through a chain or agent before it reaches the model, catching jailbreaks, system-prompt extraction, data-exfiltration attempts, and indirect injection from tool/retrieval output at the input boundary. It is provider-agnostic (it validates the input, so it works across OpenAI/Anthropic/Google/open-weights).

**This PR:**
- Adds `src/oss/javascript/integrations/providers/safeprompt.mdx` (provider page, following the existing provider template).
- Registers it in `all_providers.mdx`.

**Notes:**
- Package: `@safeprompt.dev/langchain`, MIT, `@langchain/core` as a peer dependency. Source: https://github.com/ianreboot/safeprompt/tree/main/packages/langchain
- It is an input-boundary guardrail (a callback handler) rather than a chat model / retriever / vector store, so there is no perfect `all_providers` section. I placed it under **Tools** as the closest fit. Happy to move it wherever you prefer.

Thanks for taking a look.

PR #5311, opened 2026-08-05
Point SafePrompt provider card at its LangChain integration page

ianreboot:
Follow-up to #4649: the SafePrompt card linked to the docs home; this points it at the dedicated LangChain (JS) integration page (https://docs.safeprompt.dev/langchain) so click-throughs land on the SafePromptCallbackHandler docs directly.


npentrel (approved):
Thank you!

PR #5307, opened 2026-08-05
docs(langsmith): document enabling Engine on self-hosted 

baskaryan:
## Overview
Adds the self-hosted enablement path for LangSmith Engine, which ships in
Helm chart 0.16.0 and is licensed separately, in the same way as Sandboxes.

- deploy-self-hosted-full-platform: new "Enable Engine" section covering
  components, prerequisites (Sandboxes, license entitlement, egress to the
  LangSmith Intelligence gateway, an externally reachable hostname, the
  Fernet key), Helm values, verification, the in-product steps that remain,
  and how to disable. Also adds Engine to the page title and intro.
- engine-self-hosted: replaces the "coming soon" note with the version and
  entitlement requirement, links to the enablement steps, and documents the
  fail-closed behavior when the inference gateway is unreachable.
- self-host-egress: documents the second required egress destination for
  Engine's model inference, and that offline installs cannot run Engine.
- self-host-using-an-existing-secret: adds engine_encryption_key and the
  rotation key.
- mirroring-images: adds the Engine image section and corrects the retired
  langsmith-clio reference in the signature-verification list.
- Insights now runs on the combined langsmith-insights-engine image, which
  the chart uses by default, so warn installs that still pin langsmith-clio
  to drop the pin before upgrading.

## Type of change

**Type:** New feature

## Checklist

- [ ] I have read the [contributing guidelines](README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)
- [ ] I have tested my changes locally using `docs dev`
- [ ] All code examples have been tested and work correctly
- [ ] I have used **root relative** paths for internal links
- [ ] I have updated navigation in `src/docs.json` if needed

(Internal team members only / optional): Create a preview deployment as necessary using the [Create Preview Branch workflow](https://github.com/langchain-ai/docs/actions/workflows/create-preview-branch.yml)

PR #5317, opened 2026-08-05
docs(langsmith): restore the Engine self-hosted architecture diagrams

baskaryan:
## Overview

Restores the three per-cloud architecture diagrams that #5310 removed from `engine-self-hosted.mdx`. The PNGs were never deleted, only the `<Frame>` references, and the diagrams are still live on the published page — so this looks like a regression rather than a decision.

Captions now say which environment each component runs in instead of asserting a connection type, and the GCP and Azure captions are marked as intended architecture to match the availability table above them.

## Type of change

**Type:** Restore removed content

## Related PRs

- Removed the diagrams: #5310
- Landed the surrounding Engine self-hosted docs: #5307

Written by Claude (via Claude Code) on @baskaryan's behalf.


fjmorris (approved):
merged as-is, will do a docs review next week

PR #5364, opened 2026-08-07
docs: fix stale pre-migration link in agent-auth.mdx

SpiliosDimakopoulos:
The interrupt() link in agent-auth.mdx still pointed to the old langchain-ai.github.io/langgraph how-tos URL from before docs were consolidated into docs.langchain.com. Every other reference to this same concept in the repo (cancel-run.mdx, streaming.mdx, core-capabilities.mdx, managed-deep-agents-middleware.mdx) already uses the internal /langsmith/add-human-in-the-loop path. This brings the last outlier in line.

### DIFF
--- a/src/oss/langchain/frontend/branching-chat.mdx
+++ b/src/oss/langchain/frontend/branching-chat.mdx
@@ -13,11 +13,10 @@ import { PatternEmbed } from "/snippets/pattern-embed.jsx"
 
 <PatternEmbed pattern="branching-chat" />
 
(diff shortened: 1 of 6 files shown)


src/oss/langchain/frontend/branching-chat.mdx
npentrel, line 1: why these changes? You're replacing them exactly with what's in the include as far as I can tell

npentrel (approved):

PR #5365, opened 2026-08-07
docs: fix broken/typo'd streaming links in evaluate-complex-agent.mdx

SpiliosDimakopoulos:
Two links pointed to a mistyped domain path (langchain-ai.github.io/langgra/... — missing the "ph", combined with a langsmith/observability-concepts/streaming path that doesn't exist on that domain at all). The remaining two links used the old pre-migration how-tos/streaming-subgraphs URL. Point all of them at the current /oss/langgraph/streaming page (#subgraph-outputs and #debug anchors), which already covers both topics referenced here.


npentrel (approved):
Thanks!

PR #5372, opened 2026-08-07
fix: create .bin/vale on the macOS Homebrew install path

Conrad-sudo:
## Overview

`make lint_prose` and `make install_vale` fail on macOS with Homebrew, because `scripts/install-vale.sh` installs Vale but never creates the binary the `Makefile` then tries to execute.

The Homebrew branch of the script exits before creating `$dest_path`:

```bash
if [ "$os_name" = "Darwin" ] && command -v brew >/dev/null 2>&1; then
  brew install vale
  exit 0          # $dest_path is never created
fi
```

The `Makefile` guards only the *install*, then runs `$(VALE_BIN)` unconditionally:

```make
@if [ ! -x "$(VALE_BIN)" ]; then bash scripts/install-vale.sh "$(VALE_BIN)" "$(VALE_VERSION)"; fi
...
        "$(VALE_BIN)" --glob='!**/node_modules/**' src/;
```

So the script "succeeds", and the next line is guaranteed to fail:

```
Linting prose with Vale...
/bin/sh: .bin/vale: No such file or directory
make: *** [lint_prose] Error 127
```

This fix symlinks the Homebrew binary into `$dest_path`:

```bash
ln -sf "$(brew --prefix)/bin/vale" "$dest_path"
```

`brew --prefix` resolves to `/opt/homebrew` on Apple Silicon and `/usr/local` on Intel, so both are covered.

### Secondary effect

It also makes the idempotency guard at the top of the script work. Because `$dest_path` never existed, this check could never be satisfied on macOS:

```bash
if [ -x "$dest_path" ]; then
  exit 0
fi
```

Every `make lint_prose` therefore re-ran `brew install vale`. With the symlink in place, repeat invocations short-circuit as intended.

### Scope

- **Affects:** local `make lint_prose` / `make install_vale` on macOS with Homebrew—which is the supported macOS path, since the script exits with *"Homebrew is required to install Vale on macOS"* when brew is absent.
- **Does not affect CI:** `lint-prose.yml` runs on `ubuntu-latest`, and both Linux branches (the Docker wrapper and the release-tarball download) already create `$dest_path`.
- **Does not affect** anyone who already has a working `.bin/vale`, or who overrides `VALE_BIN`.

## Type of change

## Related issues/PRs

- GitHub issue: none found—searched open and closed issues for `install-vale.sh` and `lint_prose`
- Feature PR: n/a

## Checklist

- [ ] ~~I have tested my changes locally using `docs dev`~~—n/a, this changes the Vale install script, not documentation content. Verified with `make install_vale` and `make lint_prose` instead (see below).
- [ ] ~~I have used **root relative** paths for internal links~~—n/a, no links changed
- [ ] ~~I have updated navigation in `src/docs.json` if needed~~—n/a, no pages added or moved

### Verification

On macOS 26.5.2 (Apple Silicon, Homebrew 6.0.15), from a clean state with `.bin/` removed:

| Step | Before | After |
|---|---|---|
| `make install_vale` | exits 0, `.bin/vale` absent | creates `.bin/vale` → `/opt/homebrew/bin/vale` |
| `.bin/vale --version` | `No such file or directory` | `vale version 3.17.1` |
| `make lint_prose` | `Error 127` | runs, lints 2263 files |
| second `make install_vale` | re-runs `brew install` | short-circuits, no-op |

I confirmed the script is byte-identical to `main` before the change, and reproduced the failure with a throwaway `dest_path` to rule out anything specific to my machine.

### Unrelated observation, deliberately not changed here

`make lint_prose` on a clean `main` currently reports 62 pre-existing `LangChain.DashesSpaces` errors in files this PR does not touch. CI does not catch them because the workflow lints only changed files. Flagging it rather than expanding this PR's scope.

There is also a Vale version spread that may or may not be intentional:

- `Makefile` pins `VALE_VERSION ?= v3.9.6`
- `lint-prose.yml` curl-installs `3.12.0` to `/usr/local/bin`, which appears to go unused: `VALE_BIN` is `.bin/vale`, and on a runner with Docker available the script takes the Docker branch first
- that Docker branch runs `jdkato/vale:latest`
- the Homebrew branch takes whatever brew ships (3.17.1 today); the `vale_version` argument is ignored on macOS both before and after this change

I left all of that alone, since pinning it is a maintainer call. Happy to follow up if you'd like it made consistent.


npentrel (approved):
Thanks!

npentrel:
I think we'll probably fix the vale version if that ever becomes an issue, could fix it now but it's not a priority for me. And the pre-existing dashes are annoying and fixable but they don't really hurt anyone either. I was considering making that rule a warning anyway. The installation issue is a great fix though for anyone onboarding!

Conrad-sudo:
Thanks for the quick turnaround, and for merging.

It makes sense that the version spread isn't worth chasing on its own. I only flagged it because I was already in there and figured it was better written down than not. If it ever does make local and CI disagree, easier to debug with a real symptom in hand.

PR #5404, opened 2026-08-08
docs: fix stale pre-migration reference links in use-graph-api.mdx

SpiliosDimakopoulos:
Two links pointed to the old langchain-ai.github.io/langgraph/reference/graphs.md API reference site. Replaced them with the repo's own @[] auto-link syntax (already used for state_schema in the same paragraph), which resolves to the current reference.langchain.com URLs. The second link also incorrectly attributed add_conditional_edges to MessageGraph, a class that no longer exists in the repo's link map (it belongs to StateGraph). Verified by running the repo's own handle_auto_links preprocessor against the new text to confirm it resolves to the correct URLs — output: 

A [StateGraph](https://reference.langchain.com/python/langgraph/graph/state/StateGraph) accepts...
...utilizing both standard edges and [conditional_edges](https://reference.langchain.com/python/langgraph/graph/state/StateGraph/add_conditional_edges)...

### DIFF
--- a/src/oss/langgraph/use-graph-api.mdx
+++ b/src/oss/langgraph/use-graph-api.mdx
@@ -753,7 +753,7 @@ Output of graph invocation: {"a":"set by node3"}
 :::python
 ### Use pydantic models for graph state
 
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thx!

PR #5405, opened 2026-08-08
docs: fix stale StateGraph reference links in langgraph error pages

SpiliosDimakopoulos:
Four links across three error reference pages (GRAPH_RECURSION_LIMIT.mdx,
INVALID_CONCURRENT_GRAPH_UPDATE.mdx, INVALID_GRAPH_NODE_RETURN_VALUE.mdx)
pointed to the old langchain-ai.github.io/langgraph/reference/graphs/ API
reference site. Replaced them with the repo's @[] auto-link syntax, which
resolves to the current reference.langchain.com URLs.

Notably, INVALID_GRAPH_NODE_RETURN_VALUE.mdx has separate :::python and
:::js blocks that both linked to the same python-only reference URL. Using
@[] instead of a hardcoded link lets the preprocessor's scope tracking pick
the correct language-specific reference automatically:
- :::python block -> reference.langchain.com/python/langgraph/graph/state/StateGraph
- :::js block     -> reference.langchain.com/javascript/langchain-langgraph/index/StateGraph

Verified by running the repo's own handle_auto_links.replace_autolinks()
against the edited files.

### DIFF
--- a/src/oss/langgraph/errors/GRAPH_RECURSION_LIMIT.mdx
+++ b/src/oss/langgraph/errors/GRAPH_RECURSION_LIMIT.mdx
@@ -2,7 +2,7 @@
 title: GRAPH_RECURSION_LIMIT
 ---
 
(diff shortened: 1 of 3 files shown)


npentrel (approved):
This will not get work for TS but we do know about this

PR #5407, opened 2026-08-08
docs: fix stale Command reference link in use-graph-api.mdx

SpiliosDimakopoulos:
The Command link in the "Combine control flow and state updates with Command" section still pointed to the old langchain-ai.github.io/langgraph/reference/types/ API reference site, while Command is already linked correctly via the repo's @[] auto-link syntax elsewhere in the same file. Replaced it with @[Command] for consistency; verified via the repo's own handle_auto_links.replace_autolinks() that it resolves to https://reference.langchain.com/python/langgraph/types/Command.


npentrel (approved):
Thank you!

PR #5408, opened 2026-08-08
docs: fix @[] auto-links resolving to wrong/broken URLs on JS-only pages

SpiliosDimakopoulos:
docs: fix @[] auto-links resolving to wrong/broken URLs on JS-only pages

These 3 pages are 100% JavaScript content with no :::python/:::js fences,
so the @[] auto-link preprocessor falls back to its 'global' default, which
resolves against the Python link map (see handle_auto_links._transform_link).
This caused two different failure modes:

- openai.mdx and openui.mdx: openAIModerationMiddleware and createDeepAgent
  only exist in the JS map, so the links failed entirely and rendered as
  literal, broken '@[...]' text on the page.
- aws.mdx: BedrockPromptCachingMiddleware happens to exist in BOTH the
  Python and JS maps (under an unrelated Python package), so the link
  silently resolved and looked fine, but pointed JS readers at the Python
  API reference instead of the JavaScript one.

Replaced all three with plain markdown links using the correct, verified
JavaScript reference.langchain.com URLs, bypassing the scope-defaulting
issue without restructuring the pages.

### DIFF
--- a/pipeline/preprocessors/link_map.py
+++ b/pipeline/preprocessors/link_map.py
@@ -500,6 +500,7 @@ LINK_MAPS: list[LinkMap] = [
             "WebsocketConnection": "langchain-mcp-adapters/sessions/WebsocketConnection",
             # JS-only references added here for cross-scope compatibility
             "createAgent": "https://reference.langchain.com/javascript/langchain/index/createAgent",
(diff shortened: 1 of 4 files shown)


npentrel:
The first two seem to work on main, I double checked. For aws.mdx you're right, but the proper fix is to add it to a link mapping - I've edited that here. Thanks for reporting!

npentrel (approved):

PR #5411, opened 2026-08-08
docs: fix @[] auto-links resolving to Python on JS-only pages

SpiliosDimakopoulos:
Same root cause as the earlier /oss/javascript/ fixes: pages with no
:::python/:::js fences fall back to the Python link map. Found via a
repo-wide scan for zero-fence pages whose code blocks are TypeScript/
JavaScript-dominant, cross-checked against SCOPE_LINK_MAPS.

- structured-output.mdx: @[AIMessage] on a 100% React/TypeScript frontend
  guide resolved to the Python AIMessage reference. Fixed to the JS URL.
- vitest-jest.mdx: @[evaluate()][Client.evaluate], on a guide exclusively
  about the JS Vitest/Jest test runners, has no JS entry in the link map
  at all and silently resolved to the Python Client.evaluate reference.
  Verified the correct current URL (reference.langchain.com/javascript/
  langsmith/evaluation/evaluate) and hardcoded it.


npentrel (approved):
Fixing in link map instead

PR #5476, opened 2026-08-12
fix: replace removed mda --configure-slack flag in docs

casparb:
`mda deploy . --configure-slack` no longer exists in the `mda` CLI so the Slack channel and CLI reference pages pointed at a command that fails. Replaced those references with the current `mda channel add slack .` command and listed the command in the CLI command overview.


christian-bromann (approved):
LGTM 👍

PR #5508, opened 2026-08-14
docs: document how list-form add_edge differs from separate edges into one node

serhiy-bzhezytskyy:
The page shows both fan-in spellings — separate edges into one node (with and without `defer=True`) and, once, `add_edge(["c", "d"], "a")` in a loop example — but never states that the two forms behave differently. The difference is what users trip on: a list-form edge whose start node is not selected by a conditional edge never runs its target and raises no error (langchain-ai/langgraph#3249, langchain-ai/langgraphjs#2666), and separate edges from branches of different lengths run the target more than once (langchain-ai/langgraphjs#1535).

This adds four sentences to the "Defer node execution" section: what the list form waits for, what happens when a listed node never runs, how that differs from separate edges, and that separate edges plus `defer=True` — the combination the section's example already uses — run the target once with whichever branches were selected. Behavior verified on `langgraph` 1.2.11.


serhiy-bzhezytskyy:
Ready for review @npentrel. This adds four sentences to the "Defer node execution" section: how list-form `add_edge` differs from separate edges into one node, and what each form does when a branch is skipped.

serhiy-bzhezytskyy:
@npentrel — following up on my note from the 14th. This documents how the list form of `add_edge` differs from separate edges into the same node, and names the combination that runs a fan-in once on whichever branches arrived. Measured on both runtimes; the table is in the PR body.

npentrel (approved):
Thanks!

PR #5519, opened 2026-08-14
docs(langsmith): clarify Include extended stats evaluator data

fjmorris:
Fixes DOC-1523

## Summary

Documents the **Include extended stats** evaluator setting so users can determine when to enable it and which data it makes available.

- Clarifies run-level support and the multi-turn evaluator limitation.
- Distinguishes `feedback_stats` from individual feedback records available to code evaluators.
- Adds supported LLM-as-a-judge variable-mapping guidance and a feedback-filtered chaining link.

## Validation

- `make lint_prose FILES='src/langsmith/evaluators.mdx src/langsmith/online-evaluations-code.mdx'`
- `make broken-links-with-anchors`

## Notes

This PR was prepared with AI agent assistance and reviewed against the LangSmith UI and backend implementation.


src/langsmith/evaluators.mdx
```diff
@@ -136,6 +136,21 @@ The change applies to traces scored after you save the evaluator. Existing score
 
 The **Extend trace retention** toggle described above applies to both trace-level and thread-level (multi-turn) online evaluators. For more information on multi-turn evaluators, see [Set up multi-turn online evaluators](/langsmith/online-evaluations-multi-turn).
 
+## Include extended stats
```
madams0013, line 140: I think we should give a full example of how to 'chain' evaluators. that's the main use case here for why people want this feature. ex.

People often want to reference an existing feedback score in a code evaluator.
If you set up 2 evaluators on a tracing project:

1. llm as a judge
2. code evaluator, toggle Advanced -> Include extended stats on and filter to runs with the specific feedback key from evaluator 1


the code evaluator will only run after the feedback has been added from evaluator 1 and you can reference this feedback score within the code evaluator!

lnhsingh: Adding a couple sentences about using this to chain evaluators but will leave a full code example for Florence to decide if we should add when she's back!

lnhsingh: https://linear.app/langchain/issue/DOC-1545/follow-up-on-extended-stats-example

PR #5523, opened 2026-08-15
docs(mda): omit redundant current-directory paths

casparb:
removes redundant `.` arguments from mda documentation examples


npentrel:
This change is already in main - apologies for the delay!

PR #5542, opened 2026-08-16
docs: fix broken JS cross-references for ToolErrorMiddleware (Python-only)

SpiliosDimakopoulos:
## Overview
Fixes 4 broken cross-references in src/oss/deepagents/fault-tolerance.mdx.
The page's own JS-scope note confirms `ToolErrorMiddleware` is not yet
available in the JavaScript SDK, but 4 places in the shared (non-fenced)
prose referenced it with the `@[ToolErrorMiddleware]` cross-reference
syntax, which has no entry in the JS scope of
pipeline/preprocessors/link_map.py. This produced 4 broken links when the
page builds in JS scope. Replaced them with plain inline code so they no
longer attempt to resolve against the JS link map. No content or behavior
change beyond fixing the broken links.

Verified locally by running the full pipeline build before and after:
4 "not found in scope 'js'" warnings for this file → 0.

## Type of change

## Related issues/PRs
Fixes #ΒΑΛΕ_ΕΔΩ_ΤΟΝ_ΑΡΙΘΜΟ_ΤΟΥ_ISSUE

### DIFF
--- a/src/oss/deepagents/fault-tolerance.mdx
+++ b/src/oss/deepagents/fault-tolerance.mdx
@@ -70,7 +70,7 @@ const agent = createAgent({
 :::python
   <Tab title="LLM-recoverable" icon="brain">
 
-  Use @[ToolErrorMiddleware] to catch tool exceptions and convert them into error `ToolMessage`s so the LLM can see what went wrong and try again:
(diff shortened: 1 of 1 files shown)


npentrel:
This works now due to https://github.com/langchain-ai/docs/pull/5656

PR #5545, opened 2026-08-16
docs(langgraph): document listing checkpoints across all threads

MrinalCom:
## Summary

`graph.get_state_history(config)` is documented as the way to see checkpoint history, but it always requires a `thread_id` and only returns history for one thread. There's no documented way to answer "what threads exist in this checkpoint database" -- which is a real need for building admin/debugging tooling on top of a checkpointer.

The reference `.list()`/`.alist()` implementations already support this: passing `config=None` (Python) or `{}` (JS, no `thread_id` in `configurable`) returns checkpoints across every thread. I found this by reading the `SqliteSaver` source while building a small checkpoint-inspection CLI, and confirmed it also holds for `InMemorySaver`/`MemorySaver` and `PostgresSaver` before writing this up -- see the verification snippets below.

This PR:
- Adds a "List every thread" example under **Get state history**, with working Python and JS snippets.
- Adds a one-line pointer to it from the `.list` bullet in the **Checkpointer interface** section (both Python and JS).

## Verification

Python (`InMemorySaver` and `SqliteSaver`, `langgraph` 0.6.11 / `langgraph-checkpoint` 2.1.2):

```python
from langgraph.checkpoint.memory import InMemorySaver
# ... compile graph, invoke for thread "a" and thread "b" ...
{t.config["configurable"]["thread_id"] for t in checkpointer.list(None)}
# {'a', 'b'}
```

JS (`MemorySaver` and `SqliteSaver`, `@langchain/langgraph` latest):

```typescript
// ... compile graph, invoke for thread "a" and thread "b" ...
for await (const t of app.checkpointer.list({})) {
  threads.add(t.config.configurable.thread_id);
}
// Set(2) { 'a', 'b' }
```

## Test plan

- [x] Ran the exact Python and TypeScript snippets above against real compiled graphs with `InMemorySaver`/`MemorySaver` and `SqliteSaver`, confirmed both threads are returned.
- [x] Checked heading/anchor doesn't collide with an existing heading in the file.
- [x] `codespell` clean on the changed file.
- [x] Matched existing `:::python`/`:::js` block conventions and `@[...]` symbol-reference usage already in this file.

### DIFF
--- a/src/oss/langgraph/checkpointers.mdx
+++ b/src/oss/langgraph/checkpointers.mdx
@@ -550,6 +550,31 @@ const interrupted = history.find(
 ```
 :::
 
+#### List every thread
+
+`graph.get_state_history(config)` always requires a `thread_id` in `config` -- it only returns history for one thread. To see every thread a checkpointer holds state for (for example, to build an admin or debugging view), call the checkpointer's `.list` directly and omit `thread_id`:
+
+:::python
+```python
+# No thread_id in config -- returns checkpoints across every thread
+for checkpoint_tuple in checkpointer.list(None):
+    print(checkpoint_tuple.config["configurable"]["thread_id"])
+```
+
+This works with the built-in @[`InMemorySaver`], @[`SqliteSaver`], and @[`PostgresSaver`] checkpointers. `.list` still returns one row per checkpoint, not per thread, so group by `thread_id` yourself to build a thread list.
+:::
+
+:::js
+```typescript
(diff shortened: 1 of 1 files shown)


MrinalCom:
@npentrel @lnhsingh The PR is ready for review. All checks are passing, and I’ve verified the Python and JS examples against the relevant checkpointers. Would appreciate a review when convenient. Thanks!

PR #5565, opened 2026-08-18
docs: list langchain-yantrikdb in integration_external_docs.yaml

spranab:
## Overview

Adds `langchain-yantrikdb` to `scripts/data/integration_external_docs.yaml` so it appears in the vector store and chat message history download tables.

`langchain-yantrikdb` is a standalone package (PyPI: [`langchain-yantrikdb`](https://pypi.org/project/langchain-yantrikdb/), repo: [yantrikos/langchain-yantrikdb](https://github.com/yantrikos/langchain-yantrikdb), MIT) that implements two standard interfaces on top of YantrikDB, an embedded memory engine:

- `YantrikDBVectorStore` — `VectorStore`. Stored texts get an importance and a half-life; ranking blends similarity with decay and recency, and retrieval can report why each hit surfaced.
- `YantrikDBChatMessageHistory` — `BaseChatMessageHistory`, namespace-scoped, for `RunnableWithMessageHistory`.

Two rows, both pointing at the package's own GitHub README — no docs page added here.

## Type of change

**Type:** Other (integration listing metadata only)

This follows [Publish an integration → List in the download table](https://docs.langchain.com/oss/contributing/publish-langchain#list-in-the-download-table-default). The package is at ~122 monthly PyPI downloads, well under the 50,000 threshold for a hosted guide, so this is a YAML listing rather than a new docs page.

The capability flags are set from the actual implementation:

| Flag | Value | Why |
| --- | --- | --- |
| `delete_by_id` | true | `delete(ids=[...])` tombstones by rid |
| `search_by_vector` | true | `similarity_search_by_vector` |
| `search_with_score` | true | `similarity_search_with_score` |
| `async_api` | false | sync only today |
| `passes_standard_tests` | false | the package has its own real-engine test suite, but does not run `langchain-tests` |
| `multi_tenancy` | true | stores are namespace-scoped within one database file |
| `ids_in_add_documents` | false | the engine assigns UUIDv7 rids; caller-supplied ids raise `NotImplementedError` |

`filtering` is deliberately omitted rather than guessed: search accepts engine-level filters (`domain`, `memory_type`, `source`, `namespace`) but not arbitrary metadata-dict filters, so neither `true` nor `false` would be accurate.

## Related issues/PRs
- GitHub issue:
- Feature PR:

## Checklist
- [ ] I have tested my changes locally using `docs dev` — YAML-only change; validated that the file still parses and that both entries load
- [x] All code examples have been tested and work correctly (no code examples in this change)
- [x] I have used **root relative** paths for internal links (no internal links added)
- [x] I have updated navigation in `src/docs.json` if needed (not needed — no new page)

## Additional notes

Happy to adjust the display names, the `docs_url` targets, or drop the chat message history row if one row per package is preferred.


spranab:
Update: `langchain-yantrikdb` now passes the `langchain-tests` standard `VectorStoreIntegrationTests` suite (25/25, run against a real embedded engine rather than mocks), so I've flipped `passes_standard_tests` to `true` in the entry.

That required implementing caller-supplied ids, which the package previously rejected — they now work as upsert keys, so `ids_in_add_documents` is `true` as well. The remaining declared flags are unchanged and accurate: `async_api` stays `false` (no native async; the engine is synchronous).

Tests: https://github.com/yantrikos/langchain-yantrikdb/blob/main/tests/integration_tests/test_standard.py — no rush on review, just keeping the entry honest.

npentrel (approved):

PR #5577, opened 2026-08-18
docs: add XNS provider page and packages.yml entry

FaustianAGI:
Adds a provider page and registry entry for `langchain-xns`, an integration for [XNS](https://xns.tech) S3-compatible object storage.

- PyPI: https://pypi.org/project/langchain-xns/
- Source: https://github.com/xns-cloud/langchain-xns

## What the package provides

- `XNSLoader` — streams text objects from a bucket as `Document` objects, with server-side prefix filtering and client-side suffix/glob/size filtering so excluded keys are never downloaded.
- `XNSBlobLoader` — yields `Blob` objects for parser pipelines, with `mimetype` from the object's `Content-Type`.
- `XNSByteStore` — a full `BaseStore[str, bytes]` implementation for `CacheBackedEmbeddings` and `MultiVectorRetriever`.

Rather than re-wrapping the generic S3 loader, it adds credential resolution from `~/.xns/credentials`, so a loader typically needs only a bucket name.

## Files

- `packages.yml` — one entry appended. `downloads` and `downloads_updated_at` are omitted so the weekly job fills them; `highlight` is not set.
- `src/oss/python/integrations/providers/xns.mdx` — provider page, structured to match `box.mdx` in the same directory.


FaustianAGI:
@mdrxy ready for review when you have a moment. All checks are green.

The package is published (https://pypi.org/project/langchain-xns/) and the source repo is public (https://github.com/xns-cloud/langchain-xns), so both references in the `packages.yml` entry resolve.

npentrel (approved):
Thansk!

PR #3733, opened 2026-04-25
docs(deepagents): OpenAI Responses API and data retention on Models page

AdemBoukhris457:
Add "OpenAI and the Responses API" to models.mdx: default Responses path for openai: strings, Chat Completions via use_responses_api=False, and store=False + include=["reasoning.encrypted_content"] with a pre-built model instance. Aligns with create_deep_agent docstring; JS points to OpenAI chat integration.

## Overview

Documents Deep Agents behavior for OpenAI models on the [Models](/oss/deepagents/models) page: Responses API as the default for `openai:...` strings, how to switch to Chat Completions with a pre-initialized model, and how to opt out of data retention on Responses while keeping encrypted reasoning output (`store=False`, `include=["reasoning.encrypted_content"]`). Clarifies that OpenAI-specific kwargs require passing a configured model instance, not only a provider string. TypeScript readers are directed to the OpenAI chat integration for equivalent options.

## Type of change

**Type:** Update existing documentation

## Related issues/PRs

- GitHub issue: closes #3732
- Feature PR:

- Linear issue:
- Slack thread:

## Checklist

- [x] I have read the [contributing guidelines](README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)
- [x] I have tested my changes locally using `docs dev`
- [x] All code examples have been tested and work correctly
- [x] I have used **root relative** paths for internal links
- [x] I have updated navigation in `src/docs.json` if needed

PR #5587, opened 2026-08-19
docs(deepagents): document OpenAI Responses API default on models page

### LINKED ISSUE #3730: [deepagents]: Document OpenAI Responses API default on models page
### Type of issue

issue / bug

### Language

Python

## Summary

When you pass an OpenAI model **string** such as `openai:gpt-5.4` to `create_deep_agent`, Deep Agents resolves it via `resolve_model` → `init_chat_model` with harness profile kwargs from `libs/deepagents/deepagents/profiles/_openai.py`, which registers `use_responses_api=True` by default.

That behavior is documented in Python docstrings (`resolve_model`, `create_deep_agent` “OpenAI Models and Data Retention” note) but **not** on the public Deep Agents **Models** page (`src/oss/deepagents/models.mdx`). Readers who configure models only from the docs may assume Chat Completions–style defaults or miss data-retention implications.

## Library references

- `libs/deepagents/deepagents/profiles/_openai.py` — `_HarnessProfile(init_kwargs={"use_responses_api": True})` for provider `openai`.
- `libs/deepagents/deepagents/_models.py` — `resolve_model` merges profile `init_kwargs` into `init_chat_model`; docstring states OpenAI string models default to the Responses API.
- `libs/deepagents/deepagents/graph.py` — `create_deep_agent` docstring note on Responses vs chat completions (`use_responses_api=False`) and optional `store` / `include` for retention.

## Expected documentation

On `src/oss/deepagents/models.mdx` (and optionally a one-line cross-link elsewhere):

- State explicitly that **`openai:` string specs` default to **`use_responses_api=True`** when used with Deep Agents.
- Explain how to opt into **Chat Completions** instead: `init_chat_model("openai:...", use_responses_api=False)` and pass the **initialized model instance** to `create_deep_agent` (so harness defaults for that flag are not applied the same way as for a bare string).
- Summarize or link the **data retention** guidance already in the `create_deep_agent` docstring for Responses API users.

## Why it matters

Users should not have to read source or API docstrings to discover a default that changes which OpenAI API surface is used and affects retention-related knobs.

SpiliosDimakopoulos:
Fixes #3730

Documents that Deep Agents applies `use_responses_api=True` by default
for all `openai:` model strings via the built-in OpenAI provider profile.
Adds guidance on how to opt into Chat Completions and how to disable
Responses API data retention.

### DIFF
--- a/src/oss/deepagents/models.mdx
+++ b/src/oss/deepagents/models.mdx
@@ -100,6 +100,26 @@ See [Profiles](/oss/deepagents/profiles) for the full field list, merge semantic
     For shaping how the *agent* behaves once the model is built, use a [harness profile](/oss/deepagents/profiles#harness-profiles).
 </Tip>
 
+### OpenAI Responses API default
+
+:::python
+When you pass an OpenAI model as a `provider:model` string (for example `openai:gpt-5.4`), Deep Agents applies a built-in provider profile that sets `use_responses_api=True`. This means the agent calls the [OpenAI Responses API](https://platform.openai.com/docs/api-reference/responses) instead of Chat Completions, and it applies whenever a string is passed — it does not depend on your own profile registrations.
+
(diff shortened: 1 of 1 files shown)


npentrel:
I've merged https://github.com/langchain-ai/docs/pull/3733 which has similar changes

PR #5606, opened 2026-08-20
docs: fix client.runs.cancel Python scope entry pointing to JS reference URL

SpiliosDimakopoulos:
## Summary

In the Python scope block of `pipeline/preprocessors/link_map.py`
(host: `reference.langchain.com/python/`), every `client.runs.*` and
`client.threads.*` sibling uses a relative path under the Python SDK:


npentrel (approved):
edited slightly

PR #5139, opened 2026-07-28
Add PolarDB-X vector store integration docs

Jiniqi-ty:
## Overview
Adds documentation for the `langchain-polardbx` vector store integration, including setup, instantiation, basic CRUD, and PolarDB-X specific features (metadata filtering, score threshold, search mode control, dynamic index management, index monitoring, async API).

## Type of change

**Type:** New documentation page

## Related issues/PRs
- GitHub issue: N/A
- Feature PR: N/A

## Checklist
- [x] I have read the [contributing guidelines](README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)
- [ ] I have tested my changes locally using `docs dev`
- [x] All code examples have been tested and work correctly
- [x] I have used **root relative** paths for internal links
- [x] I have updated navigation in `src/docs.json` if needed

## Additional notes
- PyPI package: https://pypi.org/project/langchain-polardbx/
- Source code: https://github.com/polardb/langchain-polardbx
- The package is published to PyPI (v0.1.2) and passes LangChain standard tests.

PR #5620, opened 2026-08-21
docs(polardbx): add feature flags to PolarDBX vector store entry

Jiniqi-ty:
The PolarDBXVectorStore entry in integration_external_docs.yaml was
added in #5139 without capability flags, so the vectorstore downloads
table shows empty cells for all feature columns.

This PR declares the supported capabilities, verified against the
langchain-polardbx 0.4.1 source and its test suite (which subclasses
VectorStoreIntegrationTests from langchain-tests):

- delete_by_id, filtering, search_by_vector, search_with_score,
  async_api, passes_standard_tests, ids_in_add_documents: true
- multi_tenancy: false (not supported)

The downloads table will pick these up on the next scheduled
refresh_integration_downloads.py run.

Follow-up to #5139.

### DIFF
--- a/scripts/data/integration_external_docs.yaml
+++ b/scripts/data/integration_external_docs.yaml
@@ -1298,6 +1298,14 @@ python:
   - name: PolarDBXVectorStore
     pypi: langchain-polardbx
     docs_url: https://github.com/polardb/langchain-polardbx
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thanks!

PR #5630, opened 2026-08-21
fix: prefix mda init with npx in the TypeScript quickstart

casparb:
## Description

The TypeScript quickstart told readers to run bare `mda init` after `npm install managed-deepagents`, but a local install does not put `mda` on `PATH`. Prefix the command with `npx` so it resolves the locally installed binary.

## Test Plan

- [ ] `npm install managed-deepagents && npx mda init research-assistant`

Made by [Open SWE](https://openswe.vercel.app/agents/77ca81f3-f11f-5d57-8fda-f6af12c57052)


npentrel:
something like this has been changed in the meantime, already - sorry for the delay!

PR #5631, opened 2026-08-22
docs: list langchain-lyrenth in the package table

lyrenth:
Adds a package-table row for `langchain-lyrenth`, published on PyPI: https://pypi.org/project/langchain-lyrenth/

Lyrenth is an independent web index. Reading a URL through it returns cleaned Markdown instead of raw HTML, typically 80 to 90 percent fewer input tokens for the same page, and each result carries its canonical source URL so an answer built on it can be attributed. Pages are served from a standing index, so many agents reading the same URL does not become many requests to that website.

The package provides:

- `LyrenthReadTool`, a tool an agent can call to read a page (sync and async), and
- `LyrenthLoader`, a document loader for pipelines with known URLs.

Repository: https://github.com/lyrenth/langchain-lyrenth (MIT, Python 3.9+, tests included)

Per the publishing guide, the package lives in its own repository and this is the download-table listing rather than a hosted guide, since it does not meet the 50,000 monthly download threshold. A free API key is available at https://lyrenth.com/signup so the components can be tried without a card.


lyrenth:
Ready for review. @lnhsingh, tagging you as the welcome bot suggested, since I cannot request reviewers on a fork PR.

This is the download-table listing from the publishing guide (two lines in `packages.yml`), not a docs page. All checks are green.

npentrel (approved):

PR #4140, opened 2026-05-22
chore: remove JS community integrations

npentrel:
As we are sunsetting langchain community, we are removing associated integrations pages. Many important integrations have already moved into dedicated packages featured on our integrations page: https://docs.langchain.com/oss/python/integrations/providers/overview. Companies interested in maintaining an official integration can follow our contributor docs at https://docs.langchain.com/oss/python/contributing/integrations-langchain, create their own standalone package, and contact us to be featured on our integrations page: https://docs.langchain.com/oss/python/integrations/providers/overview.

PR #5636, opened 2026-08-22
oss(js): rewrite PDFLoader integration guide, fix WebPDFLoader docs

### LINKED ISSUE #5627: Add PDF Loader for Javascript Docs
### Type of issue

request for content

### Language

JavaScript

### Description

I was looking for the PDFLoader documentation for LangChain.js while working on a TypeScript project, but I couldn't find a dedicated PDFLoader guide in the current documentation.

PDFLoader is still available, but without a dedicated guide it's difficult to know the recommended way to use it, especially for someone learning LangChain through the official docs.

It would be helpful to have a PDFLoader page covering the basic setup, installation, import path, a simple TypeScript example, the `Document` objects returned by `load()`, available options such as page splitting, and any Node.js requirements or limitations.

Having this documentation would make it much easier for JavaScript/TypeScript developers to get started with PDF loading without having to rely on older tutorials or search through the source code.

Payal2000:
## Overview

Issue #5627 reports that there is no `PDFLoader` guide for LangChain.js. A page does exist at `src/oss/javascript/integrations/document_loaders/file_loaders/pdf.mdx`, but its content had drifted far enough from the current `@langchain/community` implementation that following it does not work, which is a reasonable explanation for why the reporter concluded it was missing.

I installed `@langchain/community` 1.1.29 with `@langchain/core` 1.2.9 and `pdf-parse` 2.4.5 (plus a separate project on `pdf-parse` 1.x) and ran every example on the page. Each of the following was reproduced, not inferred from reading source:

* **`parsedItemSeparator` does not do what the page says.** The default is `""`, not a space, and the option is ignored entirely on the `pdf-parse` v2 code path. The page's featured "Eliminating extra spaces" fix was therefore a no-op. `parseWithV2()` never reads the field, which looks like an upstream bug worth filing separately against `langchainjs-community`.
* **The custom build example cannot run.** `pdfjs-dist` v6 ships `legacy/build/pdf.mjs`; `legacy/build/pdf.js` no longer exists. The `pdfjs` option also expects an object tagged with `isV2`.
* **The documented install command fails.** `npm install @langchain/community @langchain/core pdf-parse` errors with `ERESOLVE` on a clean project, because npm resolves `@langchain/core` `0.3.x` through a transitive optional peer dependency. Pinning `@langchain/core` to `^1` fixes it.
* **`pdf.version` is misleading.** It reports `unknown` under `pdf-parse` v2. `info.PDFFormatVersion` is the reliable source for the PDF specification version.
* Reference links pointed at the retired `api.js.langchain.com` host.

The rewrite also covers what the issue explicitly asked for and the page did not have: the `Document` metadata shape returned by `load()`, an options table, `Blob` input, and the Node.js only limitation with a pointer to `WebPDFLoader`.

Housekeeping in the same pass: removed roughly 200 lines of raw Nike 10-K output dumps (one of which leaked a contributor's absolute home directory path), an empty `### Credentials` heading, and a dangling "(See note below)" marker in the integration details table that had no matching note.

The second commit applies the same corrections to the `WebPDFLoader` page, which shares a parser implementation and carried the same broken `pdfjs-dist` path, the same `parsedItemSeparator` claim, and the same dead reference links. That change is deliberately limited to what is demonstrably wrong rather than a full rewrite.

### Areas that need careful review

* **The `pdf-parse` v1 vs v2 comparison table** in "Choose a `pdf-parse` version" is the most opinionated part of this PR. It documents observed behavior on both majors, including the `parsedItemSeparator` discrepancy. If maintainers would rather fix that upstream and leave it undocumented here, that section should come out.
* **`docs.json` navigation is untouched.** No individual document loader integration page is listed in navigation today, so adding only this one would break the existing pattern. See the note below about discoverability.
* I did not rewrite the `WebPDFLoader` page's example output blocks, so they still show `pdf-parse` v1 metadata (`version: '1.10.100'`). That is still correct for v1 users, but it is inconsistent with the sibling page now.

### Separate problem, not fixed here

Most JavaScript document loader integration pages return 404 on the live site even though they are present in `src/` and land correctly in `build/`:

| URL under `/oss/javascript/integrations/document_loaders/` | Live |
| :--- | :---: |
| `file_loaders/pdf` | 404 |
| `file_loaders/csv` | 404 |
| `file_loaders/docx` | 404 |
| `web_loaders/pdf` | 404 |
| `file_loaders/text` | 200 |
| `file_loaders/json` | 200 |

I confirmed `make build` writes `pdf.mdx` to `build/oss/javascript/integrations/document_loaders/file_loaders/`, and the page renders correctly under `make dev`, so this looks like a publishing gap rather than a content one. It is very likely the direct cause of the issue as filed, and this PR alone will not resolve it. Happy to open a separate issue if that is useful.

## Type of change

## Related issues/PRs

- GitHub issue: closes #5627
- Feature PR:

## Checklist

- [x] I have updated navigation in `src/docs.json` if needed (no change needed, see above)

## Additional notes

Verification run locally:

* `make build` succeeds, with no autolink or link map warnings for either changed file.
* `make dev` serves both pages with the expected titles, matching a known good sibling page.
* `make test` passes (103 tests).
* `markdownlint` is clean on both files, including two pre-existing table style violations that are fixed here.
* In-page anchors were checked against the generated heading slugs in `build/`.

**AI disclosure:** an AI coding agent (Claude Code) was used for this contribution. It performed the codebase and upstream source investigation, executed the verification described above, and drafted the page text and this description. Every factual claim in the rewritten page was checked against output from a real run rather than accepted from the model, and the findings, scope decisions, and final wording were reviewed by me before opening this PR.

### DIFF
--- a/src/oss/javascript/integrations/document_loaders/file_loaders/pdf.mdx
+++ b/src/oss/javascript/integrations/document_loaders/file_loaders/pdf.mdx
@@ -4,173 +4,77 @@ description: "Integrate with the PDFLoader document loader using LangChain JavaS
 ---
 
 <Tip>
-**Compatibility**: Only available on Node.js.
+**Compatibility**: Only available on Node.js. To load PDFs in the browser or another environment without filesystem access, use [`WebPDFLoader`](/oss/integrations/document_loaders/web_loaders/pdf).
 </Tip>
 
-This notebook provides a quick overview for getting started with `PDFLoader` [document loaders](/oss/integrations/document_loaders). For detailed documentation of all `PDFLoader` features and configurations head to the [API reference](https://api.js.langchain.com/classes/langchain_community_document_loaders_fs_pdf.PDFLoader.html).
+`PDFLoader` reads a PDF from the local filesystem and returns one @[Document] per page. For detailed documentation of all `PDFLoader` features and configurations, head to the [API reference](https://reference.langchain.com/javascript/classes/_langchain_community.document_loaders_fs_pdf.PDFLoader.html).
 
 ## Overview
 
 ### Integration details
 
 | Class | Package | Compatibility | Local | PY support |
-| :--- | :--- | :---: | :---: |  :---: |
-| [PDFLoader](https://api.js.langchain.com/classes/langchain_community_document_loaders_fs_pdf.PDFLoader.html) | [@langchain/community](https://api.js.langchain.com/modules/langchain_community_document_loaders_fs_pdf.html) | Node-only | ✅ | 🟠 (See note below) |
+| :--- | :--- | :---: | :---: | :---: |
+| [PDFLoader](https://reference.langchain.com/javascript/classes/_langchain_community.document_loaders_fs_pdf.PDFLoader.html) | [@langchain/community](https://reference.langchain.com/javascript/modules/_langchain_community.html) | Node-only | ✅ | ❌ |
+
+Python has a separate set of PDF loaders, so there is no Python equivalent of this class.
 
 ## Setup
 
-To access `PDFLoader` document loader you'll need to install the `@langchain/community` integration, along with the `pdf-parse` package.
-
-### Credentials
-
 ### Installation
 
-The LangChain PDFLoader integration lives in the `@langchain/community` package:
+The LangChain `PDFLoader` integration lives in the `@langchain/community` package. Parsing is delegated to [`pdf-parse`](https://www.npmjs.com/package/pdf-parse), which is an optional peer dependency you install yourself:
 
 <CodeGroup>
 ```bash npm
-npm install @langchain/community @langchain/core pdf-parse
+npm install @langchain/community@^1 @langchain/core@^1 pdf-parse@^2
 ```
 ```bash yarn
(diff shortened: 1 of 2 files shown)


Payal2000:
Closing this. I opened it against a stale fork and got two things wrong.

**1. These pages were removed on purpose.** #4140 deleted both `file_loaders/pdf.mdx` and `web_loaders/pdf.mdx` while sunsetting `@langchain/community`, and the package is now deprecated on npm. This PR would have reverted that decision and told readers to install a deprecated package. My claim in the original description that the live 404s were a publishing gap was simply wrong; the pages are gone by design.

**2. I opened it before requesting assignment.** The contributing guide requires that a maintainer approve the approach and assign the linked issue first. That is on me.

The verification work still stands, so I have moved the useful part to a proposal on #5627: a short PDFs section on the JavaScript document loaders index that parses with `pdf-parse` directly and wraps each page in a `Document`, with no dependency on the deprecated package. Happy to open a fresh PR if a maintainer thinks that is worth having.

Sorry for the review noise.

**AI disclosure:** an AI coding agent was used to investigate and draft this contribution. All findings were verified by running the code locally.

PR #5666, opened 2026-08-25
docs(langgraph): document retry boundaries and idempotency for node side effects

### LINKED ISSUE #8702: Document retry and idempotency expectations for node tasks
### Summary

Document what LangGraph retries can repeat and where application code must provide idempotency for side effects.

### Current implementation facts

LangGraph supports durable execution and retryable graph work, while nodes can contain arbitrary application code.

### Problem

A retry can re-enter a node after a transient failure. Users may put payments, email, or mutations directly in retryable code and perform them twice.

### Scope

- Document retry boundaries.
- Show a side effect protected by an idempotency key.
- Explain checkpoint state versus external state.

### Non-goals

- Do not add an idempotency service.
- Do not change retry defaults.

### Acceptance criteria

- The example retries without duplicating the operation.
- It names the application-owned idempotency boundary.
- Docs build successfully.

I searched LangGraph issues and PRs for retry, idempotency, durable execution, and side-effect documentation before filing this proposal.

PiedPiper911:
## Overview

Adds a new section **"Retry boundaries and idempotency"** to the LangGraph [Fault tolerance](https://docs.langchain.com/oss/python/langgraph/fault-tolerance) guide, documenting what retries can repeat and where application code must provide idempotency for side effects.

The section covers:

- **What a retry can repeat** — retries give at-least-once semantics for node execution; external side effects (payments, emails, queues) can run more than once.
- **Protect side effects with an idempotency key** — Python and JS examples showing a stable idempotency key passed to an external service so retries deduplicate.
- **Checkpoint state vs external state** — a comparison table clarifying what the checkpointer guarantees vs what application code must guard.

Scope follows the issue: no new retry semantics, no idempotency service, no changes to defaults.

## Type of change

## Related issues/PRs

- GitHub issue: fixes langchain-ai/langgraph#8702

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/langchain-ai/docs/blob/main/README.md), including the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy)

## Additional notes

The existing `fault-tolerance.mdx` covers retry configuration, timeouts, and error handling but does not discuss retry semantics for side effects or idempotency — this PR fills that gap. No navigation changes needed (new section within an existing page).

### DIFF
--- a/src/oss/langgraph/fault-tolerance.mdx
+++ b/src/oss/langgraph/fault-tolerance.mdx
@@ -5,9 +5,9 @@ description: Configure per-node timeouts, retries, and error handlers in LangGra
 
 When a node fails—from a slow external API, a transient network error, or an unhandled exception—LangGraph gives you three composable mechanisms to respond:
 
-- [**Retries**](#retries) — automatically re-run failed attempts based on exception type and backoff settings
-- [**Timeouts**](#timeouts) — cap how long a single attempt may run
-- [**Error handling**](#error-handling) — run a recovery function after all retries are exhausted
+- [**Retries**](#retries)—automatically re-run failed attempts based on exception type and backoff settings
+- [**Timeouts**](#timeouts)—cap how long a single attempt may run
+- [**Error handling**](#error-handling)—run a recovery function after all retries are exhausted
 
 :::python
 Use [**`set_node_defaults`**](#graph-defaults) to configure these mechanisms once for all nodes instead of repeating them on every `add_node` call.
--- a/src/oss/langgraph/fault-tolerance.mdx
+++ b/src/oss/langgraph/fault-tolerance.mdx
@@ -269,6 +269,90 @@ const graph = new StateGraph(State)
 `executionInfo` is available even without a retry policy—`nodeAttempt` defaults to `1`.
 :::
 
+## Retry boundaries and idempotency
+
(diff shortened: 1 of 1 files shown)


stringsofthemind-oss:
Small correctness issue in the proposed idempotency example: the generated UUID only becomes durable if the node successfully returns its state update.

If provider.charge() succeeds and the process dies before the node returns/checkpoints payment_key, a retry starts from the previous state, generates a new UUID, and sends a different idempotency key. That can duplicate the very side effect this section is intended to protect.

I think the example should either derive the idempotency key deterministically from already-durable business identity (for example operation type + immutable payment/order ID), or persist a generated key in a separate checkpointed step before performing the external side effect.

It may also be worth distinguishing provider-side deduplication from outcome reconciliation: if the first attempt’s acknowledgement is lost, absence of a local receipt alone does not prove the external side effect did not occur, so a retry should not assume a second write is safe without provider-side truth.

stringsofthemind-oss:
The ordering barrier proposed here looks like the right framework fix for this specific persistence race. One separate application-level boundary still remains for external mutations: the provider commit and the LangGraph checkpoint cannot generally be one atomic transaction.

We reproduced that boundary with real StateGraph + SqliteSaver + hard process death: a fresh-process resume can re-enter the node after the external effect already committed. Stable operation identity prevents blind duplication when the provider supports idempotency; if the first acknowledgement is lost, recovery also needs provider reconciliation rather than treating missing checkpoint state as proof the effect did not happen.

Repro: https://github.com/stringsofthemind-oss/once/tree/main/examples/langgraph-retry-lab

PR #5669, opened 2026-08-25
docs(deepagents): fix Python dynamic skill list paths

wangjiawei-vegetable:
## Overview

Fix the Python "Dynamic skill lists" example so it matches the source layout expected by the current Python `SkillsMiddleware`.

The previous example passed individual leaf Skill directories through `skills=`. Python currently treats each configured path as a container and only discovers Skills in its immediate subdirectories, so the example could silently load no Skills.

The Python example now uses container directories, and the shared Skills page documents the language-specific path behavior. The JavaScript example remains unchanged because JavaScript already supports direct leaf Skill paths.

## Background

The Skills documentation shows how to select a different set of Skills based on runtime context such as a user's role.

The previous Python example configured each role with individual Skill directories:

```python
SKILLS_BY_ROLE = {
    "engineering": [
        "/skills/code-review/",
        "/skills/testing/",
        "/skills/deployment/",
    ],
    "data": [
        "/skills/sql-analysis/",
        "/skills/visualization/",
        "/skills/data-pipeline/",
    ],
    "support": [
        "/skills/ticket-triage/",
        "/skills/runbook/",
    ],
}
```

This suggests that a Python `skills=` source can point directly to a leaf Skill directory:

```text
/skills/code-review/
└── SKILL.md
```

That does not match the current Python loader contract.

For every source path passed through `skills=`, Python calls `backend.ls(source_path)`, keeps the immediate subdirectories, and checks those subdirectories for `SKILL.md`. It does not check whether `source_path/SKILL.md` exists.

Python therefore expects this structure:

```text
<configured-source>/
├── skill-a/
│   └── SKILL.md
└── skill-b/
    └── SKILL.md
```

When the configured source is already a leaf Skill directory, the loader lists the contents of that Skill instead of treating the directory itself as a Skill. This can result in no Skills being discovered even though the leaf directory contains a valid `SKILL.md`.

This mismatch is reported in langchain-ai/deepagents#4820.

## What changed

Each role in the Python example now maps to a container directory:

```python
SKILLS_BY_ROLE = {
    "engineering": ["/skills/engineering/"],
    "data": ["/skills/data/"],
    "support": ["/skills/support/"],
}
```

The example also includes a compact directory tree:

```text
/skills/
├── engineering/
│   ├── code-review/
│   │   └── SKILL.md
│   └── testing/
│       └── SKILL.md
├── data/
│   └── sql-analysis/
│       └── SKILL.md
└── support/
    └── ticket-triage/
        └── SKILL.md
```

[...]

PR #5675, opened 2026-08-25
docs(deepagents): skills paths are source dirs, not skill dirs

magalhaes3:
## Overview

The "Load skills at runtime" sample on `/oss/deepagents/skills` passes individual skill directories in `skills`, which discovers no skills at all.

`SkillsMiddleware` treats every entry in `skills` as a *source* directory: it lists the path, keeps only the subdirectories, and looks for `SKILL.md` one level down (`_discover_skills_from_source` in `deepagents/middleware/skills.py`). A source pointing at `/skills/code-review/` therefore finds that skill's `references/`, `scripts/`, and `assets/` as candidate skills, none of which contain a `SKILL.md`, and loads nothing.

It fails silently. The source directory exists, `ls` succeeds, `source_error` stays `None`, and nothing is logged. The only symptom is `(No skills available yet. You can create skills in ...)` in the system prompt.

The rest of the page is already correct — "Pass the path to your top-level skills directory in the `skills` argument", and every other sample uses `/skills/`. The runtime-loading sample is the odd one out, and it's the one people copy when building per-user skill sets, which is exactly when a skills tree is laid out one-directory-per-skill and the mistake looks right.

Changes:

- `skills-dynamic-lists` (Python + TS) passes source directories (`/skills/shared/`, `/skills/engineering/`) instead of skill directories. Role names kept so the per-role layering still reads as the point of the sample.
- The `skills` `ParamField` states that each entry is a directory that *contains* skill directories.
- New bullet under "Skills missing at startup" for the wrong-path-level case, since none of the existing four bullets covers a failure that produces no error.
- `src/snippets/code-samples/skills-dynamic-lists-{py,js}.mdx` regenerated with `make code-snippets`, not hand-edited.

## Type of change

## Related issues/PRs

- GitHub issue: none — happy to open one and wait for approval if you'd rather have the discussion there first.
- Feature PR: n/a

## Checklist

## Additional notes

Validation, honestly:

- Reproduced the bug against `deepagents` 0.2.x with a real `FilesystemBackend` behind a `CompositeBackend`: per-skill source paths give `No skills available yet.`; the parent directory lists every skill with its description.
- `make code-snippets` is clean — the two generated mirrors in this PR are the only diff, so samples and snippets are in sync.
- Ran the changed Python snippet from `src/code-samples-generated/` and confirmed both branches construct (`create_agent_for_user("engineering")` and the unknown-role fallback).
- `make test-code-samples FILES="src/code-samples/deepagents/skills.py"` fails before reaching the assertions, on the unrelated `skills-subagents-py` block which needs `GOOGLE_API_KEY`. Same failure on `main`, so it isn't from this change, but I couldn't get a green full-file run.
- I did not run `docs dev` — no Mintlify setup locally. The prose edits are plain text plus one existing-anchor link, so rendering risk is low, but I haven't looked at the built page.
- No navigation change needed; no new pages or anchors.

### DIFF
--- a/src/code-samples/deepagents/skills-dynamic-lists.ts
+++ b/src/code-samples/deepagents/skills-dynamic-lists.ts
@@ -2,17 +2,9 @@
 import { createDeepAgent } from "deepagents";
 
 const SKILLS_BY_ROLE: Record<string, string[]> = {
-  engineering: [
-    "/skills/code-review/",
-    "/skills/testing/",
-    "/skills/deployment/",
-  ],
-  data: [
-    "/skills/sql-analysis/",
-    "/skills/visualization/",
-    "/skills/data-pipeline/",
-  ],
-  support: ["/skills/ticket-triage/", "/skills/runbook/"],
+  engineering: ["/skills/shared/", "/skills/engineering/"],
+  data: ["/skills/shared/", "/skills/data/"],
+  support: ["/skills/shared/", "/skills/support/"],
 };
 
 function createAgentForUser(userRole: string) {
--- a/src/code-samples/deepagents/skills.py
+++ b/src/code-samples/deepagents/skills.py
@@ -30,9 +30,9 @@ agent = create_deep_agent(
 from deepagents import create_deep_agent
 
 SKILLS_BY_ROLE = {
-    "engineering": ["/skills/code-review/", "/skills/testing/", "/skills/deployment/"],
-    "data": ["/skills/sql-analysis/", "/skills/visualization/", "/skills/data-pipeline/"],
-    "support": ["/skills/ticket-triage/", "/skills/runbook/"],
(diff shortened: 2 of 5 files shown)


magalhaes3:
Closing as a duplicate of #5669, which fixes the same Python sample and was opened first.

Mine also changed the TypeScript sample, which was wrong of me — JS `SkillsMiddleware` does support leaf skill directories (`deepagentsjs/libs/deepagents/src/middleware/skills.ts:619`), so that sample was correct as written. #5669 correctly left it alone.

I left the two prose additions from here as a comment on #5669 (the `skills` `ParamField` clarification and a troubleshooting bullet for the silent no-error case), language-fenced since the behavior differs between Python and JS. Happy to open those separately if they're wanted and #5669 stays scoped to the sample.

Underlying issue is langchain-ai/deepagents#4820.

PR #5701, opened 2026-08-26
integrations: add attenu-guard (middleware)

rafaelasor:
Adds one `python: middleware:` row for [attenu-guard](https://github.com/attenu-io/attenu-guard) (Apache-2.0, PyPI `attenu-guard`, extra `attenu-guard[langchain]`).

It uses the `AgentMiddleware.wrap_tool_call` / `ToolNode(wrap_tool_call=...)` hook to check every tool call against the agent's permission set before the tool body runs, and treats the deepagents `task(...)` call as the delegation point so a subagent's permissions are a computed subset of the parent's. Every decision lands in a hash-chained audit log verified offline. Docs page: https://attenu.io/docs/example-langgraph/ (same markdown as the repo's example README). Tested against langgraph 1.2 / langchain 1.3 / deepagents 0.7.

Disclosure: I maintain attenu-guard.


rafaelasor:
@lnhsingh ready for review — a single `python: middleware:` row (default track per the July 28 process), no hosted page requested.

npentrel (approved):

PR #5706, opened 2026-08-26
docs(callbacks): list AsqavCallbackHandler in the external download table

jagmarques:
Converted to the listing-only format per the new integration submission process
(effective 2026-07-28): asqav does not meet the hosted-guide threshold, so this adds the
component-table entry instead of a hosted page.

- Entry in `scripts/data/integration_external_docs.yaml` under python > callbacks
- `pypi: asqav` for the downloads badge (the LangChain callback ships in the `asqav` package,
  install with `pip install "asqav[langchain]"`)
- `docs_url` points to the partner docs (https://asqav.com/docs) per the doc-link priority

The previous hosted-page version is dropped from this PR.


jagmarques:
Ready for review — per the triage bot's guidance, requesting @mdrxy (Python integrations) when capacity allows. Converted to the listing-only format (YAML entry) per the July 28 process; the hosted-page version was dropped from this PR.

PR #5727, opened 2026-08-27
docs(a2a): document supported methods, capabilities, and conformance testing

HugoDurand:
## Overview

The A2A page was written for A2A v0.3 and never revisited. It listed three RPC methods, claimed a
`langgraph-api >= 0.4.21` floor, and documented none of the capabilities shipped since. A reader
would conclude that `CancelTask`, `ListTasks`, `FilePart` and A2UI do not exist.

Four changes:

1. **Protocol version + supported methods.** States that we speak the v1.0 JSON-RPC binding and also
   accept the v0.3 method names, and warns that the name you send selects the response shape
   (v1.0 wraps the task in `result.task` with `TASK_STATE_*`/`ROLE_*`, v0.3 returns it directly with
   lowercase enums). Replaces the three-item list with a table of all eight operations, including the
   two that return `-32601`.
2. **Optional capabilities.** New section for what shipped and was never documented: per-assistant
   input/output modes, `FilePart` in both directions, A2UI v0.9 opt-in, and the
   `A2A_ALLOWED_TOOL_CALL_RESULTS` allowlist.
3. **Test your integration.** New section. How to fetch the card and send a message against your own
   deployment, how to run the official TCK against it, and a link to
   `a2a-conformance-baseline.yml`, the checked-in list of every requirement we currently fail, with
   a reason and an owner per entry.
4. **Requirements.** Per-feature version table instead of a single stale floor.

## Type of change

## Verification

Every snippet and payload here was run against a live `langgraph-api` 0.15.0 server rather than
written from the source:

- agent card shape, including `protocolBinding: "JSONRPC"` and `protocolVersion: "1.0"`
- all eight operations in the table, under both naming families, confirming `SubscribeToTask` and
  the push-notification config methods return `-32601`
- the response-shape difference between `SendMessage` and `message/send`
- A2UI: card advertisement with `metadata.a2a.a2ui`, and a round trip returning
  `metadata.mimeType: "application/a2ui+json"`
- custom `input_modes`/`output_modes` flowing to both card defaults and skill modes
- `SendStreamingMessage` SSE, first event being the `Task`

`<Warning>` and `<Note>` match existing usage in this file and across `src/`. `make format-check`
could not run locally (`ruff` not installed in this environment), this change touches one `.mdx`
file and no Python.

## Question

Change 3 links the conformance baseline publicly. I think naming what we fail, with reasons, is
worth more to an integrator than a feature list, and the file is already public. Happy to drop that
subsection if we would rather not point at it.

### DIFF
--- a/src/langsmith/server-a2a.mdx
+++ b/src/langsmith/server-a2a.mdx
@@ -45,6 +45,23 @@ per client and stay on it.
 
 Only the JSON-RPC binding is available. gRPC and HTTP+JSON are not implemented.
 
+### Task history in responses
+
+A context holds many tasks. `SendMessage`, `GetTask` and `ListTasks` return the history of the
+**whole context**, not just the task you asked about. The second task in a context replays the
+first task's messages, so a client that renders every history entry shows earlier turns again —
+including earlier tool results and A2UI payloads.
+
+Two ways to work with this today: read only the entries after the ones you have already rendered,
+or start a new `contextId` when you want isolation. Per-task history is being added as a
+`historyScope` request option.
+
+<Warning>
+Do not resend a `taskId` from a completed task. Each new turn starts a new task inside the same
+context — send the `contextId` alone. A message naming a terminal task is rejected with `-32004`,
+and a `taskId` minted by another agent is rejected with `-32001`.
+</Warning>
+
 ## Agent card discovery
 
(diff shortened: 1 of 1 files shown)


HugoDurand:
> Thanks for opening a docs PR, Hugo DURAND (@HugoDurand)! When it's ready for review, please add the relevant reviewers:
> 
> * `@fjmorris` or `@lnhsingh` (LangSmith)

@hari-dhanushkodi @syachamaneni-lc @victorm-lc

syachamaneni-lc (approved):

PR #5749, opened 2026-08-28
integrations: add piighost (middleware)

Athroniaeth:
Adds **PIIAnonymizationMiddleware** (from `piighost`) to the middleware integrations list.

It de-identifies PII before the model sees it and restores the real values in the reply and at the tool boundary, with pluggable regex/NER/LLM detectors, thread-consistent tokens, and token-by-token streaming restoration.

Published on PyPI as `piighost` (`pip install "piighost[langchain]"`), MIT-licensed. Docs: https://athroniaeth.github.io/piighost/getting-started/langchain/


Athroniaeth:
Thanks for the quick triage. This follows the external integrations process: a single row in `scripts/data/integration_external_docs.yaml` under `python: middleware`, for the already-published `piighost` package, with no new MDX page. All checks are green (including `check-external-docs-urls`). Ready for review whenever the queue allows.

cc @lnhsingh

PR #5777, opened 2026-08-31
docs: list safeprompt-langchain (Python callback handler)

ianreboot:
Adds the Python entry for `SafePromptCallbackHandler` to `scripts/data/integration_external_docs.yaml`.

The JS/TS counterpart (`@safeprompt.dev/langchain`) is already listed in the `javascript.callbacks` section. This is the same integration for the Python runtime, published independently to PyPI as [`safeprompt-langchain`](https://pypi.org/project/safeprompt-langchain/).

```yaml
- name: SafePromptCallbackHandler
  pypi: safeprompt-langchain
  docs_url: https://docs.safeprompt.dev/langchain
```

It is a `BaseCallbackHandler` that screens prompts (and, in agents, tool outputs) for prompt injection before they reach the model.

Also updates the existing JS row's `docs_url` from the docs home to the integration guide, so both rows link to the page that actually documents the handler. That page now covers both runtimes.

Default listing tier, per the contributing guide, so no MDX page is added.


npentrel (approved):
Thanks!

PR #5778, opened 2026-08-31
Add langchain-ictfax to external tools integrations

tahiralmas:
Adds a row to `scripts/data/integration_external_docs.yaml` (python → tools) for **`langchain-ictfax`**, now on PyPI: https://pypi.org/project/langchain-ictfax/

`ICTFaxToolkit` provides tools to send and track faxes through an [ICTFax](https://ictfax.org) / ICTCore server (`send_fax`, `get_fax_status`, `list_faxes`, `upload_fax_document`). Under 50k monthly downloads, so this is the external-docs row rather than a hosted guide, per the contributing guide. `docs_url` points at the partner site (ICT Innovations), following the doc-link priority (partner docs > GitHub > PyPI).

### DIFF
--- a/packages.yml
+++ b/packages.yml
@@ -1370,3 +1370,7 @@ packages:
   name_title: PIIGhost
   repo: Athroniaeth/piighost
   js: "n/a"
+- name: langchain-ictfax
+  name_title: ICTFax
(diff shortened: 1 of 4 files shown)


npentrel (approved):
Thanks!

PR #1196, opened 2025-10-30
Update retriever imports to use langchain_classic for v1 compatibility

bart0401:
The langchain.retrievers module was removed in LangChain 1.0+. 
All retriever imports have been updated to use langchain_classic as documented in the v1 migration guide.

Fixes #1195

## Summary

Updated outdated retriever imports that don't work in LangChain 1.0+. Changed `from langchain.retrievers` to `from langchain_classic.retrievers` as per the v1 migration guide.

## Changes

31 files updated across integrations:
- 9 document_transformers files
- 7 retrievers files
- 12 providers files
- 2 vectorstores files
- 1 document_loaders file

Before:
```python
from langchain.retrievers import ContextualCompressionRetriever
from langchain.retrievers.multi_query import MultiQueryRetriever
```

After:
```python
from langchain_classic.retrievers.contextual_compression import ContextualCompressionRetriever
from langchain_classic.retrievers.multi_query import MultiQueryRetriever
```

All changes follow the pattern in the [v1 migration guide](https://docs.langchain.com/oss/migrate/langchain-v1#langchain-classic).

Issue #1195, opened 2025-10-30
[langchain]: Outdated Import Statements for Retrievers in Documentation


bart0401:
### Type of issue

issue / bug

### Language

Python

## Summary

Multiple documentation files contain outdated import statements using `from langchain.retrievers import ...`, which is incompatible with LangChain 1.0+. According to the [v1 migration guide](/src/oss/python/migrate/langchain-v1.mdx), all retriever imports should now use `langchain_classic.retrievers` instead.

## Impact

**23 documentation files** contain code examples that will fail when users try to run them with LangChain 1.0+.

### Error Users Will Encounter

```python
ImportError: cannot import name 'ContextualCompressionRetriever' from 'langchain.retrievers'
```

or

```python
ModuleNotFoundError: No module named 'langchain.retrievers'
```

### Document Transformers (7 files)
- `/src/oss/python/integrations/document_transformers/cross_encoder_reranker.mdx:61`
- `/src/oss/python/integrations/document_transformers/openvino_rerank.mdx:294`
- `/src/oss/python/integrations/document_transformers/voyageai-reranker.mdx:295`
- `/src/oss/python/integrations/document_transformers/jina_rerank.mdx:68`
- `/src/oss/python/integrations/document_transformers/volcengine_rerank.mdx:298`
- `/src/oss/python/integrations/document_transformers/dashscope_rerank.mdx:266`
- `/src/oss/python/integrations/document_transformers/infinity_rerank.mdx:283`

### Retrievers (5 files)
- `/src/oss/python/integrations/retrievers/merger_retriever.mdx:13`
- `/src/oss/python/integrations/retrievers/flashrank-reranker.mdx:274`
- `/src/oss/python/integrations/retrievers/llmlingua.mdx:264`
- `/src/oss/python/integrations/retrievers/fleet_context.mdx:20`
- `/src/oss/python/integrations/retrievers/re_phrase.mdx:18`

### Provider Integrations (11 files)
- `/src/oss/python/integrations/providers/chroma.mdx:34`
- `/src/oss/python/integrations/providers/cohere.mdx:28,143`
- `/src/oss/python/integrations/providers/ragatouille.mdx:107`
- `/src/oss/python/integrations/providers/wikipedia.mdx:35`
- `/src/oss/python/integrations/providers/outline.mdx:23`
- `/src/oss/python/integrations/providers/vespa.mdx:28`
- `/src/oss/python/integrations/providers/metal.mdx:18`
[...]

PR #5791, opened 2026-08-31
docs: fix outdated 'langchain import hub' to use langchain_classic

SpiliosDimakopoulos:
## What
Fixes an outdated import in two integration docs that raises
`ImportError: cannot import name 'hub' from 'langchain'` under
LangChain 1.0+.

- src/oss/python/integrations/retrievers/egnyte.mdx
- src/oss/python/integrations/tools/ibm_watsonx_sql.mdx

## Why
Per the LangChain v1 migration guide (src/oss/python/migrate/langchain-v1.mdx),
the `hub` module moved from the `langchain` package to `langchain-classic`
in v1. These two pages still show:

    from langchain import hub

which fails on LangChain 1.0+. Changed to:

    from langchain_classic import hub

This follows the same pattern already applied to retriever imports in
PR #1196 (issue #1195), which fixed the same class of outdated import
for a different module.

## Type of change

2-line, docs-only change with no functional impact outside the code samples.

### DIFF
--- a/src/oss/python/integrations/retrievers/egnyte.mdx
+++ b/src/oss/python/integrations/retrievers/egnyte.mdx
@@ -161,7 +161,7 @@ results = await retriever.abatch(
 Like other retrievers, EgnyteRetriever can be added to a LangGraph agent as a tool.
 
 ```python
-from langchain import hub
+from langchain_classic import hub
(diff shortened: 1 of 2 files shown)


npentrel (approved):
Thanks!

PR #1619, opened 2025-11-25
add missing import

dgtql:
Missing 'from langchain.agents.structured_output import ToolStrategy' in the example code. Users will fail to run through the example code without this line.

## Type of change

**Type:** [Replace with: New documentation page / Update existing documentation / Fix typo/bug/link/formatting / Remove outdated content / Other]

## Related issues/PRs

- GitHub issue:
- Feature PR:

- Linear issue:
- Slack thread:

## Checklist

- [x] I have read the [contributing guidelines](README.md)
- [x] I have tested my changes locally using `docs dev`
- [x] All code examples have been tested and work correctly
- [x] I have used **root relative** paths for internal links
- [x] I have updated navigation in `src/docs.json` if needed

(Internal team members only / optional): Create a preview deployment as necessary using the [Create Preview Branch workflow](https://github.com/langchain-ai/docs/actions/workflows/create-preview-branch.yml)

PR #5792, opened 2026-08-31
docs: add missing agent imports in AWS and Neo4j provider examples

SpiliosDimakopoulos:
## What
Two integration provider pages call an agent-creation function that
is never imported in the code sample, so copying the example as-is
raises NameError.

- src/oss/python/integrations/providers/aws.mdx
  Two Bedrock AgentCore toolkit examples (browser + code interpreter)
  call `create_react_agent(...)` without importing it.
  Added: `from langgraph.prebuilt import create_react_agent`

- src/oss/python/integrations/providers/neo4j.mdx
  The Neo4jSaver checkpointer example calls `create_agent(...)`
  without importing it.
  Added: `from langchain.agents import create_agent`

## Why
Both symbols are used but never defined/imported anywhere else on
the page, so running the snippet as written fails immediately.
Same bug class as #1619 ("add missing import"), which was merged
for a different page.

## Type of change

3-line, docs-only change, no functional impact outside the code samples.

### DIFF
--- a/src/oss/python/integrations/providers/aws.mdx
+++ b/src/oss/python/integrations/providers/aws.mdx
@@ -168,6 +168,7 @@ from langchain_aws import AmazonKnowledgeBasesRetriever
 See a [usage example](/oss/integrations/tools/bedrock_agentcore_browser).
 
 ```python
+from langgraph.prebuilt import create_react_agent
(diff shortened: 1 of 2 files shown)


npentrel (approved):
Thanks!

PR #5793, opened 2026-08-31
docs(langsmith): add filter-traces page for the rebuilt tracing filter UI

fjmorris:
## Why

Linear: DOC-1570

This is stage 2 of 2 for the LangSmith filtering revamp (Milestone 1), which launched September 10.

Stage 1 was #5975, which added `src/langsmith/filter-traces.mdx` unlisted, so beta customers could be sent a direct link ahead of launch. That page is now merged and live at [/langsmith/filter-traces](https://docs.langchain.com/langsmith/filter-traces).

## What this PR does

- **Makes the page discoverable.** Adds `langsmith/filter-traces` to `src/docs.json` under Monitor > Debug > "Viewing & managing traces", directly above the ClickHouse page.
- **Disambiguates the two pages.** Retitles `filter-traces-in-application` to "Filter traces (ClickHouse)" so the two no longer share a sidebar label, and gives it the `description` it was missing. Its body is otherwise unchanged: the ClickHouse behavior it documents is not affected by this revamp.
- **Routes readers between them.** Renders `snippets/langsmith/filtering-experience-note.mdx` at the top of both pages. The note routes on the filter UI a reader can see, a single search bar with a scope selector or an **Add filter** button that builds chips, rather than on a backend name they have no way of knowing.
- **Repoints inbound links.** 17 links across 16 files now target `/langsmith/filter-traces`. Counting the snippet, which already carried one, 18 links across the docs point at the new page.
- **Corrects the default scope for the Traces table selection.** See below. This is the only change to `filter-traces.mdx` itself.

## How the links were split

The rule: a link that lands the reader on the tracing project table goes to the new page, because that table is what the revamp changed. A link sitting in an automation rule, evaluator, or chart context stays on the ClickHouse page, because those panes keep the chip builder, and the link text now names it.

Nine files deliberately keep pointing at the ClickHouse page: `billing.mdx`, `dashboards.mdx`, `ls-metadata-parameters.mdx`, `rules.mdx`, `webhooks.mdx`, `online-evaluations-code.mdx`, `online-evaluations-llm-as-judge.mdx`, plus `filter-traces-in-application.mdx` and the snippet themselves.

Four sentences claiming those filters "work the same way" as project filters were rewritten, since that stops being true for SmithDB projects. They now take the shape "X use the filter builder, with the same fields and operators described in [Filter traces (ClickHouse)]".

## The default scope correction

Requested by @lc-arjun on DOC-1570 after the page went live: the default scope of the traces tab changed from **Any run** to **Root run**, and the scope table on the live page still carried the old value. A reader following that table would expect a first Traces query to match every run in the tree when it actually matches entry runs only.

The Examples section states that each query assumes the default scope for its table selection, so fixing the cell changed what two Traces examples mean:

- **"Traces that called a specific tool"** (`run_type:tool AND name:search_documents`) now names the **Any run** scope explicitly. Under **Root run** it matches only traces whose entry run is a tool call, which is the mistake the Troubleshoot a query section warns about.
- **"Expensive traces"** no longer names the **Root run** scope, which is now the default and does not need calling out.

Available scopes for Traces (Root run, Any run) are unchanged, as are the defaults for Threads and Runs.

## Needs careful review

1. **The link split itself.** This was asked on the Linear ticket and is still unanswered. The split above is my best reading of the PRD scope (tracing page and table only), not a confirmed product answer.
2. **`engine.mdx`** loses its cross-reference outright rather than being repointed, because the page already describes its two-condition control inline.
3. **`configure-input-output-preview.mdx`** is split into two links, one per filter experience.
4. **`threads.mdx`** now links to `#save-a-filter` on the new page. Worth confirming that anchor is the right destination for that sentence.

## Verification

- `make lint_prose` on all 26 changed MDX files: 0 errors, 0 warnings, 0 suggestions
- `make broken-links`: no broken links, with the nav entry in place

## Follow-ups (not in this PR)

- Confirm whether the Tip on both `online-evaluations-*` pages still holds: "Any filters you apply to the runs table will automatically be reflected in filters on your evaluator." The runs table and the evaluator pane can now use different filter UIs, so this needs a product answer rather than a code check

## AI disclosure

Drafted by Claude Opus 5 via Claude Code, reviewed and edited by me. Source material: the DOC-1570 Linear issue and its attachments, the filter revamp PRD and filter syntax spec, and the `langchain-ai/langchainplus` implementation.

### DIFF
--- a/src/docs.json
+++ b/src/docs.json
@@ -2053,6 +2053,7 @@
                     "group": "Viewing & managing traces",
                     "pages": [
                       "langsmith/view-traces",
+                      "langsmith/filter-traces",
                       "langsmith/filter-traces-in-application",
                       "langsmith/configure-input-output-preview",
                       "langsmith/manage-trace",
--- a/src/langsmith/add-metadata-tags.mdx
+++ b/src/langsmith/add-metadata-tags.mdx
@@ -7,7 +7,7 @@ LangSmith supports sending arbitrary metadata and tags along with traces.
 
 Tags are strings that can be used to categorize or label a trace. Metadata is a dictionary of key-value pairs that can be used to store additional information about a trace.
 
-Both are useful for associating additional information with a trace, such as the environment in which it was executed, the user who initiated it, or an internal correlation ID. For more information on tags and metadata, see the [Concepts](/langsmith/observability-concepts#tags) page. For information on how to query traces and runs by metadata and tags, see the [Filter traces in the application](/langsmith/filter-traces-in-application) page.
+Both are useful for associating additional information with a trace, such as the environment in which it was executed, the user who initiated it, or an internal correlation ID. For more information on tags and metadata, see the [Concepts](/langsmith/observability-concepts#tags) page. For information on how to query traces and runs by metadata and tags, see the [Filter traces](/langsmith/filter-traces) page.
 
 <CodeGroup>
 
--- a/src/langsmith/attach-user-feedback.mdx
+++ b/src/langsmith/attach-user-feedback.mdx
@@ -100,4 +100,4 @@ If you need to collect feedback from a browser or other client-side environment 
 
 See [Collect feedback with presigned URLs](/langsmith/presigned-feedback-tokens) for the full guide.
 
-To learn more about how to filter traces based on various attributes, including user feedback, see [Filter traces](/langsmith/filter-traces-in-application).
+To learn more about how to filter traces based on various attributes, including user feedback, see [Filter traces](/langsmith/filter-traces).
--- a/src/langsmith/composite-evaluators-ui.mdx
+++ b/src/langsmith/composite-evaluators-ui.mdx
@@ -53,7 +53,7 @@ Composite scores are attached to a run as **feedback**, similarly to feedback fr
 
 **On a tracing project**:
 - Composite scores appear as feedback on runs.
-- [Filter for runs](/langsmith/filter-traces-in-application) with a composite score, or where the composite score meets a certain threshold.
+- [Filter for runs](/langsmith/filter-traces) with a composite score, or where the composite score meets a certain threshold.
 - [Create a chart](/langsmith/dashboards#custom-dashboards) to visualize trends in the composite score over time.
 
 **On a dataset**:
--- a/src/langsmith/configure-input-output-preview.mdx
+++ b/src/langsmith/configure-input-output-preview.mdx
@@ -114,6 +114,6 @@ If you see `"No paths available"` in the tree:
 
 ## Next steps
 
-- Learn more about [viewing and filtering traces](/langsmith/filter-traces-in-application).
+- Learn more about [viewing](/langsmith/view-traces) and [filtering](/langsmith/filter-traces) traces.
 - Explore [custom output rendering](/langsmith/custom-output-rendering) for advanced visualization.
 - Set up [metadata and tags](/langsmith/add-metadata-tags) to organize your traces.
--- a/src/langsmith/dashboards.mdx
+++ b/src/langsmith/dashboards.mdx
@@ -108,7 +108,7 @@ When you add a filter, it defaults to filtering at the [run](/langsmith/observab
 
 The active scope appears as a suffix on the **Advanced** item (for example, **Advanced Tree Filter**). Click the **X** next to it to reset back to a plain run filter.
 
-Dataset sources do not expose run/trace/tree filters. Data is scoped by the selected dataset. For filter syntax, refer to [filtering traces in application](/langsmith/filter-traces-in-application).
+Dataset sources do not expose run/trace/tree filters. Data is scoped by the selected dataset. For filter syntax, refer to [filtering traces (ClickHouse)](/langsmith/filter-traces-in-application).
 
 **Grouping** creates multiple series on the same chart in one of two ways:
 
@@ -166,7 +166,7 @@ Create tailored collections of charts for tracking metrics that matter most for 
 #### Select tracing projects and filter runs
 
 - Select one or more tracing projects to track metrics for.
-- Use the **Chart filters** section to refine the matching runs. This filter applies to all data series in the chart. For more information, view the guide on [filtering traces in application](/langsmith/filter-traces-in-application).
+- Use the **Chart filters** section to refine the matching runs. This filter applies to all data series in the chart. For more information, view the guide on [filtering traces (ClickHouse)](/langsmith/filter-traces-in-application).
 
 #### Pick a metric
 
--- a/src/langsmith/engine.mdx
+++ b/src/langsmith/engine.mdx
@@ -150,7 +150,7 @@ Set the scope in either of two places, using the same control:
 - **Engine setup**: In the **Find and fix your agent's issues** panel, under **Focus on specific traces**.
 - **Engine Settings**: In the **Focus on specific traces** section of the [**Engine Settings**](#configure-engine) panel. Edits here save automatically.
 
-Add scope conditions with the same [filter editor](/langsmith/filter-traces-in-application#create-and-apply-filters) used on the tracing project's **Tracing** tab. You can add one condition of each kind, **up to two**:
+Add scope conditions with the filter editor. You can add one condition of each kind, **up to two**:
 
 - **Run Name**: Pick a run or agent name. The value field autocompletes from the run names in your project's recent traces.
 - **Metadata**: Pick a metadata key, then a value. Both autocomplete from the metadata present on your project's recent runs.
--- a/src/langsmith/filter-traces-in-application.mdx
+++ b/src/langsmith/filter-traces-in-application.mdx
@@ -1,9 +1,14 @@
 ---
-title: Filter traces
-sidebarTitle: Filter traces
+title: Filter traces (ClickHouse)
+sidebarTitle: Filter traces (ClickHouse)
+description: Filter threads, traces, and runs in a LangSmith tracing project backed by ClickHouse, using the filter bar, filter shortcuts, and saved filters.
 ---
 
+import FilteringExperienceNote from '/snippets/langsmith/filtering-experience-note.mdx';
+
 Tracing projects can accumulate large amounts of data across [threads](/langsmith/observability-concepts#threads), [traces](/langsmith/observability-concepts#traces), and [runs](/langsmith/observability-concepts#runs). LangSmith's filtering tools let you navigate and analyze that data precisely.
+
+<FilteringExperienceNote />
 
 This page covers:
 
--- a/src/langsmith/log-llm-trace.mdx
+++ b/src/langsmith/log-llm-trace.mdx
@@ -389,7 +389,7 @@ def chat_model(inputs: dict) -> dict:
 
 ## Identify a custom model in traces
 
-When using a custom model, it is recommended to also provide the following `metadata` fields to identify the model when viewing traces and when [filtering](/langsmith/filter-traces-in-application).
+When using a custom model, it is recommended to also provide the following `metadata` fields to identify the model when viewing traces and when [filtering](/langsmith/filter-traces).
 
 - `ls_provider`: The provider of the model, e.g., `"openai"`, `"anthropic"`.
 - `ls_model_name`: The name of the model, e.g., `"gpt-5.4-mini"`, `"claude-opus-4-8"`.
--- a/src/langsmith/ls-metadata-parameters.mdx
+++ b/src/langsmith/ls-metadata-parameters.mdx
@@ -10,7 +10,7 @@ import LsMetadataParametersConfiguredKt from '/snippets/code-samples/ls-metadata
 
 When you trace LLM calls with LangSmith, you often want to [track costs](/langsmith/cost-tracking), compare model configurations, and analyze performance across different providers. LangSmith's native integrations (like [LangChain](/langsmith/trace-with-langchain) or the [OpenAI](/langsmith/trace-openai)/[Anthropic](/langsmith/trace-anthropic) wrappers) handle this automatically, but custom model wrappers and self-hosted models require a standardized way to provide this information. LangSmith uses `ls_` metadata parameters for this purpose.
 
-These metadata parameters (all prefixed with `ls_`) let you pass model configuration and identification information through the standard `metadata` field. Once set, LangSmith can automatically calculate costs, display model information in the UI, and enable [filtering](/langsmith/filter-traces-in-application) and analytics across your traces.
+These metadata parameters (all prefixed with `ls_`) let you pass model configuration and identification information through the standard `metadata` field. Once set, LangSmith can automatically calculate costs, display model information in the UI, and enable [filtering](/langsmith/filter-traces) and analytics across your traces.
 
 Use `ls_` metadata parameters to:
 
@@ -642,4 +642,4 @@ metadata_key = 'ls_run_depth' AND metadata_value = 0
 - [Trace query syntax](/langsmith/trace-query-syntax): Complete reference for filtering and searching traces.
 - [Evaluation quickstart](/langsmith/evaluation-quickstart): Run experiments on datasets to compare model configurations.
 - [Add metadata and tags](/langsmith/add-metadata-tags): General guide to adding metadata to traces.
-- [Filter traces in application](/langsmith/filter-traces-in-application): Programmatically filter traces in your code.
+- [Filter traces (ClickHouse)](/langsmith/filter-traces-in-application): Programmatically filter traces in your code.
--- a/src/langsmith/manage-datasets-in-application.mdx
+++ b/src/langsmith/manage-datasets-in-application.mdx
@@ -25,7 +25,7 @@ The following sections explain the different ways you can create a dataset in La
 A common pattern for constructing datasets is to convert notable traces from your application into dataset examples. This approach requires that you have [configured tracing to LangSmith](/langsmith/observability-concepts).
 
 <Check>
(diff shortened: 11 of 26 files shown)


src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+This page covers:
+
+- [Building a query](#build-a-query) in the search bar
+- [Query syntax](#query-syntax) and the [fields](#fields) available to filter on
+- [Choosing what a filter matches](#choose-what-a-filter-matches) with scopes
+- [Filtering in one click](#filter-with-shortcuts) from the **Shortcuts** panel
+- [Saving a filter](#save-a-filter) as a view
+- [Filtering the runs inside a trace](#filter-the-runs-in-a-trace)
+
+To query runs programmatically instead, refer to [Query traces using the SDK](/langsmith/export-traces). The SDK and REST API use a separate structured query language, documented in [Trace query syntax](/langsmith/trace-query-syntax).
```
lc-arjun, line 25: note (no action needed): eventually we will update sdk to use this language too

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+The operators a field accepts depend on its type.
+
+| Operator | Syntax | Matches |
+| -------- | ------ | ------- |
+| is | `field:value` | An exact match on the value. |
+| is not | `-field:value` | Everything except an exact match. |
+| matches | `field:~value` | A full-text search of the field. Quote the value to search for a phrase. |
+| does not match | `-field:~value` | Everything the full-text search does not match. |
+| wildcard | `field:*value*` | A pattern, where `*` stands for any sequence of characters. |
```
lc-arjun: for wildcard maybe we only show one `*` in the example

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+| exists | `field:*` | Records where the field is set. |
+| does not exist | `-field:*` | Records where the field is not set. |
+| greater than | `field:>value` | Values above the given value. |
+| greater than or equal to | `field:>=value` | Values at or above the given value. |
+| less than | `field:<value` | Values below the given value. |
+| less than or equal to | `field:<=value` | Values at or below the given value. |
+| inclusive range | `field:[10 TO 20]` | Values from 10 to 20, including both bounds. |
+| exclusive range | `field:{10 TO 20}` | Values between 10 and 20, excluding both bounds. |
+| mixed range | `field:[10 TO 20}` | Values from 10 up to but not including 20. Swap the delimiters to invert which bound is inclusive. |
+| unbounded range | `field:[10 TO *]` | Values from 10 upward. Use `*` in place of either bound. |
```
lc-arjun: we can document this, but we removed it from the prodcut as its the same technically as > or <

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+| `name` | Run name. Supports `*` wildcards. |
+| `status` | Run status: `pending`, `success`, `error`, or `interrupted`. |
+| `run_type` | The [type of run](/langsmith/run-data-format#run-types), for example `llm`, `chain`, `tool`, or `retriever`. |
+| `tags` | Run tags. A clause matches if any tag on the run matches. |
+
+**Content**
+
+| Field | Description |
+| ----- | ----------- |
+| `inputs` | Run inputs. Use `~` to search the text, or `inputs.<key>` to match a JSON key. |
```
lc-arjun: renamed to input

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+| `outputs` | Run outputs. Use `~` to search the text, or `outputs.<key>` to match a JSON key. |
```
lc-arjun: renamed to output

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+| `metadata` | Run metadata. Use `~` to search the text, or `metadata.<key>` to match a JSON key. |
+| `error` | Error text. Use `~` to search the text. |
+| `attachments` | Attachment names. |
```
lc-arjun, line 222: i believe this only supports exists/not exists

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+- **Any run**: Matches anywhere in the tree.
+
+Each option in the scope dropdown includes a preview illustrating which runs it covers.
+
+The scopes offered, and the one selected by default, depend on the table selection:
+
+| Table selection | Default scope | Available scopes |
+| --------------- | ------------- | ---------------- |
+| **Threads** | Thread | Thread, Root run, Any run |
+| **Traces** | Root run | Root run, Any run |
```
lc-arjun: we updated the default for threads and traces to be `any run`

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+1. Click **Save View**.
+1. Enter a name and description, then save.
+
+The view then appears in the **Views** dropdown alongside **Default view**. Each entry lists the scopes its filter uses, which shows at a glance whether it applies to the current table selection.
+
+To update a saved view, select it, change the query, then click **Save View**. To rename or delete a view, click the <Icon icon="dots-vertical"/> icon next to it in the **Views** dropdown.
+
+To discard unsaved changes and return to the selected view, click **Reset**.
+
+### Views saved with the previous syntax
```
lc-arjun, line 386: this will be changed and we will actually translate the syntax for the user into v2. more specifics to come

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+Views saved before this filtering experience shipped continue to work. LangSmith still runs their original queries, and applying one restores its filters and selects a compatible table selection.
+
+Because those views use the structured query language rather than the syntax on this page, they are read-only in the search bar and cannot be renamed. A warning icon marks them in the **Views** dropdown. To move a view to the new syntax, rebuild its filters in the search bar and save the result as a new view.
+
+## Filter the runs in a trace
```
lc-arjun, line 390: maybe also worth documenting somewhere that filters applied using the `any run` scope will get auto applied into the tree view. this makes it easy to see which runs matched the filter

lc-arjun (commented):
leaving some comments, this looks really great though!

src/langsmith/filter-traces.mdx
fjmorris: update when changes are available

src/langsmith/filter-traces.mdx
fjmorris: updated with:
"A filter applied in the **Any run** scope carries over automatically, so opening a thread or trace from the table shows which of its runs matched."

fjmorris (commented):
update already committed.

src/langsmith/filter-traces.mdx
fjmorris: updated

src/langsmith/filter-traces.mdx
fjmorris: removed

src/langsmith/filter-traces.mdx
fjmorris: updated

src/langsmith/filter-traces.mdx
fjmorris: updated

src/langsmith/filter-traces.mdx
fjmorris: Checked this against the schema and it looks like the restriction is not there. I
ran the regenerated WASM from current origin/main:

  {
    "name": "attachments",
    "kind": "wildcard_string",
    "doc": "Attachment names",
    "operators": ["is", "is_not", "wildcard", "exists"]
  }

So attachments takes is, is_not, wildcard, and exists, and nothing in
isHiddenOperator suppresses any of them for this field. The field's own doc
string, "Attachment names", also reads as name-matchable rather than
presence-only.

I had applied your comment as written, so the page briefly claimed exists and
does not exist only. Now corrected to:

  | `attachments` | Attachment names. Accepts the is, is not, wildcard, and
  exists operators. |

and I dropped the matching entry from the operator-constraints list, since
attachments is then unremarkable, accepting everything except full-text ~ and
the comparisons and ranges.

Flagging in case the product intends the restriction and the schema has not
caught up. If the intent is presence-only, I will put the narrower wording back
and we should probably trim the schema's operator list to match.

src/langsmith/filter-traces.mdx
fjmorris: updated

fjmorris (commented):
commited updates

src/langsmith/filter-traces.mdx
lc-arjun: you're right! i had the wrong understanding. thanks

fjmorris:
@lnhsingh @lc-arjun I think this pr is complete. Please review the new page: [src/langsmith/filter-traces.mdx](https://langchain-5e9cc07a-preview-fjmorr-1788457050-a9421f9.mintlifysite.com/langsmith/filter-traces).

Arjun: I did a docs review rewrite.
Lauren: I very carefully repointed the links to the new page where necessary so you don't have to spend time reviewing those changes.

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
...
+<img
+    className="hidden dark:block"
+    src="/langsmith/images/filter-search-bar-dark.png"
+    alt="The filter toolbar of a tracing project, with a scope selector reading 'in any run' beside an empty search field, a Filter button on the right, and a second row holding the time range control, the Threads, Traces, and Runs selection, Reset, Save View, and Views"
+/>
+
+Click the search bar and start typing to build one:
+
+<Steps>
+  <Step title="Choose a field">
```
lnhsingh, line 50:
```suggestion
  <Step title="Choose a field.">
```

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
...
+    Type a field name or select one from the suggestion list. Suggestions are grouped by category, and **Recents** lists queries run previously. Use the arrow keys to move through the list and press `Enter` to select.
+  </Step>
+  <Step title="Choose an operator">
```
lnhsingh, line 53:
```suggestion
  <Step title="Choose an operator.">
```

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
...
+    Type `:` to see the operators the field supports, each with an example.
+  </Step>
+  <Step title="Enter a value">
```
lnhsingh, line 56:
```suggestion
  <Step title="Enter a value.">
```

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
...
+    alt="The filter toolbar of a tracing project, with a scope selector reading 'in any run' beside an empty search field, a Filter button on the right, and a second row holding the time range control, the Threads, Traces, and Runs selection, Reset, Save View, and Views"
+/>
+
+<img
+    className="hidden dark:block"
+    src="/langsmith/images/filter-search-bar-dark.png"
+    alt="The filter toolbar of a tracing project, with a scope selector reading 'in any run' beside an empty search field, a Filter button on the right, and a second row holding the time range control, the Threads, Traces, and Runs selection, Reset, Save View, and Views"
+/>
+
+Click the search bar and start typing to build one:
```
lnhsingh:
```suggestion
To build a query, click the search bar and start typing:
```

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
+---
+title: Filter traces
+sidebarTitle: Filter traces
+description: Find specific threads, traces, and runs in a LangSmith tracing project by writing filter queries in the search bar.
+---
+
+import FilteringExperienceNote from '/snippets/langsmith/filtering-experience-note.mdx';
+
+Filtering is how you get from every [run](/langsmith/observability-concepts#runs) in a tracing project to the few you need: the conversations that errored, the tool calls that ran slowly, the runs a low feedback score points at. Write a query in the search bar, and choose which runs in the [trace](/langsmith/observability-concepts#traces) or [thread](/langsmith/observability-concepts#threads) hierarchy that query is tested against.
```
lnhsingh:
```suggestion
Filtering narrows the [runs](/langsmith/observability-concepts#runs) in a tracing project to the ones that need attention, such as the conversations that errored, slow tool calls, or runs with low feedback scores. Write a query in the search bar, then choose which runs in the [trace](/langsmith/observability-concepts#traces) or [thread](/langsmith/observability-concepts#threads) hierarchy the query is matched against.
```

lnhsingh: suggestion to tighten a little

lnhsingh (approved):
nice job! just a few nits you can take or leave

src/langsmith/filter-traces.mdx
```diff
@@ -0,0 +1,448 @@
...
+| wildcard | `field:value*` | A pattern, where `*` stands for any sequence of characters. |
+| exists | `field:*` | Records where the field is set. |
+| does not exist | `-field:*` | Records where the field is not set. |
+| comparison | `field:>value` | Values above the given value. Also accepts `>=`, `<`, and `<=`. |
+| inclusive range | `field:[10 TO 20]` | Values from 10 to 20, including both bounds. |
+| exclusive range | `field:{10 TO 20}` | Values between 10 and 20, excluding both bounds. |
+| mixed range | `field:[10 TO 20}` | Values from 10 up to but not including 20. Swap the delimiters to invert which bound is inclusive. |
+
+<Note>
+`*` means two different things depending on where it appears. On its own after the colon it tests for existence, so `name:*` matches every run that has a name. Inside a value it is a wildcard, so `name:docs*` matches names beginning with `docs`.
```
lc-arjun: i dont know if we actually need this callout. technically

`name:*` means exists bc the `*` serves as a wildcard for any sequence of characters

lc-arjun (approved):
docs look great, just one comment!

src/langsmith/filter-traces.mdx
fjmorris: Applied in #5975 (aafe4011), but without the trailing period.

439 `<Step title="...">` across `src/` carry no trailing period and 3 do, so adding it here would make this the only procedure in the LangSmith docs punctuated that way. Happy to add the periods if you would rather move the convention that direction.

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

src/langsmith/filter-traces.mdx
fjmorris: Applied in #5975 (aafe4011), but without the trailing period.

439 `<Step title="...">` across `src/` carry no trailing period and 3 do, so adding it here would make this the only procedure in the LangSmith docs punctuated that way. Happy to add the periods if you would rather move the convention that direction.

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

src/langsmith/filter-traces.mdx
fjmorris: Applied in #5975 (aafe4011), but without the trailing period.

439 `<Step title="...">` across `src/` carry no trailing period and 3 do, so adding it here would make this the only procedure in the LangSmith docs punctuated that way. Happy to add the periods if you would rather move the convention that direction.

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

src/langsmith/filter-traces.mdx
fjmorris: Applied verbatim in #5975 (aafe4011).

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

src/langsmith/filter-traces.mdx
fjmorris: Applied verbatim in #5975 (aafe4011), taking the tightened version from your follow-up.

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

src/langsmith/filter-traces.mdx
fjmorris: Removed in #5975 (aafe4011).

You are right that it is one behavior rather than two: `*` stands for any sequence of characters, so `field:*` matches anything, which is what exists means. The operator table above the deleted callout still lists wildcard and exists as separate rows, since that is how the operator dropdown presents them. Tell me if you would rather collapse those rows too.

On the location: the page has moved out of this PR. #5975 ships `filter-traces.mdx` unlisted ahead of the September 14 launch, so beta testers can be sent a link, and this PR keeps the nav entry, the ClickHouse retitle, and the inbound link sweep for launch day.

PR #5795, opened 2026-08-31
docs(js): add missing HumanMessage import in OpenAI multi-turn image example

SpiliosDimakopoulos:
## What
The "Multi-turn editing" TypeScript example under Image Generation
constructs `new HumanMessage(...)` but the file never imports
HumanMessage, so the snippet fails to compile/run as written.

## Why
`HumanMessage` is used but not imported anywhere in this page.
Added: `import { HumanMessage } from "@langchain/core/messages";`
— the same path already used for this class elsewhere on the site.

## Type of change

2-line, docs-only change, no functional impact outside the code sample.


npentrel (approved):
Thanks!

PR #5867, opened 2026-09-03
fix: replace dict[str, Any] tool param that breaks OpenAI strict mode

### LINKED ISSUE #3161: Code is throwing BadRequestError while executing Long-Term-Memory code block
### Type of issue

issue / bug

### Language

Python

### Description

Your existing code in langchain page [LINK](https://docs.langchain.com/oss/python/langchain/tools#long-term-memory-store) is throwing below error

BadRequestError: Error code: 400 - {'error_code': 'BAD_REQUEST', 'message': 'Invalid JSON schema - the "additionalProperties" keyword must be False or not specified\n'}

Image

Solution
=====
Kindly, replace that given code snippet with below one: -

```
from typing import Any
from langgraph.store.memory import InMemoryStore
from langchain.agents import create_agent
from langchain.tools import tool, ToolRuntime
import json

@tool
def get_user_info(
    user_id: str,
    runtime: ToolRuntime
) -> str:
    """Look up user info by user_id."""
    store = runtime.store
    user_info = store.get(("users",), user_id)
    return json.dumps(user_info.value) if user_info else "Unknown user"

@tool
def save_user_info(
    user_id: str,
    user_info_json: str,
    runtime: ToolRuntime
) -> str:
    """Save user info. user_info_json should be a JSON string."""
    store = runtime.store
    user_info = json.loads(user_info_json)
    store.put(("users",), user_id, user_info)
    return "Successfully saved user info."

store = InMemoryStore()
agent = create_agent(
    model,
    tools=[get_user_info, save_user_info],
    store=store
)

# Save user info
agent.invoke({
    "messages": [
        {
            "role": "user",
            "content": "Save the following user: userid: abc123, name: Foo, age: 25, email: foo@langchain.dev"
        }
    ]
})

agent.invoke({
    "messages": [
        {
            "role": "user",
            "content": "Save the following user: userid: xyz123, name: Zoo, age: 26, email: zoo@langchain.dev"
        }
    ]
})

# Get user info
result = agent.invoke({
    "messages": [
        {
            "role": "user",
            "content": "Get user info for user with id 'abc123'"
        }
    ]
})

print(result["messages"][-1].content)
# Here is the user info for user with ID "abc123":
# - Name: Foo
# - Age: 25
# - Email: foo@langchain.dev

```
My above code is able to return below output
_The user with id 'abc123' has the following information: name: Foo, age: 25, email: foo@langchain.dev._

LLM will extract user details as a JSON string. You need to parse this JSON string and then store it in memory.

SpiliosDimakopoulos:
## What
Replaces the `user_info: dict[str, Any]` parameter in the `save_user_info`
tool example on the Tools page with explicit `name`, `age`, `email` fields.

## Why
`dict[str, Any]` generates a JSON schema with `additionalProperties: true`
on that field (verified directly against installed langchain), which
OpenAI's strict function-calling mode rejects with:
`BadRequestError: Invalid JSON schema - the "additionalProperties"...`

This is the exact error reported in #3161 (Long-Term-Memory code block
throwing BadRequestError). That page was already fixed via a TypedDict
approach, but the same anti-pattern remained here. The fix matches the
JS version of this same example, which already uses a flat Zod schema.

Closes #3161

## Type of change
- [x] Fix typo/bug/link/formatting

Diff size: 1 file changed, 2 insertions(+), 3 deletions(-)

### DIFF
--- a/src/oss/langchain/tools.mdx
+++ b/src/oss/langchain/tools.mdx
@@ -328,7 +328,6 @@ Access the store through `runtime.store`. The store uses a namespace/key pattern
 </Tip>
 
 ```python expandable
-from typing import Any
 from langgraph.store.memory import InMemoryStore
 from langchain.agents import create_agent
 from langchain.tools import tool, ToolRuntime
@@ -344,10 +343,10 @@ def get_user_info(user_id: str, runtime: ToolRuntime) -> str:
 
 # Update memory
 @tool
-def save_user_info(user_id: str, user_info: dict[str, Any], runtime: ToolRuntime) -> str:
+def save_user_info(user_id: str, name: str, age: int, email: str, runtime: ToolRuntime) -> str:
     """Save user info."""
     store = runtime.store
-    store.put(("users",), user_id, user_info)
+    store.put(("users",), user_id, {"name": name, "age": age, "email": email})
     return "Successfully saved user info."
 
 model = ChatOpenAI(model="gpt-5.5")


npentrel (approved):
Thanks!

PR #5871, opened 2026-09-03
docs(langsmith): document Context Hub sandbox mounts

wbbradley:
## Overview

Context Hub sandbox mounts were documented only on the sandbox SDK page, away from the S3, GCS, and Git mount types, and that section covered the happy path only. This gives them a full section on the sandbox mounts page and adds the behavior a caller needs to plan around:

- Repo handle (`owner/name`, `-` for the current workspace) and the read-access requirement.
- Mount paths outside `/mnt/mounts`, plus a `/memories` row in the mount path table.
- Flattened tree: linked agent and skill repos land at the parent's reference path.
- One-way read-only sync, whole-tree atomic refresh, new commits within roughly 30 seconds as a best-effort cadence rather than a freshness guarantee.
- Latest commit only; pull a commit or environment tag with the Context Hub SDK to read a fixed version.
- `initial_pull_only` / `initialPullOnly`, and when a single pull is the right choice.
- Startup: reads block until the first commit tree arrives instead of seeing an empty directory.
- Failures: retries keep serving the last published tree; a rejected request gives `EIO` (access revoked after creation is re-checked on every pull); a repo over 2500 files or 25 MiB of content serves no tree.
- Minimum versions: `langsmith[sandbox]>=0.11.0` for Python, `langsmith>=0.8.11` for TypeScript.
- Four new `## Limits` bullets, including that restore reconnects the mount but a restored process holding an open file or working directory inside it must reopen.

`sandbox-sdk.mdx` keeps its example and read-only warning; its duplicated parameter prose becomes a pointer to the mounts page, so the two pages cannot drift. The `use-the-context-hub.mdx` next-steps link now targets the mounts page. Both pages keep the `#mount-a-context-hub-repo` anchor.

Behavior is sourced from `smith-go/sandboxes` (`sandboxapi/mounts.go`, `sandboxapi/context_hub_sync.go`, `context_hub_sync.go`, `context_hub_sync_rpc.go`, `docs/sandbox-runtime.md`) and the SDK helpers added in langchain-ai/langsmith-sdk#3351.

## Type of change

## Related issues/PRs

- GitHub issue:
- Feature PR: langchain-ai/langsmith-sdk#3351 (SDK mount helpers)

## Checklist

## Additional notes

Two checklist items are unchecked and want a reviewer's eyes:

- I did not run `docs dev`, so the rendering of the new `<CodeGroup>` blocks and subheadings is unverified. `make lint_prose` passes. `markdownlint` is not installed locally, so that check is left to CI.
- The snippets are not executed. They are adapted from the mount helper signatures and the SDK README examples, and match the surrounding page's client patterns.

One judgment call worth a second opinion: making the mounts page canonical and trimming the SDK page section to a pointer. Happy to instead leave `sandbox-sdk.mdx` untouched and accept the duplication.


wbbradley:
TODO: inform https://langchain.slack.com/archives/C0AS4BVECTY/p1788539937145099?thread_ts=1788461526.428969&cid=C0AS4BVECTY thread when this is merged.

lnhsingh (approved):

PR #5873, opened 2026-09-04
docs(langsmith): document trajectory variable and trajectory online evaluators

fjmorris:
Closes [DOC-1598](https://linear.app/langchain/issue/DOC-1598/document-trajectory-variable-and-trajectory-evaluation-rules-for).

## Why

Trajectory online evaluators score the flat, deduplicated message list for a session, including the tool calls and tool results that the thread message variables leave out. The docs described only `all_messages`, `human_ai_pairs`, and `first_human_last_ai`, so there was nothing covering the `trajectory` variable or how a trajectory evaluator differs from a thread evaluator.

## What changed

**`src/langsmith/prompt-template-format.mdx`**

- Adds `trajectory` as a fourth evaluator variable, with its resolved message shape: `role` of `human`/`ai`/`system`/`tool`, and `content` as either a string or a list of content blocks (`text`, `reasoning`, `tool_call`).
- Adds an "Example with trajectory context" showing a template and the input LangSmith provides.
- States the mutual-exclusion fact at the variable level and points to the evaluator page for how it is enforced, so the rule has one owner rather than two full copies.

**`src/langsmith/online-evaluations-multi-turn.mdx`**

- Adds an "Evaluate trajectories" section: what the judge receives compared to a thread evaluator, creating an evaluator with `{{trajectory}}` in the prompt and mapping, where feedback is written, the evaluator kind being fixed at creation, and the supported and rejected settings.
- Applies a five-persona editorial review of that section, then a verification pass against `langchainplus` `origin/main` at `82916d89fd5`.
- Presents `trajectory` as one of the variables available to a thread evaluator, at the same level as `all_messages`, per the feature owner. "Trajectory evaluator" is internal vocabulary and does not appear on either page.

### Availability risk on the removed cloud-only callout

The section carried a callout saying trajectory evaluators are available on LangSmith Cloud only. **That callout was removed on the product owner's instruction ahead of GA, against the evidence below.** Flagging it as the one change most worth a second opinion, because the page now makes no deployment-availability claim at all.

- Every trajectory evaluator entry in `langchainplus/.changelog` carries `self_hosted: false`, through `2026-09-11-trajectory-retention-all-traces.yaml`.
- `charts/langsmith` in `langchain-ai/helm` ships no trajectory template, no `TRAJECTORY_GRPC_SERVICE_URL`, and no trajectory key in `values.yaml`, at `langsmith-0.17.0-rc.35` and at `main`. Every template file was checked, not just the directory listing.
- `TrajectoryEvaluatorsConfigured()` in `smith-go/config/config.go:2172` is true only when `TRAJECTORY_GRPC_SERVICE_URL` is non-empty, and the env default is empty. A v0.17 install as tagged today therefore cannot run a trajectory evaluator, and BYOC runs the same chart.

Self-hosted support also requires SmithDB to be enabled, per the feature owner, which is a second gate on top of the gRPC service and is not reflected anywhere on the page now that the callout is gone.

The stated plan is that v0.17 carries trajectory evaluators for self-hosted, with BYOC a week later. If that is right, this page wants a version-added admonition ("requires LangSmith Self-hosted v0.17 or later") rather than silence. Please confirm against the chart before treating the page as accurate for self-hosted readers.

Related, and more urgent now that the page implies general availability: `useTrajectoryEvaluatorsEnabled.ts` reads the Eppo flag only and never `InstanceFlags.trajectory_evaluators_supported`, so a self-hosted tenant with the flag on would get the trajectory UI with no data plane behind it.

### Framing: a variable, not an evaluator kind

The page now presents `trajectory` as a thread evaluator variable rather than as its own kind of evaluator, per the feature owner. Worth a reviewer's eye on one point: the implementation does treat it as a distinct rule kind (`RuleKind` includes `trajectory`, `group_by: 'trajectory_id'`, its own applier and validation table), and three consequences of choosing the variable are user-visible, so the page still states them as constraints rather than dropping them:

- The variable requires a tracing project. An evaluator on a dataset cannot use it.
- It cannot be combined with `all_messages`, `human_ai_pairs`, or `first_human_last_ai`, in either direction.
- The choice is fixed at creation. Moving an existing evaluator between `trajectory` and the other thread variables means deleting and recreating it.

If the intent is that readers should not perceive any of those as evaluator-level differences, that is a product change rather than a wording change, and this page should follow it.

### Claims the verification pass corrected

Four statements in the first draft of this PR did not survive a check against the code, and are fixed here:

- **Other evaluators on the same rule.** Only *code* evaluators are excluded. `_UNSUPPORTED_BY_KIND["trajectory"]` (`smith-backend/app/schemas.py:5390`) lists `code_evaluators`, not `evaluators`, and the applier loops `for evaluator in evaluators`. Additional LLM-as-a-judge evaluators are permitted.
- **Grouping.** Not an excluded option. Grouping is how the rule is defined (`group_by: 'trajectory_id'`).
- **What the trajectory is built from.** `llm` and `tool` runs, minus runs marked `ls_message_view_exclude` (`smith-go/runs/v2/messages/messages.go:30`), not model calls alone.
- **The classification discriminant.** The variable mapping (`isTrajectoryVariableMapping`), not the presence of `{{trajectory}}` in the prompt text. A prompt without the mapping errors with "Map the trajectory prompt variable to the trajectory source."

The earlier note in this description claiming the mutual-exclusion guardrail was backend-only is now stale: `getTrajectoryPathDisabledReason` ships the client-side check and the copy "Trajectory variables and non-trajectory variables cannot be used together." The page reflects the shipped behavior.

### Product-side inconsistency worth a bug

`smith-frontend/src/components/EvaluatorCrudPane/utils/trajectoryEvaluator.ts:98` shows "Trajectory evaluators cannot use dataset corrections or be combined with ordinary/code evaluators," but the surrounding condition only tests `code_evaluators?.length`. The copy promises a restriction the code does not enforce. Not a docs change, but it is why this PR does not repeat that claim.

### Naming collision on a page outside this PR

The section was first written as "Evaluate agent trajectories," which mirrored the `sidebarTitle` on `src/langsmith/trajectory-evals.mdx:3`. It is now "Evaluate trajectories," because the docs already define a trajectory as the path an agent took, so "agent" repeated what the noun carries.

Renaming this heading does not resolve the collision, because the nav entry for `trajectory-evals.mdx` still reads "Evaluate agent trajectories" while that page is about the OSS `agentevals` package for offline, dataset-based evaluation, a different feature from the online evaluator variable documented here. Its `title` is already "How to evaluate your agent with trajectory evaluations," so a `sidebarTitle` closer to that, naming `agentevals`, would separate the two in search and in the sidebar. That is a change to a page outside this PR's scope, so I would rather raise it than fold it in silently.

### Still open from the original review

- **Limits.** I did not claim the existing thread-level limits (500 threads per execution, 7-day run age) apply to trajectory evaluators, since trajectory resolution runs through its own paging path. Worth confirming whether they do, and whether trajectory evaluators count against the 10-per-workspace cap.
- **Trace scope.** `_VALID_SCOPE_TYPES` allows `trace` as well as `thread`, but a rule always sends `thread_idle_seconds` (`_resolve_thread_idle_seconds` always returns an int), so the service always resolves thread-scoped trajectories. The page documents threads only. Confirm the trace branch is genuinely unreachable from a rule.
- **Empty or unresolvable trajectories.** The design doc describes a yellow diamond and "Trajectory is not available. Use other variables." Neither string exists on `main`, so the page does not document it. Tracked separately rather than written from the design doc.

## Checks

- `make lint_prose` clean on both files.
- `make broken-links-with-anchors` reports no broken links, so every anchor this PR adds resolves against the built output.
- Behavioral claims traced to `langchainplus` `origin/main` at `82916d89fd5`.

## AI disclosure

Drafted with Claude Code (Opus 5), reviewed by me. Product behavior was verified against the `langchainplus` implementation and the `langchain-ai/helm` charts rather than taken from the design doc.

### DIFF
--- a/src/langsmith/online-evaluations-multi-turn.mdx
+++ b/src/langsmith/online-evaluations-multi-turn.mdx
@@ -80,6 +80,45 @@ Please refer to the [troubleshooting](/langsmith/online-evaluations-multi-turn#t
 
     After saving, your evaluator will appear in the **Evaluators** tab. You can test it once the idle time has passed for any new threads created after saving.
 
+## Evaluate agent trajectories
+
+A trajectory evaluator scores the [trajectory](/langsmith/observability-concepts#trajectories) of a conversation: the flat, ordered list of messages exchanged from start to finish, including tool calls and tool results. Use a trajectory evaluator when the quality you want to measure depends on the path the agent took, not only on what it said to the user.
+
+Trajectory evaluators and the thread evaluators described above differ in what the judge receives:
+
+- **Thread evaluators** receive the conversation LangSmith assembles from the top-level `messages` key of each trace in the thread, through the `all_messages`, `human_ai_pairs`, or `first_human_last_ai` variable. Some of these views leave out tool calls, tool results, and system messages.
+- **Trajectory evaluators** receive the resolved trajectory through the `trajectory` variable. LangSmith builds the trajectory from the model calls inside each trace, so it keeps the tool calls, tool results, and reasoning that produced the conversation.
+
+For the message shape each variable resolves to, see [Thread message variables](/langsmith/prompt-template-format#thread-message-variables).
+
+### Create a trajectory evaluator
+
+To create a trajectory evaluator, follow the [configuration](#configuration) steps above and reference `{{trajectory}}` in your prompt instead of a thread variable. LangSmith saves an evaluator whose prompt contains `{{trajectory}}` as a trajectory evaluator.
+
+A prompt cannot combine `{{trajectory}}` with `{{all_messages}}`, `{{human_ai_pairs}}`, or `{{first_human_last_ai}}`. LangSmith rejects an evaluator that mixes the trajectory variable with thread variables.
+
+A trajectory evaluator runs against a tracing project, and it is the only evaluator on its rule. When the evaluator completes, LangSmith writes feedback to the most recent root run in the thread, under the feedback key you configured, and associates that feedback with the thread.
+
+<Warning>
+The kind of an evaluator is fixed when you create it. To turn a trajectory evaluator into a thread evaluator, or the reverse, delete the evaluator and create a new one.
+</Warning>
+
+### Trajectory evaluator settings
+
+A trajectory evaluator supports the following settings:
+
+- **Filters**: Limit which threads the evaluator runs on.
+- **Sampling rate**: Evaluate a percentage of the threads that match your filters.
+- **Idle time**: The project-level thread idle time, which applies to every thread-level evaluator in the project.
+
+The following options are not available to a trajectory evaluator, and LangSmith rejects a configuration that includes them:
+
+- Datasets and experiments as a source.
+- Backfill over existing threads.
+- Grouping, trace filters, and tree filters.
+- Few-shot examples.
+- Additional LLM-as-a-judge or code evaluators on the same rule, and non-evaluator actions such as adding threads to an annotation queue.
+
 ## Limits
 
 These are the current limits for thread-level processing (subject to change). They apply to multi-turn online evaluators and to [automation rules](/langsmith/rules#set-the-item-type-to-runs-or-threads) whose item type is **Threads**. Reach out if you are running into any of these limits.
--- a/src/langsmith/prompt-template-format.mdx
+++ b/src/langsmith/prompt-template-format.mdx
@@ -480,7 +480,7 @@ Evaluators need to analyze conversations holistically—looking at patterns acro
 
 ### Thread message variables
 
-LangSmith provides three pre-structured views of conversation [threads](/langsmith/evaluation-concepts#threads):
+LangSmith provides four pre-structured views of a conversation: three over a [thread](/langsmith/evaluation-concepts#threads) and one over a [trajectory](/langsmith/observability-concepts#trajectories).
 
 ```mustache
 {{!-- Access all messages in the thread --}}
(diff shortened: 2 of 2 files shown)


fjmorris:
### Follow-up: UI guardrails for trajectory evaluators are shipped

> **Updated 2026-09-18.** This comment originally said the client-side guardrails were not shipped and were waiting on langchainplus#37972-#37974. That was true on Sep 8 and is wrong now. The guardrails landed through different PRs. The original text is replaced rather than appended, so the corrected state is below.

Rechecked against `langchainplus` `origin/main` (`d4bb93e24b8`, Sep 18). Previously checked at `bd22472222` (Sep 8) and `82916d89fd5`.

**What shipped, and through which PRs:**

- **#39285** (Sep 12, LSO-4262) made `group_by: 'trajectory_id'` the only discriminant on the rule, and sends `evaluators` like any other rule.
- **#39688** (Sep 15) added `getTrajectoryPathDisabledReason`, which disables the conflicting options in the variable source dropdowns.
- **#40719** (Sep 17) settled the copy for the conflict case.

`utils/trajectoryEvaluator.ts` on `main` now carries the mutual-exclusion check in both directions, plus `getTrajectoryAuthoringState`, which returns the save-time and preview-time errors: "Map the trajectory prompt variable to the trajectory source.", "An existing evaluator cannot be converted to a trajectory evaluator.", "Trajectory evaluators are not enabled for this workspace.", and "Trajectory evaluators require a tracing project." The disabled-path reason is wired into the dropdowns through `useVariableMappingProps`, and the authoring state through `PreviewPanel` and `useCommitAndSaveRule`.

So a mixed prompt is no longer left to the backend. The client blocks the second variable at selection time, with the tooltip "Trajectory variables and non-trajectory variables cannot be used together." The backend check (`_validate_trajectory_variable_mapping`, `smith-backend/app/schemas.py:5580`) is still there as the save-time backstop, so the rejection the docs describe holds either way.

**Two things from the Sep 8 comment that are still accurate:**

- **langchainplus#37972, #37973, and #37974 are still open**, now as drafts, last touched Sep 18. They describe the same three behaviors that #39688 and #40719 already landed, so they look superseded rather than pending, but I have not confirmed that with the owner and have not closed anything. Worth a glance from whoever owns them, since leaving them open reads as missing work that in fact exists.
- **The exact tooltip wording from LSO-3852** ("Trajectory evaluators can't be converted from thread evaluators. Create a new evaluator instead.") still returns zero hits. The shipped equivalent is "An existing evaluator cannot be converted to a trajectory evaluator." The behavior the issue asked for exists; only the string differs.

**What this changed in the PR:** the optional follow-up this comment floated is done, not deferred. The page documents the dropdown behavior directly, in what is now **Use the trajectory variable** under **Evaluate trajectories**: once you map the trajectory source the other sources are disabled, once you map one of the others the trajectory source is disabled, the dropdown explains why, and saving a mixed evaluator fails the same way. Nothing here is blocking.

Two notes on vocabulary and scope, so this comment is not read against the current diff:

- "Trajectory evaluator" in this comment is the internal and code-level name. The docs deliberately do not use it. Per the feature owner, `trajectory` is presented as a variable available to a thread evaluator, at the same level as `all_messages`.
- `trajectoryEvaluator.ts:98` still shows "Trajectory evaluators cannot use dataset corrections or be combined with ordinary/code evaluators," while the guard it sits behind only tests `code_evaluators?.length`. The copy promises more than the code enforces. Carried in the PR description as a product-side bug, and the reason the page does not repeat that claim.

Winston-Huynh-LangChain (approved):

PR #5899, opened 2026-09-07
docs(langgraph): document which checkpoints exist before the first node runs per durability mode

impartshadow:
## Overview

Adds a short subsection, **What exists before the first node runs**, under *Durability modes* in `src/oss/langgraph/checkpointers.mdx`. It states which checkpoints are committed before the first node executes in each mode, that a crash during the first checkpoint write leaves the thread indistinguishable from one that was never invoked, and that callers who need to tell those apart must record admission outside the checkpointer.

Verified against `langgraph==1.2.11` / `langgraph-checkpoint-sqlite==3.1.1` by recording `SqliteSaver.put()` calls and counting `checkpoints` rows from inside the first node:

| durability | committed before first node runs |
|---|---|
| `"sync"` | input (step -1) and step 0, deterministically (200/200 runs) |
| `"async"` | not guaranteed: 0 rows in 90/200 runs, input only in 48, input + step 0 in 13, remainder mid-commit |
| `"exit"` | none (200/200 runs) |

The `"async"` row is the one that is easy to get wrong; the write is submitted before the node runs but nothing waits on it (`Pregel.stream` only calls `loop._put_checkpoint_fut.result()` when `durability == "sync"`).

## Type of change

## Related issues/PRs

- GitHub issue: See langchain-ai/langgraph#8764 for context (reporter agreed on the narrowed scope of "document the pre-first-node checkpoint guarantee and the admission limitation")
- Feature PR: none; no runtime change proposed

## Checklist
- [ ] I have tested my changes locally using `docs dev` (prose linted with the pinned Vale 3.17.1: 0 errors)
- [x] All code examples have been tested and work correctly (no new code examples; the Python/JS names `get_tuple`/`getTuple`, `EmptyInputError` were checked against both repos)
- [x] I have used **root relative** paths for internal links (no new links)
- [ ] I have updated navigation in `src/docs.json` if needed (not needed; subsection only)

## Additional notes

The Python and JS variants differ only in `None`/`undefined` and `invoke(None, …)`/`invoke(null, …)`, so they are split with `:::python` / `:::js` fences.

### DIFF
--- a/src/oss/langgraph/checkpointers.mdx
+++ b/src/oss/langgraph/checkpointers.mdx
@@ -602,6 +602,26 @@ The durability modes, from least to most durable, are as follows:
 * `"async"`: LangGraph persists changes asynchronously while the next step executes. This provides good performance and durability, but there is a small risk that LangGraph does not write checkpoints if the process crashes during execution.
 * `"sync"`: LangGraph persists changes synchronously before the next step starts. This ensures that LangGraph writes every checkpoint before continuing execution, providing high durability at the cost of some performance overhead.
 
+### What exists before the first node runs
+
+The durability mode also determines which checkpoints are committed before the first node executes:
+
+* `"exit"`: No checkpoint exists until the run ends. A crash during the run leaves no record of the thread.
+* `"async"`: LangGraph starts writing the input checkpoint before the first node executes, but does not wait for the write to finish. The first node can run before any checkpoint is committed.
+* `"sync"`: LangGraph commits the input checkpoint (step `-1`) and the step `0` checkpoint before the first node executes.
+
+In every mode, a crash during the first checkpoint write leaves the thread indistinguishable from a thread that was never invoked:
+
+:::python
+`get_tuple()` returns `None` in both cases, and `invoke(None, config)` raises `EmptyInputError`.
+:::
+
+:::js
+`getTuple()` returns `undefined` in both cases, and `invoke(null, config)` raises `EmptyInputError`.
+:::
+
+If you need to distinguish "never accepted" from "accepted but lost", record admission outside the checkpointer, for example by committing an outbox row to your own store before you call `invoke`.
+
 ## Optimize checkpoint storage
 
 :::python
--- a/src/oss/langgraph/checkpointers.mdx
+++ b/src/oss/langgraph/checkpointers.mdx
@@ -604,13 +604,13 @@ The durability modes, from least to most durable, are as follows:
 
 ### What exists before the first node runs
 
-The durability mode also determines which checkpoints are committed before the first node executes:
+For a new thread, the durability mode also determines which checkpoints are committed before the first node executes:
 
 * `"exit"`: No checkpoint exists until the run ends. A crash during the run leaves no record of the thread.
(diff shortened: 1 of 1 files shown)


mstevens843:
Thanks for correcting the `async` row and moving the documentation half forward.

I reviewed the proposed subsection. For Python, it matches the boundary I intended in #8764: `sync` waits for the pre-node checkpoints, `async` starts persistence without guaranteeing that anything is committed before the node runs, and `exit` leaves no checkpoint until completion.

The admission wording also looks right. Checkpoint absence alone cannot distinguish “never accepted” from “accepted but runtime evidence missing,” so callers needing that distinction require an authoritative admission record outside the checkpointer.

I have not independently measured the JavaScript implementation, so my confirmation here is limited to the Python behavior.

impartshadow:
@mstevens843 I tested the JavaScript half on `@langchain/langgraph@1.4.14` and pushed a scope correction in `5d314e9`.

Using a `MemorySaver` subclass that delays each `put()` by 10 ms and records completed writes, 30 runs per mode observed the following inside the first node:

| Mode | Completed checkpoint steps before node | Trials |
| --- | --- | --- |
| `sync` | `-1`, `0` | 30/30 |
| `async` | none with the deliberately delayed saver | 30/30 |
| `exit` | none | 30/30 |

An empty saver returned `undefined` from `getTuple()`, and `invoke(null, config)` raised `EmptyInputError`. These are API-ordering checks, not a disk-crash durability test. The delayed `async` arm demonstrates that zero completed writes is possible; it does not claim every saver always behaves that way.

A separate existing-thread control retained its earlier checkpoint during an `exit` invocation. I therefore narrowed the subsection to **a new thread**, and changed “a crash during the first checkpoint write” to **“a crash before the first checkpoint becomes durable”**. A process can die after a storage commit but before `put()` returns, so the original wording was too broad.

The revised file passes the repository's Vale configuration with zero errors or warnings. This is a Shadow-authored test and correction; the external Python verdict remains yours, and the JavaScript measurement is mine.

mstevens843:
@impartshadow Thanks for measuring the JavaScript path and tightening the scope.

The revised wording matches the boundary I intended. Limiting it to a new thread avoids implying that an `exit` invocation removes an existing checkpoint, and “before the first checkpoint becomes durable” is more precise than “during the first checkpoint write,” since a storage commit may succeed before `put()` returns.

The JavaScript caveat also reads correctly: the delayed `MemorySaver` establishes that zero completed writes before the first node is possible under `async`, without turning that observation into a universal saver or disk-durability claim.

I made the same wording correction in crashpoint’s #8764 experiment notes. No further scope changes from my side.

impartshadow:
@npentrel @lnhsingh Requesting review per the bot note above. This is ready:

- One file, `src/oss/langgraph/checkpointers.mdx`, +20 lines: a *What exists before the first node runs* subsection under *Durability modes*, scoped to a new thread.
- Python behavior confirmed by the #8764 reporter (@mstevens843) for `sync` / `async` / `exit`; JavaScript ordering measured on `@langchain/langgraph@1.4.14` (table in the comment above). No further scope changes from either side.
- All 20 checks green on `5d314e9`, including Vale and the docs build/link check.

Happy to adjust wording or placement.

PR #5919, opened 2026-09-08
docs: add opsen as a deep agents sandbox backend

md322613:
Adds sandbox and provider pages for opsen, following the existing Daytona pages.

`langchain-opsen` is published and installable: https://pypi.org/project/langchain-opsen/

**What it is.** opsen runs the sandbox and proxies the agent's model calls through the same session, so machine, model and tool spend are attributed to one task id — `backend.cost()` returns the three together. `budget_usd` is a ceiling enforced during the run, so an agent that would cross it is refused mid-flight rather than found on an invoice.

**Interface.** `OpsenSandbox` subclasses `BaseSandbox` and implements `id`, `execute`, `upload_files` and `download_files`, per the contributing guide. The filesystem tools and async variants are inherited.

**Standard suite: 79/86.** Running `SandboxIntegrationTests` found three real bugs, all now fixed: absolute paths were being made relative on upload; large writes silently failed by exceeding the shell argument limit; and errors were raw shell text rather than the normalised codes (`file_not_found`, `invalid_path`, `is_directory`, `permission_denied`).

The remaining 7 appear to be a disagreement between deepagents 0.7.13 and langchain-tests 1.1.9, and are all in inherited methods, so no provider implementation affects them:

- **6 glob tests.** `_parse_glob_output` calls `_absolutize_glob_path` and its docstring states "GlobResult with absolute paths". The suite asserts `"file1.txt" in paths`.
- **`test_write_existing_file_fails`.** `BaseSandbox.write` documents "creating or overwriting it if it already exists". The suite asserts `result.error is not None` when the file exists.

Flagging rather than working around, since if the intended contract changed then the base class and the suite disagree for every provider. Happy to adjust if I have misread it.

**Two behaviours the page calls out**, because they differ from the other backends:

- There is no provider SDK object to construct first. The session starts on the first command, so an agent built and never run starts no sandbox.
- Port exposure depends on the runtime underneath and raises with a message naming the limitation where unavailable, rather than returning a URL that does not answer.

Happy to adjust to house style — I followed `sandboxes/daytona.mdx` and `providers/daytona.mdx`.

### DIFF
--- a/src/oss/python/integrations/providers/opsen.mdx
+++ b/src/oss/python/integrations/providers/opsen.mdx
@@ -1,0 +1,13 @@
+---
+title: "opsen integrations"
+sidebarTitle: "opsen"
+description: "Integrate with opsen using LangChain Python."
+---
+
+[opsen](https://opsen.dev) runs agent sandboxes and proxies their model calls, so machine, model and tool spend land under one task id. See the [opsen docs](https://opsen.dev/docs) for signup, authentication, and platform details.
+
+<Columns cols={2}>
+    <Card title="OpsenSandbox" href="/oss/integrations/sandboxes/opsen" cta="Get started" icon="terminal" arrow>
+        opsen sandbox backend for deepagents.
+    </Card>
+</Columns>
--- a/src/oss/python/integrations/sandboxes/opsen.mdx
+++ b/src/oss/python/integrations/sandboxes/opsen.mdx
@@ -1,0 +1,91 @@
+---
+title: OpsenSandbox integration
+description: Integrate with the OpsenSandbox sandbox backend using LangChain Python.
+integration:
+  name: OpsenSandbox
+  pypi: langchain-opsen
+---
+
+[opsen](https://opsen.dev) runs agent sandboxes and proxies their model calls, so machine, model and tool spend land under one task id. See the [opsen docs](https://opsen.dev/docs) for signup, authentication, and platform details.
+
+## Installation
+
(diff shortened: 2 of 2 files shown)


md322613:
@mdrxy this is ready for review. Package is published (`langchain-opsen` 0.1.4) and passes 79/86 of the sandbox standard suite — the 7 failures are noted in the description as an apparent mismatch between deepagents 0.7.13 and langchain-tests 1.1.9, both in inherited methods.

npentrel:
Hi @md322613 - we no longer list external docs for integrations in our docs but instead link out to your docs directly. I've made a separate PR to add your integration to our docs: https://github.com/langchain-ai/docs/pull/5927.

Thanks for reporting the issues with the tests, I'll look into those

md322613:
Thanks @npentrel — #5927 is a better shape than what I opened.

On the standard-suite failures, in case it saves digging: both are in inherited methods, so no provider implementation affects them.

**glob (6 tests):** deepagents 0.7.13 `_parse_glob_output` calls `_absolutize_glob_path` and documents "GlobResult with absolute paths"; langchain-tests 1.1.9 asserts `"file1.txt" in paths`.

**test_write_existing_file_fails:** `BaseSandbox.write` documents "creating or overwriting it if it already exists"; the suite asserts an error when the file exists.

Versions: deepagents 0.7.13, langchain-tests 1.1.9, langchain-core 1.6.2.

PR #5929, opened 2026-09-08
update MDA docs pattern

npentrel:
These changes make the MDA docs more guided and align the structure of each page.

PR #5931, opened 2026-09-08
Add MDA MCP endpoint page

ndrezn:
Managed Deep Agents docs covered MCP connectors (tools *into* an agent) but not the `/mcp` endpoint every deployment serves, so the recurring "can I talk to my MDA from Claude?" question had no page. Adds `managed-deep-agents-mcp-endpoint.mdx` under Build and deploy in both language dropdowns, and repoints the stale `/langsmith/managed-deep-agents-mcp` redirect (it pointed at channels).

Content comes from `server-api-ref.mdx`, `server-mcp.mdx`, `identity.mdx`, and `cloud.mdx`. Left out rather than guessed: the MCP tool description source, `disable_mcp`, and on-behalf-of OAuth for connectors.

The auth section is verified against the runtime and the platform source, not against existing docs:

- A missing header returns 401 `missing x-api-key`. An invalid, revoked, or expired key returns 403 `API key is forbidden`, since LangSmith's API-key middleware has no 401 path (`smith-go/auth/api_key.go`). A valid key from another workspace returns 403 `API key tenant mismatch`.
- `mda dev` serves `/mcp`, and every other local route, with no credential. That is deliberate, not a dev-server artifact: `mda dev` sets `MDA_LOCAL_DEV=1`, the platform-injected `LANGSMITH_AUTH_ENDPOINT` and `LANGSMITH_TENANT_ID` are absent locally, and `_resolve_langsmith_api_key_user_sync` then returns a local service principal ahead of the 401. Deploy never sets that env. The page now states this, so the earlier open question is resolved and no reviewer follow-up is needed.

Test plan: `make lint_prose` and `make broken-links` clean; the Python and TypeScript `MultiServerMCPClient` snippets run as written against both a local `mda dev` server and a live deployment.

Written with Claude Code.


victorm-lc:
should we just add this content to: langsmith/server-mcp for agent server docs and have a nice crosslink to agent server docs from MDA sidebar so that folks know? I think we should emphasize the power of agent server in MDA more. The current docs don't tell a good story there (my fault)

src/docs.json
```diff
@@ -2710,7 +2712,7 @@
     },
     {
       "source": "/langsmith/managed-deep-agents-mcp",
-      "destination": "/langsmith/python/managed-deep-agents-channels"
+      "destination": "/langsmith/python/managed-deep-agents-mcp-endpoint"
```
npentrel, line 2718: this shouldn't be changed. no page url is changing, the forwarding should stay as is

src/docs.json
```diff
@@ -2752,6 +2754,10 @@
       "source": "/langsmith/managed-deep-agents-mcp-connectors",
       "destination": "/langsmith/python/managed-deep-agents-mcp-connectors"
     },
+    {
+      "source": "/langsmith/managed-deep-agents-mcp-endpoint",
+      "destination": "/langsmith/python/managed-deep-agents-mcp-endpoint"
+    },
```
npentrel, line 2767: this is not needed

src/langsmith/managed-deep-agents-mcp-endpoint.mdx
```diff
@@ -0,0 +1,169 @@
...
+```
+:::
+
+Read credentials from the environment. Do not hard-code them in client code.
+
+## Understand session behavior
+
+Each `/mcp` request is stateless and independent. The endpoint does not support sessions, so separate calls do not share thread state. To retain knowledge across calls, enable [durable memory](/langsmith/managed-deep-agents-memory).
+
+## Distinguish the endpoint from other capabilities
```
npentrel, line 148: this will need to change to align with https://github.com/langchain-ai/docs/pull/5929

npentrel (changes requested):
Let's merge https://github.com/langchain-ai/docs/pull/5929 first please

npentrel (approved):

ndrezn:
@npentrel thoughts on Vic's comment?

npentrel:
> Naomi Pentrel (@npentrel) thoughts on Vic's comment?

I had missed that, let's keep a lightweight page here and then point at the agent server docs. I can take a stab at that

Have you spoken to an engineer yet about the unauthenticated requests on dev?

npentrel:
@ndrezn  I think https://github.com/langchain-ai/docs/pull/5931/changes/1c87904478cc07fd094d17a35e07f17d728e1532 is all we need to do to link more to Agent Server.

PR #5934, opened 2026-09-08
Add Speechify TTS client to the voice-agent guide

luke-speechify:
Adds a "Speechify Client" accordion to the voice-agent guide's Text-to-speech section, alongside the existing Cartesia one, and extends the "Streaming TTS" concept to cover HTTP-streaming providers.

Speechify streams over HTTP chunked transfer rather than a WebSocket: the whole turn is sent in one streaming request and audio streams back in order. The accordion shows the Python and TypeScript client, matching the Cartesia example.

Also notes the optional `/v1/audio/stream/with-timestamps` (SSE) endpoint, which returns the same audio plus word-level speech marks for caption/highlight sync.

Pairs with the reference-app adapter PR: langchain-ai/voice-sandwich-demo#57

### DIFF
--- a/src/oss/langchain/voice-agent.mdx
+++ b/src/oss/langchain/voice-agent.mdx
@@ -491,7 +491,7 @@ The TTS stage synthesizes agent response text into audio and streams it back to 
 - **Upstream processing**: Passes through all events and sends agent text chunks to the TTS provider
 - **Audio reception**: Receives synthesized audio chunks from the TTS provider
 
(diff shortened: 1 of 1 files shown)


luke-speechify:
@npentrel and @lnhsingh this is ready for review, thanks!

npentrel:
I'm sorry, this is not an addition we wish to make here.

luke-speechify:
> I'm sorry, this is not an addition we wish to make here.

Any feedback on why not?

PR #5970, opened 2026-09-10
chore: add shared docs-authoring skills in .agents/skills

lnhsingh:
## Overview

Adds one tracked, agent-neutral home for this repo's authoring skills, plus the first two skills.

**Why.** Skills already exist here, split across two homes with different fates. `.deepagents/skills/` is tracked but read only by Deep Agents Code. `.claude/skills/` is gitignored (`.gitignore:203`), so anything written there is personal tooling that no teammate can use. Nothing shared existed for the tasks people repeat most.

Separately, `AGENTS.md` and `CLAUDE.md` are always-on context: ~550 lines loaded on every task, including a ~200-line navigation map most sessions never need. Skills are the place for procedures only some tasks need, which keeps them out of that budget.

**Why `.agents/skills/`.** It is the project path already read by Cursor, Codex, GitHub Copilot, Gemini CLI, OpenCode, Deep Agents, Droid, Kilo Code, and roughly fifteen other agents, so those need no install step at all. Claude Code reads `.claude/skills/` only, so `make skills` symlinks each skill in. The target is idempotent, skips entries that are real directories (so personal skills are untouched), and prunes links whose source is gone.

The two skills link into `AGENTS.md` rather than restating it. The guidelines already fan out into four derived files kept in sync by hand; a skill that copies the style guide would become a fifth copy that drifts.

### What is here

| File | |
|---|---|
| `.agents/skills/add-docs-page/SKILL.md` | Directory choice, frontmatter, `src/docs.json` nesting (including the two-dropdown rule for language-versioned OSS pages), `docs mv` for renames, redirects, verification |
| `.agents/skills/docs-review/SKILL.md` | Reviewing changed prose against Vale and the style guide, reporting the rule each finding breaks. Promoted from a personal slash command, plus a working-tree target mode so it can run straight after an authoring pass |
| `.agents/skills/document-tooling-in-notion/SKILL.md` | Which of the five internal Notion tooling pages owns a topic, what stays in the repo, and the known Notion MCP failure modes |
| `.agents/skills/README.md` | Install, the `AGENTS.md` boundary, how to add a skill |
| `Makefile` | `make skills` target and help line |
| `AGENTS.md` / `CLAUDE.md` | Skills section and a quick-reference row, byte-identical |
| `.cursorrules`, `.github/copilot-instructions.md` | Mirrored quick-reference row |
| `src/oss/contributing/documentation.mdx` | Contributor install line |
| `tests/unit_tests/test_skills.py` | Structural tests for the skills tree, running in the existing suite |
| `.agents/skills/{docs-code-samples,submit-integration,update-integrations-prs}/SKILL.md` | Moved from `.deepagents/skills/`, so there is one skills tree instead of two |

### Chaining

`add-docs-page` invokes `docs-review` at its verification step, on the files that pass touched, after the edit is complete and before the commit. That required a fourth target mode in `docs-review`: the three existing modes all resolve a committed ref, which cannot express "the files I just edited."

Both skills state the constraint that makes this useful rather than noisy: review finished edits, not drafts. Findings on a half-written section go stale as soon as writing resumes.

### Consolidating the Deep Agents skills

The three skills in `.deepagents/skills/` moved into `.agents/skills/`, and **no symlink or shim is needed**. Deep Agents Code reads `.agents/skills/` as a project skill directory at *higher* precedence than `.deepagents/skills/`, per the discovery order this repo publishes at [Memory and skills](https://docs.langchain.com/oss/deepagents/code/memory-and-skills#skill-discovery). `langchain-ai/deepagents` keeps its own skills at `.agents/skills/` too. So `.github/workflows/integration-submission.yml` still resolves `skill: submit-integration`.

This is the change that matters most in the PR: integration work is the largest recurring workload here (82 of the last 200 merged PRs carry the `integration` label), and the two skills covering it were readable only by Deep Agents Code.

### Testing

`tests/unit_tests/test_skills.py` runs in the existing suite, so no new CI dependency. Per skill it checks frontmatter, that the name matches the directory, that every frontmatter key is recognized (catching `allowed_tools` for `allowed-tools`), that every repository path the skill names exists, and that every make target it tells an agent to run is defined. It also checks the README and `AGENTS.md` tables list exactly the skills in the tree.

The two checks that matter most were mutation-tested: renaming a referenced script and renaming a make target each fail the suite. A skill that points at something renamed is worse than no skill, because the agent follows it confidently.

It earned its keep immediately. Pulling `docs-code-samples` into the tree surfaced two dead references in it: `make code-snippets-langsmith` and `CODE_SNIPPET_LANGSMITH_SOURCES`, neither of which exists any more. Both are replaced with the `CODE_SNIPPET_SOURCES` environment variable against `make code-snippets`, and the same dead target is fixed in the `scripts/extract_code_snippets.py` docstring.

Behavior evals (does the skill change what an agent produces) are still deliberately out of scope.

## Type of change

**Type:** Other (repository tooling, plus one contributing-docs line)

## Related issues/PRs

- GitHub issue:
- Feature PR:
- Linear issue:
- Slack thread:

## Please look closely at

- **`make skills` behavior.** `claude plugin validate` refuses a symlinked component directory, and warns on symlinked entries inside a real one with the note "A session loading this directory does follow them." That is why the target links per skill rather than linking the parent. Verified working: both skills loaded in a live session immediately after running it.
- **Whether `npx skills add` should be recommended anywhere.** It is documented in the README as the escape hatch for other agents, with a caveat: tested against a local source it *copies* rather than symlinks, and `npx skills update` then reports "No project skills to update" and leaves the copy stale.
- **Accuracy of the new skills.** `add-docs-page` encodes the `src/docs.json` structure and the redirect rules enforced by `scripts/check_removed_pages_redirects.py`. `document-tooling-in-notion` encodes the one-home routing rule across the five Notion pages.

## Not in this PR

- `openwiki/workflows/integration-listing-automation.md` still cites `repo://.deepagents/skills/submit-integration/SKILL.md`. It is generated, so it is left for the OpenWiki workflow to refresh rather than hand-edited.
- No behavior evals. Structural tests are in; skill-creator evals and LangSmith invocation telemetry are designed but deferred pending a decision on which are worth the cost at this size.
- `docs-style` is deliberately not a skill yet. Claude Code supports a `paths:` frontmatter field that would scope the style guide to `src/**/*.mdx` the way Cursor and Copilot already scope it, but whether other agents honor `paths:` in a SKILL.md is unverified, so it stays in `AGENTS.md` for now.

## Checklist

- [x] Verified locally: `claude plugin validate .agents/skills --strict` passes, `AGENTS.md` and `CLAUDE.md` diff clean, `make lint_prose` clean on the changed MDX, `make skills` tested for idempotency, personal-skill coexistence, and stale-link pruning
- [x] I have used root relative paths for internal links
- [x] No `src/docs.json` navigation change needed (no new page)

## Additional notes

Written with Claude Code. Every claim about agent skill-directory behavior in the skills and README was measured against the tooling rather than assumed.


.agents/skills/add-docs-page/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+   `src/langsmith/sandboxes.mdx` is `"langsmith/sandboxes"`.
+3. **A new group leads with an index page:**
+   `"pages": ["group/index", "group/page"]`.
+
+Integration pages are the exception. Add them to the component's `index.mdx`
+instead, and touch `docs.json` only when creating a brand-new component group.
+
+## Step 4. Add redirects for a move, rename, or deletion
+
+For a move or rename, run the repo's mover first. It rewrites cross-references
```
npentrel: should add caveat that it's only changes from main to new commit that matter. moving files before they hit main in separate commits does not matter and should not get redirects

.agents/skills/add-docs-page/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+typo in a page path is caught there rather than at build time.
+
+## Step 5. Verify
+
+Run all three, in this order:
+
+```bash
+make lint_prose FILES="src/path/to/page.mdx"
+make build
+make broken-links
```
npentrel:
```suggestion
make broken-links-with-anchors
```

.agents/skills/add-docs-page/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Read `make broken-links` output by skipping to the `⎿` lines. Those are the only
```
npentrel:
```suggestion
Read `make broken-links-with-anchors` output by skipping to the `⎿` lines. Those are the only
```

.agents/skills/add-docs-page/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Checklist
+
+- [ ] File in the directory the source-directory table names, not `build/`.
+- [ ] Frontmatter present, `description` free of markdown.
+- [ ] `src/docs.json` entry in the right product, menu item, tab, and group.
+- [ ] Both language entries added if the page is language-versioned.
+- [ ] Index page first if a new group was created.
+- [ ] Redirect added for every moved, renamed, or deleted path.
+- [ ] `make lint_prose` clean, `make broken-links` shows no new `⎿` lines.
```
npentrel:
```suggestion
- [ ] `make lint_prose` clean, `make broken-links-with-anchors` shows no new `⎿` lines.
```

npentrel (approved):
few small things

.agents/skills/add-docs-page/SKILL.md
lnhsingh:
```suggestion
For a move or rename of a file that was already on `main`, run the repo's mover first. It rewrites cross-references
```

PR #6003, opened 2026-09-11
docs(langsmith): document the SSO login link redirect_to param

AloofBuddha:
Documents the optional `redirect_to` path on the shareable SSO login link, added in langchain-ai/langchainplus#38421 (ENT-697).

The example is percent-encoded on purpose. Because the only author of this URL is a human or a script, an unencoded value silently loses part of the destination: `URLSearchParams` splits on `&`, and a `#` never reaches the app. This is an understood and acceptable risk - its on the creator of the link to make sure its encoded properly but docs should call it out. 

| link | resulting redirect |
| --- | --- |
| `?redirect_to=/o/ORG/projects/p/P/r/RUN` | correct |
| `?redirect_to=/o/ORG/projects/p/P?peek=RUN&tab=traces` | truncated to `/o/ORG/projects/p/P?peek=RUN` |
| `?redirect_to=/o/ORG/prompts#tab-2` | truncated to `/o/ORG/prompts` |
| `?redirect_to=%2Fo%2FORG%2Fprojects%2Fp%2FP%3Fpeek%3DRUN%26tab%3Dtraces` | correct, full path preserved |

The page also states when *not* to use the parameter: a plain LangSmith URL already returns the user to the page after SSO → work email, so this is for links that must start SAML sign-in themselves.

Closes ENT-1638 (the `docs_link` changelog backfill ships as its own langchainplus PR once this merges).

## Checks

- `make lint_prose` (Vale): 0 errors, 0 warnings, 0 suggestions.
- `markdownlint` is not installed locally; deferred to CI.
- No `src/docs.json` change: this adds no new page.


src/langsmith/user-management.mdx
```diff
@@ -384,6 +384,18 @@ Have a user with access sign in via the unique login URL from the `SSO Configura
...
+
+```text
+https://smith.langchain.com/sso/login/<slug>?redirect_to=%2Fo%2F<organization-id>%2Fprojects%2Fp%2F<project-id>%2Fr%2F<run-id>
+```
+
+The value must decode to a path on the same host, starting with `/`. In the example above it decodes to `/o/<organization-id>/projects/p/<project-id>/r/<run-id>`, so the user lands on that run. A value pointing at another host is rejected, and the user lands on the LangSmith home page instead.
+
+Encode the path. An unencoded path that contains its own query string or fragment is truncated at the first `&` or `#`, which sends the user to an incomplete destination.
+
+Most links do not need this parameter. Opening a LangSmith URL directly, then signing in through **SSO** with a work email, already returns the user to that page. Use `redirect_to` when the link must start SAML sign-in itself, such as a link generated by a script or posted for a team that signs in through a single identity provider.
```
lnhsingh, line 398:
```suggestion

<Note>
Most links do not need this parameter. Opening a LangSmith URL directly, then signing in through **SSO** with a work email, already returns the user to that page. Use `redirect_to` when the link must start SAML sign-in itself, such as a link generated by a script or posted for a team that signs in through a single identity provider.
</Note>
```

lnhsingh (approved):
small nit

PR #6024, opened 2026-09-14
add context hub page for mda

npentrel:


src/langsmith/managed-deep-agents-context-hub.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+For ownership of managed runtime fields, see [Agent definition](/langsmith/managed-deep-agents-agent-definition).
+
+## How deploy syncs context
+
+`mda deploy` syncs deploy-owned context as part of the deploy pipeline:
+
+- **Instructions and skills**: Each deploy updates the agent's Context Hub repo from the project files. Later deploys sync the project copies again. For skills, deploys also remove deployed skill files that no longer exist locally.
+- **Durable memory**: When memory is enabled, the deployment mounts one Context Hub tree at `/memories/agent/`. Deploys do not overwrite content already stored there.
+- **Conflicts**: If the Context Hub repo changes while a deploy is syncing, the CLI reports a conflict. Re-run `mda deploy`.
```
ndrezn, line 31: This is maybe the most important nuance to document and maybe even deserves its own header (`## Source of truth` or something?)

In the case of conflict, the CLI will show:
```
◇  Checked Context Hub context
│
▲  Context Hub instructions.md changed since the last MDA sync.
│
◆  Overwrite Hub-edited instructions.md with the local version?
│  ○ Yes / ● No
└
```
In a non-interactive mode, a flag is required to enable deploys in the case of conflict:
```
┌  v0.4.7
│
●  Project: /Users/you/support-agent
│
└  Context Hub instructions.md and skills/ changed since the last MDA sync. Re-run with --context-strategy overwrite or --context-strategy keep-hub.
```
We should add a note on `--context-strategy` here.

Something like:
```
- **Conflicts**: If `instructions.md` or `skills/` changed in Context Hub since the last
  `mda deploy`, the CLI stops before syncing and asks whether to overwrite the Context Hub
  copy with the project copy. The default answer keeps the Context Hub version and skips
  syncing that section. In a non-interactive shell there is no prompt — the deploy exits
  with an error, and you must re-run with `--context-strategy overwrite` or
  `--context-strategy keep-hub`. Re-running `mda deploy` on its own reports the same
  conflict again.
- **Concurrent edits**: If the repo changes during the sync itself, the deploy fails with
  a commit conflict. Re-run `mda deploy`.
```

is the minimum change but again we could probably be more complete! Because there are two possible sources of truth this has been confusing for people.

npentrel: I think the prompt in the CLI is so clear that we can keep it this minimal - I've added it

PR #6036, opened 2026-09-14
[MDA 0.8] Document agent and user memory layers for MDA

ndrezn:
MDA 0.8 splits durable memory into independent `agent` and `user` layers, each with an optional per-run `allow(context)` policy. Rewrites the memory page, and fixes the tutorial snippet, which used the string `scope` form that 0.8 rejects.

Sourced from the SDK on `caspar/0.8`.

**Draft:** 0.8.0 is unreleased. Merge once 0.8.0 is live.


src/langsmith/managed-deep-agents-memory.mdx
```diff
@@ -65,70 +96,196 @@ The agent decides what to remember based on prompting. To make the policy explic
...
+```
+:::
+
+:::js
+```ts
+const context = { provider: "slack", event: { /* raw Slack event */ } };
+```
+:::
+
+The default user-memory policy reads `event.channel_type`, where `"im"` means a one-to-one DM:
```
casparb, line 150: Should be clearer that this is for Slack and "im" is not just some type we've invented

ndrezn: 🙂 will fix
image

casparb (approved):

PR #6044, opened 2026-09-15
Add Deep Agents to MDA migration guide

ndrezn:
Adds a page for moving an agent from open-source Deep Agents to Managed Deep Agents, prompted by a Reddit thread reading `create_deep_agent` and `define_deep_agent` as duplicate APIs. Trims the overview to the conceptual difference and links out.

The steps convert a project **in place** rather than scaffolding with `mda init`. `mda build` merges its dependencies into the existing manifest, so the project keeps its own layout and dependency set; scaffolding first drops the old dependencies and fails later at graph load.

Verified by converting a Deep Agents project against `managed-deepagents==0.7.1.dev3` — `mda build` and `mda dev` both succeed, with the parameter tables cross-checked against the real `define_deep_agent` signature.

Also fixes two inaccuracies found on the way:

- `going-to-production` described MDA as private preview with a waitlist. It is in public beta.
- `comparison` claimed both modes run "without code changes", which overstated the managed path.

## Review notes

- The parameter mapping tables are the part most likely to drift if the SDK surface changes.
- A section describing the managed-key error messages was removed pending langchain-ai/managed-deepagents-sdk#553, which changes that behavior. See the inline comment.
- One callout remains a workaround for [MDAC-52](https://linear.app/langchain/issue/MDAC-52/mda-build-validates-almost-nothing-so-misconfiguration-surfaces-late) (`mda build` does not validate the definition call) and should come out if it lands.
- The TypeScript path is written from source rather than run end to end, so the TS snippets deserve a closer look than the Python ones.
- Product gaps found while testing are filed as MDAC-46 through MDAC-52, tagged `deepagents-migration`.

PR is agent-assisted.


src/langsmith/managed-deep-agents-migrate.mdx
```diff
@@ -0,0 +1,379 @@
...
+| `cache`, `debug` | `defineDeepAgent({ cache, debug })` | Unchanged. |
+| `name` | `defineDeepAgent({ name })` | Required, and must be a static string. It becomes the assistant ID and the default deployment name. |
+| `systemPrompt` | `instructions.md` | See [Instructions](/langsmith/managed-deep-agents-instructions). |
+| `skills` | `skills/` | One directory per skill, each with a `SKILL.md`. See [Skills](/langsmith/managed-deep-agents-skills). |
+| `memory` | `memory.ts` | Export `defineMemory({ scope: "agent" })`. Omit the file for no durable memory. See [Memory](/langsmith/managed-deep-agents-memory). |
+| `backend` | `sandbox/` | Declare a managed sandbox instead of constructing a backend. See [Sandboxes](/langsmith/managed-deep-agents-sandboxes). |
+| `store`, `checkpointer` | Runtime | Supplied by the managed runtime. Remove both. |
+:::
+
+## Move an agent
```
ndrezn, line 134: A short section sat here describing what happens if you leave a managed parameter in the definition call: `define_deep_agent` rejected it as a bare `unexpected keyword argument`, without naming the project file that takes over.

Removed pending langchain-ai/managed-deepagents-sdk#553, which replaces those errors with a message naming the replacement for each of the six managed keys, in both languages. Documenting the old behavior here would ship a page that is wrong as soon as that merges.

Nothing else was lost. The separate point about `mda build` not inspecting the definition call is untouched by #553 and still stated in the compile step below, since it belongs to MDAC-52.

If #553 stalls, this is worth restoring.

src/langsmith/managed-deep-agents-migrate.mdx
```diff
@@ -0,0 +1,383 @@
...
+
+import ManagedDeepAgentsPublicBetaNote from '/snippets/langsmith/managed-deep-agents-public-beta-note.mdx';
+
+Managed Deep Agents runs the same [Deep Agents](/oss/deepagents/overview) harness you already build against, so moving an existing agent is a repackaging job rather than a rewrite. Your tools, middleware, and subagents carry over as they are. The agent entry changes, and the system prompt, skills, and memory move into project files.
+
+<ManagedDeepAgentsPublicBetaNote />
+
+## Decide whether to move
+
+Managed Deep Agents runs the same harness, so moving does not unlock agent features you cannot build today. Sandboxes, memory that persists across conversations, and every filesystem backend are already in [Deep Agents](/oss/deepagents/overview). Moving changes who operates the infrastructure around the agent.
```
victorm-lc, line 15: not sure if i agree with this reasoning because there are some features it unlocks like native slack, easier integrations, etc.

src/langsmith/managed-deep-agents-migrate.mdx
```diff
@@ -0,0 +1,383 @@
...
+Most agents move without changes beyond the entry file. Four cases need a decision first:
+
+- **Custom backends**: A hand-written `BackendProtocol` implementation has no equivalent. The runtime owns the backend, and the [managed sandbox](/langsmith/managed-deep-agents-sandboxes) is the only choice for giving the agent a filesystem and shell.
+- **Your own store or checkpointer**: Threads, runs, and persistence come from [Agent Server](/langsmith/agent-server-overview). An agent that depends on a specific store implementation needs that logic moved into a tool or middleware.
+- **Custom application code and routes**: Managed Deep Agents hosts the agent, not an arbitrary web application. For custom routes, advanced authentication, or application code around the agent, use a [LangSmith deployment](/langsmith/deployment) instead.
+- **Custom state schemas in Python**: `define_deep_agent` does not accept `state_schema`, so an agent that extends `DeepAgentState` cannot move as is. The TypeScript `defineDeepAgent` does accept `stateSchema`.
+
+If none of these apply, the agent logic transfers as written.
+
+<Prompt
```
victorm-lc, line 52: do we know if this works well or not today

src/langsmith/managed-deep-agents-agent-definition.mdx
```diff
@@ -100,30 +100,38 @@ export const agent = defineDeepAgent({
 
 Configure the system prompt, skills, memory, sandbox, identity, channels, and schedules through their project files rather than the agent definition. See [Project structure](/langsmith/managed-deep-agents-project-structure).
 
+:::python
+`define_deep_agent` does not compile an agent. It returns a definition that `mda deploy` hands to the managed runtime, which supplies the backend, store, checkpointer, memory, skills, and system prompt, then compiles the agent with @[create_deep_agent]. The parameters below are that same surface minus the managed ones. For how the two functions relate, see [Relationship to Deep Agents](/langsmith/managed-deep-agents-overview#relationship-to-deep-agents).
```
victorm-lc, line 104: this is a good point to make but kinda hard to read, is there a more simplified way of saying this since the user can click on the link and view the relationship?

src/langsmith/managed-deep-agents-agent-definition.mdx
```diff
@@ -100,30 +100,38 @@ export const agent = defineDeepAgent({
...
+:::
+
 ## Parameters
 
 :::python
 | Parameter | What it does |
 |---|---|
-| `name=` | Required. Pass a static string that starts with a letter and contains only letters, numbers, underscores, or hyphens, such as `"research-assistant"`.<br /><br /> Managed Deep Agents uses the name as the LangGraph assistant ID and the default LangSmith deployment name. You can override the deployment name with `mda deploy --name` without changing the agent definition. |
+| `name=` | Required. Pass a static string that starts with a letter and contains only letters, numbers, underscores, or hyphens, such as `"research-assistant"`.<br /><br /> Managed Deep Agents uses the name as the LangGraph [assistant](/langsmith/assistants) ID and the default LangSmith deployment name. You can override the deployment name with `mda deploy --name` without changing the agent definition. |
```
victorm-lc, line 116: since assistants isn't a concept you use in MDA could we just say defaults to the graph's name to "deep_agent" and you can override with name field?

src/langsmith/managed-deep-agents-overview.mdx
```diff
@@ -179,6 +179,26 @@ Each part of the agent maps to a file or directory. Add the ones your agent need
 
 For the full layout, see [Project structure](/langsmith/managed-deep-agents-project-structure). Instructions, skills, and optional durable memory are stored in [Context Hub](/langsmith/managed-deep-agents-context-hub).
 
+## Relationship to Deep Agents
```
victorm-lc, line 182: nice!

src/langsmith/managed-deep-agents-migrate.mdx
ndrezn: yep! test drove it and it works pretty well

PR #6050, opened 2026-09-15
fix(self-host-scale): distinguish resource requests from limits

naresh-nampelly:
The table listed "1 CPU, 2Gi per replica" without saying that was the request. The chart's limit is 2 CPU / 4Gi. A customer read it as a limit, set it, and their ingest-queue pods started getting OOMKilled.

This labels requests and limits separately, and adds a table of the shipped defaults for every service. queue wasn't documented at all, nor playground, ace-backend, host-backend, listener or operator.

Checked against chart 0.16.22 and 0.17.0-rc.29.


src/langsmith/self-host-scale.mdx
```diff
@@ -19,19 +19,53 @@ The table below provides an overview comparing different LangSmith configuration
...
+The limits are higher in every case. For the defaults the chart ships, see
+[Default resource requests and limits](#default-resource-requests-and-limits).
+
+A container that exceeds its memory limit is OOM-killed, and one that exceeds its CPU limit is
+throttled. Do not lower either below the shipped default without load-testing the result.
+</Note>
+
 Below we go into more details about the read and write paths as well as provide a `values.yaml` snippet for you to start with for your self-hosted LangSmith instance.
 
+## Default resource requests and limits <a name="default-resource-requests-and-limits"></a>
```
lnhsingh:
```suggestion
## Default resource requests and limits
```

lnhsingh (approved):

PR #6053, opened 2026-09-15
Fix the snippet CodeGroup model substitution

lnhsingh:
## Summary

Two fixes to `scripts/generate_code_snippet_mdx.py`, the script that turns a code sample into a Mintlify `<CodeGroup>` with one tab per provider. They ship together because they touch the same generated files.

1. **The Anthropic tab named a legacy model.** Both tab tables pinned `anthropic:claude-sonnet-4-6`. [Anthropic's model overview](https://platform.claude.com/docs/en/about-claude/models/overview) now lists Claude Sonnet 4.6 under "Legacy models (still available)" and gives `claude-sonnet-5` as the current Sonnet API ID.
2. **The expansion rewrote the wrong line.** The generator took the first `model=` match in a file as the ID every tab rewrites, and a `model` argument does not always hold a routable `provider:model` string.

## The substitution bug

An embeddings constructor takes its own model, and a provider-specific chat class such as `ChatOpenAI` takes a bare ID that provider understands. A RAG sample builds its retriever before its agent, so the embeddings model was the first match and became the ID the tabs varied.

The output was exactly inverted. In `rag-deep-full-js.mdx` on `main`, every tab rewrote the embeddings line and left the agent frozen on Anthropic:

```js
// Google tab, before
const embeddings = new OpenAIEmbeddings({ model: "google-genai:gemini-3.6-flash" });
...
model: "anthropic:claude-sonnet-4-6",
```

```js
// Google tab, after
const embeddings = new OpenAIEmbeddings({ model: "text-embedding-3-small" });
...
model: "google-genai:gemini-3.6-flash",
```

The fix skips an occurrence whose line names an embeddings or `Chat*` constructor, or whose ID reads as an embedding model. It is checked after the `KEEP MODEL` handling, so that marker's semantics do not change, and a test covers that.

## Eleven snippets stop expanding, on purpose

Their only model line sits inside `ChatOpenAI` or `ChatAnthropic`, so their Google tab was offering `ChatAnthropic(model="google_genai:gemini-3.6-flash")`. A sample built on a provider-specific class cannot honestly offer another provider's tab, so those render as a single fenced block now. The affected snippets are the three `deepagents-sandbox-basic-*`, the four `rag-full-snippet-*-setup-*`, and the `tool-return-direct-*` and `tool-runtime-context-thread-*` pairs.

A blunter rule, requiring a colon in the ID, was tried and rejected: it fixed the same cases but dropped CodeGroup coverage from 200 to 163.

## Also fixed

The TypeScript OpenRouter entry read `openrouter:openrouter:z-ai/glm-5.2`, a doubled prefix the Python table does not have. It was invisible while the tabs rewrote the wrong line and is user-visible now.

## Scope

- `scripts/generate_code_snippet_mdx.py`: the two tab tables and one predicate.
- `tests/unit_tests/test_generate_code_snippet_mdx.py`: new, 5 tests. The script had none.
- `src/snippets/code-samples/`: regenerated.

## Verification

- `make test`: 308 passed, including the 5 new ones.
- `make lint` and `make lint_prose` on changed snippets: clean.
- Snippet diff audited line by line. Every changed line is a model string; no incidental reflows.
- No `Embeddings` constructor and no `Chat*` constructor is left holding a `provider:` prefixed ID anywhere under `src/snippets/`.
- The model ID was confirmed against Anthropic's published overview rather than from memory.

## Not covered

- **44 occurrences of `claude-sonnet-4-6` remain** under `src/snippets/`, from the 47 sources carrying a `KEEP MODEL` marker. Reaching those means editing `src/code-samples/` sources, which also changes what `test-code-samples` executes.
- **`src/oss/` (250) and `src/langsmith/` (35) prose** still say `claude-sonnet-4-6`, as does the style-guide example in `CLAUDE.md` and `AGENTS.md` and their four derived copies.
- **`skills-writable-py.mdx` is included.** Its committed comment read "Good for local dev" while its source in `src/code-samples/deepagents/skills.py` reads "Use for local dev", so the generated file had been hand-edited. Regenerated to the source wording. `make code-snippets` on this branch is now a no-op, which it was not on `main`.
- **Nothing in CI catches this class of drift.** `check-generated-files` diffs only `overview.mdx`, so a hand-edit to a generated snippet survives review. Recorded on the Docs automation and agents Notion page.

## AI disclosure

Written by Claude Code, under human review.


npentrel (approved):
Thanks!

PR #6058, opened 2026-09-16
Add CTRLRunMiddleware to the Python middleware integrations

arpanghoshal:
Adds one row to `scripts/data/integration_external_docs.yaml` for
[`ctrlrun-langchain`](https://pypi.org/project/ctrlrun-langchain/), a `wrap_tool_call`
middleware that gates an agent's tool calls against a policy.

## What it does

`wrap_tool_call` hands the middleware the call itself, so the handler runs *as* the executor
rather than being observed afterwards. That makes the decision and the outcome the same call,
and gives three properties:

- a refused call never reaches the tool
- an effect key executes at most once, across processes sharing a store
- a call whose outcome was never established blocks its own retry until a human resolves it,
  rather than being retried into a duplicate

Plus human approval where the policy requires it, and a hash-chained receipt for every
decision, refusals included.

## Details

| | |
|---|---|
| PyPI | `ctrlrun-langchain` 1.0.0 |
| Source | [CTRLRun/ctrlrun/adapters/langchain](https://github.com/CTRLRun/ctrlrun/tree/main/adapters/langchain) |
| Requires | `langchain>=1.0,<2.0`, `ctrlrun>=0.12,<0.13` |
| License | Apache-2.0 |

Added to the external-docs YAML rather than as a hosted guide, per that file's header: it is
under 50,000 monthly downloads and not featured. `docs_url` points at the adapter's README,
following the file's stated link priority where there is no separate partner docs site.

## Validation

```
tests/unit_tests/test_refresh_integration_downloads.py   22 passed
```

including `test_validate_external_docs_urls_accepts_repo_yaml`. The YAML parses and the row
round-trips.

Verified installable from PyPI into a clean environment:

```
$ pip install ctrlrun-langchain
ctrlrun            0.12.2
ctrlrun-langchain  1.0.0
langchain          1.4.0
```


arpanghoshal:
Thanks — I missed that listings go through the issue form. Refiled as #6059 using the integration submission template, so the automation can open the review PR from the form fields.

Closing this one to keep it out of your queue rather than leaving a duplicate open.

PR #5896, opened 2026-09-06
Add maskflow-langchain (PII anonymizer middleware)

sodanisomya:
Lists [`maskflow-langchain`](https://pypi.org/project/maskflow-langchain/), following the same four-file pattern as the recently merged [#5749 (piighost)](https://github.com/langchain-ai/docs/pull/5749):

- `packages.yml` — package entry
- `scripts/data/integration_external_docs.yaml` — row under `python.middleware`, next to the other PII-anonymization entries (`PIIAnonymizationMiddleware`, `NoPII`)
- `src/oss/python/integrations/providers/all_providers.mdx` — provider card
- `src/snippets/oss/python-middleware-downloads.mdx` — download-table row (new package → `N/A`; the weekly refresh will fill the badge)

`maskflow-langchain` is a reversible PII anonymizer / deanonymizer, a drop-in for `langchain-experimental`'s Presidio anonymizer (same `.anonymize` / `.deanonymize` / `.deanonymizer_mapping` / save-load), so a chain migrates with one import change. It adds the Indian identifiers most PII tooling misses (Aadhaar, PAN, GSTIN, UPI, IFSC, ABHA, Indian names and addresses), a streaming-aware deanonymizer `Runnable`, and an optional leak-guard callback that fails a call closed if PII reaches the model. MIT, runs locally, no external service.

PR #6064, opened 2026-09-16
integrations: add CTRLRun (middleware)

arpanghoshal:
Adds [`ctrlrun-langchain`](https://pypi.org/project/ctrlrun-langchain/) to the Python middleware table, following the community route in `scripts/data/integration_external_docs.yaml`.

CTRLRun gates every tool call through `wrap_tool_call` against a policy file: a refused call never reaches the tool and the model is told which rule refused it, an effect key executes at most once across processes sharing a store, a tool that raises leaves the effect unresolved rather than retried, an `approve` decision holds the call for a person, and every decision leaves a receipt. The kernel is Apache-2.0: https://github.com/CTRLRun/ctrlrun.

Same four files as the other community middleware entries (#5749, #5896): the external-docs row, `packages.yml`, the all-providers card, and the snippets regenerated with `scripts/refresh_integration_downloads.py --language python --component middleware --write`. `--check-docs-urls` passes. The `docs_url` is the integration guide on our docs site, per the partner-docs-first rule in the YAML header.


arpanghoshal:
Tagging @mdrxy for the review.

npentrel (approved):

PR #6081, opened 2026-09-17
docs: TypeSafe provider page for `langchain-typesafe`

sydney-runkle:
A provider page for TypeSafe, covering its [Jev](https://docs.typesafe.ai/models) model, the released `TypeSafeClassifier`, and the two agent middleware currently in review.

Flow: **Jev → Setup → Quickstart → Decision types → Agent middleware → Reference.** 60 lines of code, 93 of prose.

## Built against the released package

Everything uses `langchain-typesafe` rather than the vendor SDK. An earlier revision of this branch carried three hand-written middleware implementations over `typesafe-sdk`; those are gone, replaced by usage of the shipped classes, which took the page from 395 lines to 153.

- **Quickstart** — one `TypeSafeClassifier` asking a `Noul`, a `Choice`, and a `Score` in a single request, then reading `nouls`, `choices`, and `scores` off the response.
- **Decision types** — the three [primitives](https://docs.typesafe.ai/primitives), what each returns, and when to reach for it, including the trap the TypeSafe docs call out: a `Noul` of `0.5` is an even split between yes and no, not "medium", so a spectrum wants a `Score`.
- **Middleware** — `ModelRouterMiddleware` and `AutoModeMiddleware` from `langchain_typesafe.experimental.middleware`.

## Accuracy notes

**The middleware are not in `0.0.1a1`.** `langchain_typesafe.experimental` does not exist in the published alpha — its dependencies do not even include `langchain`. The page says so in a note and links #40543 and #40545, so nobody installs the alpha expecting them.

**Described behavior comes from running the code, not from reading it.** Each middleware was exercised against mock transports on its own PR branch: routing stores a full `ChoiceAnswer` in state and selects the right model, and the tool gate blocks at or above the threshold, allows below it, bypasses unlisted tools, and accepts both tool names and `BaseTool` instances plus custom `NoulCriteria`. The quickstart was run against the published `0.0.1a1`.

**Two API details the page has to spell out**, and which may be worth fixing in the package instead:

- `ClassificationResponse` is the classifier's return type but is not re-exported from the package root — it is in `types.__all__` and absent from `__init__.__all__`, so it has to be imported from `langchain_typesafe.types`.
- The error types are only importable from `langchain_typesafe.client`.

**Model routing has no default route.** It classifies and then indexes `self.models[answer.choice]`, with no fallback, and the latest-human-message lookup is a bare `next(...)`. The page therefore does not promise fail-open behavior. Worth deciding whether that is intended before release.

Reference-link (`@[...]`) syntax is deliberately not used: it appears on only 6 of 100 provider pages, and these symbols are unlikely to be in the API reference for a fresh alpha.

## Placement

`integrations/providers` rather than `integrations/middleware`, since every page in the latter carries an `integration:` block with a `pypi` field feeding a generated download table. Registered in `all_providers` between TypeDB and Typesense, and deliberately not added to the hand-picked Popular Providers group.

---

*Drafted with assistance from Claude Code, and reviewed by me. Middleware designs are Hunter Lovell's from #40543 and #40545.*


ndrezn:
Some other examples that could be interesting:
- No-progress detection - send the current state to Jev every few turns and have Jev determine (with [some degree of confidence](https://en.wikipedia.org/wiki/Halting_problem)) if an agent is stuck in a loop, if yes, then fail out
- LangSmith labelling - `after_agent` middleware to classify the run outcome and failure mode, so the developer has nice feedback that lands in LangSmith. Boom, free eval datasets
- Tool response gating - expand the tool-risk gating to have an evaluator for if the tool returned some risky response, to protect from unsafe injections
- Relevance-based context pruning - score all of the responses in a thread, then drop the responses that aren't relevant anymore
- Memory gating - score a response for whether we should consider storing a summary of it in memory

aolsenjazz (approved):
l'stamp

PR #6097, opened 2026-09-18
docs(langgraph): set max concurrency at the config top level

Xx-173:
## What changed

`src/oss/langgraph/use-graph-api.mdx`: the "Set max concurrency" tip told readers to pass `max_concurrency` inside `configurable`, where it is silently ignored.

- Python: `graph.invoke({"value_1": "c"}, {"configurable": {"max_concurrency": 10}})` -> `graph.invoke({"value_1": "c"}, {"max_concurrency": 10})`
- TypeScript: `{configurable: {max_concurrency: 10}}` -> `{ maxConcurrency: 10 }`
- Both tips now state that the key is standalone, not a `configurable` entry.

## Why

The async executor in `libs/langgraph/langgraph/pregel/_executor.py` reads the value with `config.get("max_concurrency")`, a top-level config key. A `Send` fan-out therefore runs unbounded when the key sits under `configurable`, matching the unset case exactly. The repo's own `test_max_concurrency` in `libs/langgraph/tests/test_pregel_async.py` passes it at the top level, as does the `recursion_limit` example on this same page.

On the JS side, `maxConcurrency` is listed as a top-level config field in `libs/langgraph-core/src/pregel/utils/config.ts` and read as `config.maxConcurrency` in `libs/langgraph-core/src/pregel/index.ts`.

Reported in langchain-ai/langgraph#8920.

## How this was checked

- Read the value sites in the LangGraph and LangGraph.js sources listed above.
- `vale --glob='!{**/node_modules/**,src/code-samples/**}' src/oss/langgraph/use-graph-api.mdx` reports 0 errors, 0 warnings, 0 suggestions (Vale 3.17.1, the version pinned in `.mise.toml`).
- Grepped `src/` for the `configurable` placement: no other page repeats it.

## Review notes

- The TypeScript tip previously spelled the key `max_concurrency` (snake case). JS uses `maxConcurrency`, so both the prose and the example were corrected. Flagging in case the JS runtime also accepts the snake-case spelling.
- No code change. This corrects the docs to match existing behavior rather than changing behavior, which is what langchain-ai/langgraph#8944 attempted and did not land.

## AI involvement

Drafted with AI assistance. The source read sites and the Vale run were verified directly.


Xx-173:
@npentrel @lnhsingh Ready for review! Thank you!

PR #4009, opened 2026-05-14
docs: fix delete_old_messages example to properly demonstrate deletion effect

uditstocks:
## Overview
The existing `delete_old_messages` example in the short-term memory documentation failed to properly demonstrate 
the deletion effect. After deleting the earliest two messages, the agent could still recall the user's name "Bob" because
the AI's own first response had echoed the name - and that response survived the deletion window.

This PR fixes the example by inserting an unrelated conversation turn (a cat poem) 
between the name introduction and the name recall query. This naturally pushes the 
name-containing messages out of the window, so when the agent is asked "what's my name?"
it correctly responds that it doesn't know.

## Type of change
[Update existing documentation]

## Checklist
- [x] I have read the contributing guidelines
- [x] I have tested my changes locally using `docs dev`
- [x] All code examples have been tested and work correctly

## Related issues/PRs
closes #3894

## Additional notes
The fix is minimal and surgical - only the example code and output block are changed. 
No API behavior, no logic, no navigation changes. The updated output block is taken 
from an actual local run to ensure accuracy.

Issue #3894, opened 2026-05-07
[langchain]: 'delete_old_messages' The deletion effect was not effectively demonstrated.


AA-G-AA:
### Type of issue

issue / bug

### Language

Python

### Description

https://docs.langchain.com/oss/python/langchain/short-term-memory#delete-messages

## Issue Description

The current `delete_old_messages` example in the official documentation fails to effectively demonstrate the deletion effect — the agent still remembers the user's name "Bob" after the earliest two messages are deleted.

## Root Cause

In the original example, the AI's first response `"Hi Bob! Nice to meet you..."` repeats the user's name "Bob". After deleting both the user's first message and the AI's first response (`messages[:2]`), the name "Bob" remains in the AI's second response, allowing the agent to still "remember" the user's name.

This does not properly demonstrate the expected "information loss after deletion" behavior.

## Suggested Improvement

Insert an unrelated conversation turn between the name introduction and the name recall query. This allows the earliest two messages (containing the name info) to be naturally deleted before the agent is asked to recall the name.

**Improved code:**

```python
for event in agent.stream(
    {"messages": [{"role": "user", "content": "hi! I'm bob"}]},
    config,
    stream_mode="values",
):
    print([(message.type, message.content) for message in event["messages"]])

# ✅ Insert an unrelated turn
for event in agent.stream(
    {"messages": [{"role": "user", "content": "write a short poem about cats"}]},
    config,
    stream_mode="values",
):
    print([(message.type, message.content) for message in event["messages"]])

# Now the agent should NOT remember the name
for event in agent.stream(
    {"messages": [{"role": "user", "content": "what's my name?"}]},
    config,
    stream_mode="values",
):
    print([(message.type, message.content) for message in event["messages"]])

Actual output (proof of improvement):

text
[('human', "hi! I'm bob")]
[('human', "hi! I'm bob"), ('ai', 'Hi Bob! Nice to meet you...')]
[('human', "hi! I'm bob"), ('ai', 'Hi Bob!...'), ('human', 'write a short poem about cats')]
[('human', "hi! I'm bob"), ('ai', 'Hi Bob!...'), ('human', 'write a short poem about cats'), ('ai', 'There once was a cat...')]
[('human', 'write a short poem about cats'), ('ai', 'There once was a cat...')]
[('human', 'write a short poem about cats'), ('ai', 'There once was a cat...'), ('human', "what's my name?")]
[...]

PR #6122, opened 2026-09-20
docs: fix undefined config variable in delete messages example

xiaoye-2018:
## Overview
The "Full example: delete messages" code block in the short-term memory
documentation uses an undefined `config` variable. Running the example
as-is raises `NameError: name 'config' is not defined` when calling
`app.stream_events(..., config, ...)`, since `config` is never assigned
anywhere in the snippet (unlike the "trim messages" and "summarize
messages" full examples on the same page, which both define it).

## Type of change
Update existing documentation / bug fix

## Checklist

## Related issues/PRs
Not related to #3894 / PR #4009 (which fixed a different issue in the
same example — the deletion-effect demonstration). This is a separate,
previously unreported issue: the missing `config` definition.

## Additional notes
Minimal fix — adds one line:

    config = {"configurable": {"thread_id": "1"}}

right after `app = builder.compile(checkpointer=checkpointer)` and before
the first `app.stream_events(...)` call. No other content changed.


npentrel (approved):
Thanks!

PR #6137, opened 2026-09-21
docs: Fix Codex plugin hooks setup and Claude Code plugin updates

QuentinBrosse:
The published half of two reports:

- langchain-ai/langsmith-codex-plugins#51
- langchain-ai/langsmith-claude-code-plugins#104

## `trace-with-codex.mdx`

The setup step told readers to set `[features] plugin_hooks = true`. That flag was removed in openai/codex#22552 and is silently ignored today (`Stage::Removed` in `codex-rs/features/src/lib.rs`). Plugin hooks follow plugin enablement, `[features] hooks` is stable and on by default, and trust is per hook via `/hooks` — so the troubleshooting bullet "confirm `plugin_hooks = true`" sent people chasing a no-op.

Replaced with plugin enablement + `/hooks` trust, and a line explaining what happened to the flag for anyone who still has it in their config.

## `trace-claude-code.mdx`

"To update the plugin" listed only `/plugin marketplace update` + `/reload-plugins`. That refreshes the marketplace catalog clone but leaves the install record pinned to the version directory it was installed from, so the hooks keep running the old bundle after a version bump.

Added `claude plugin update langsmith-tracing@langsmith-claude-code-plugins`, and a note pointing at `/plugin` → **Marketplaces** → **Enable auto-update** as the path that refreshes catalog and install together (off by default for third-party marketplaces).

Matching repo README fixes: langchain-ai/langsmith-codex-plugins#59 and langchain-ai/langsmith-claude-code-plugins#109.


ejaimez14 (approved):

src/langsmith/trace-with-codex.mdx
```diff
@@ -24,16 +24,17 @@ Add the marketplace using the Codex CLI:
...
 
 ```toml
-[features]
-plugin_hooks = true
-
 [plugins."tracing@langsmith-codex-plugins"]
 enabled = true
 ```
 
+Plugin hooks follow plugin enablement, so there is no separate flag to set. `[features] plugin_hooks` was removed in [openai/codex#22552](https://github.com/openai/codex/pull/22552) and is ignored if it is still in your config. `[features] hooks` is stable and on by default.
```
QuentinBrosse: I don’t think customers need to know these details. Should we remove this?

PR #6140, opened 2026-09-22
docs: add Amazon Bedrock AgentCore Web Search integration

sundargthb:
## Description

Documents the AgentCore Web Search tool that was added to `langchain-aws` in langchain-ai/langchain-aws#1275.

Three changes, matching what the AgentCore Browser and Code Interpreter integrations already do:

- New page `src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx`
- A row in the `## Search` table on `src/oss/python/integrations/tools/index.mdx`
- A `### Amazon Bedrock AgentCore Web Search` section on `src/oss/python/integrations/providers/aws.mdx`

`create_web_search_toolkit` is in the published `langchain-aws==1.7.9`, so `pip install -U langchain-aws` gives a reader every import on this page.

## On the integration listing form

This is not a new integration listing. `langchain-aws` is already a hosted provider in this repo, with `providers/aws.mdx` and hosted tool pages for AgentCore Browser and Code Interpreter, and it has no row in `scripts/data/integration_external_docs.yaml`. This adds a third tool page next to those two, so there is no external listing to generate. If maintainers would rather it went through the form anyway, say so and I will refile it.

## Two things this page carries that the sibling AgentCore pages do not

**It needs a setup step before instantiation.** Code Interpreter and Browser are data plane APIs, so their pages go straight from install to the factory call. Web search reaches the service through a Gateway connector target, so a gateway has to exist first. The page shows that with `create_gateway_and_wait` plus `create_web_search_target`, which is three calls and done once per account.

**The IAM story is two permissions on two different principals**, and getting it wrong is the most common setup failure:

- the caller needs `bedrock-agentcore:InvokeGateway` on the gateway ARN
- the gateway's execution role needs only `bedrock-agentcore:InvokeWebSearch` on the connector, and nothing else

The page also notes the failure mode where the execution role's trust policy narrows `aws:SourceArn` to the wrong region. That is accepted when the gateway is created and only surfaces on the first search, as `Failed to obtain execution role credentials`, which reads like an entitlement problem rather than an IAM one.

## Accuracy

Every code sample is taken from either the merged toolkit's own docstrings or from the sample notebook, which was executed end to end against a live gateway in `us-east-1`. The gateway and target creation snippet is the executed version, not a reconstruction.

- Factory is synchronous and returns `Tuple[WebSearchToolkit, List[BaseTool]]`, unlike the async browser and code interpreter factories
- One tool, `web_search`, with `query`, `max_results`, `include_domains`, `exclude_domains`, `published_after`, `published_before`
- Failures raise `ToolException`; `handle_tool_error` is the documented opt-in for letting the model retry its own query
- The three regions are the ones on the [service page](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html), and the page points there for the current list rather than treating it as fixed
- The page calls out that web search needs `bedrock-agentcore>=1.23.0`. That floor lives only in the `[tools]` extra, and `langchain_aws/tools/__init__.py` wraps the import in `try/except ImportError`, so an older pinned install silently lacks the export instead of erroring

## Related

- langchain-ai/langchain-aws#1275 (merged, the tool)
- langchain-ai/langchain-aws#1280 (open, the sample notebook this page's flow is drawn from)


src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
```diff
@@ -0,0 +1,225 @@
...
+<CodeGroup>
+    ```bash pip
+    pip install -U langchain-aws bedrock-agentcore
+    ```
+    ```bash uv
+    uv add langchain-aws bedrock-agentcore
+    ```
+</CodeGroup>
+
+Web search needs `langchain-aws` 1.7.9 or later and `bedrock-agentcore` 1.23.0 or later. The examples on this page were run against `langchain-aws` 1.7.9 with `bedrock-agentcore` 1.23.1.
```
michaelnchin, line 45: Nit: remove second sentence as it's bit repetitive

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+[Amazon Bedrock AgentCore Web Search](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html) gives agents current information from the web with a source URL on every result. It is an AWS service-managed connector, so it authenticates with the AWS credentials you already have and takes no search API key of its own.
+
+## Overview
+
+### Integration details
+
+| Class | Package | Serializable | [JS support](https://js.langchain.com/docs/integrations/tools/) |  Version |
+|:------|:--------|:------------:|:---------------------------------------------------------------:|:--------:|
+| [`WebSearchToolkit`](https://github.com/langchain-ai/langchain-aws/tree/main/libs/aws/langchain_aws/tools) | [`langchain-aws`](https://pypi.org/project/langchain-aws/) | ✅ | ❌ | ![PyPI - Version](https://img.shields.io/pypi/v/langchain-aws?style=flat-square&label=%20) |
```
michaelnchin, line 18: Serializable should be marked X, since this is the default for `is_lc_serializable` and we don't override it

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
```diff
@@ -0,0 +1,225 @@
...
+```python
+from langchain_aws.tools import create_web_search_toolkit
+
+toolkit, tools = create_web_search_toolkit(
+    region="us-east-1",
+    gateway_id="my-web-search-gateway-abc123",
+)
+```
+
+Address the gateway by `gateway_id`, by `gateway_arn`, or by `gateway_endpoint` if you already know the MCP URL. A `gateway_arn` carries its own region, so pass `region` alongside a `gateway_id`, or to override the region in an ARN.
```
michaelnchin, line 105: This part about `gateway_id` reads a bit awkwardly + I think overriding region for a gateway ARN is not useful since the gateway is locked to the region in the ARN already. I suggest this instead be:

```suggestion
Address the gateway by `gateway_id`, by `gateway_arn`, or by `gateway_endpoint` if you already know the MCP URL. A `gateway_arn` carries its own region. With a `gateway_id`, pass `region` to match the gateway's region, or it will default to `us-east-1`.
```

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
```diff
@@ -0,0 +1,225 @@
...
+print(results)
+```
+
+Results come back as numbered text blocks, one per hit, each carrying a title, a URL, a publication date when the index reports one, and an extract:
+
+```text
+1. boto3 1.41.2 release notes
+   URL: https://github.com/boto/boto3/blob/develop/CHANGELOG.rst
+   Published: 2026-09-18T00:00:00Z
+   Adds support for ...
```
michaelnchin, line 133: Nit: Tested and found that the real `publishedDate` format is different (also per the [blogpost](https://aws.amazon.com/blogs/machine-learning/introducing-web-search-on-amazon-bedrock-agentcore/)). Can change this to:

```suggestion
1. boto3 1.43.56 release notes
   URL: https://github.com/boto/boto3/blob/develop/CHANGELOG.rst
   Published: 12:31PM, Friday, July 24 2026, PDT
   Adds support for ...
```

michaelnchin (commented):
Thanks @sundargthb, tested the examples and looks good functionally, couple of minor comments.

src/oss/python/integrations/providers/aws.mdx
```diff
@@ -220,6 +220,42 @@ result = await agent.ainvoke(
 await toolkit.cleanup()
 ```
 
+### Amazon Bedrock AgentCore Web Search
+
+>[Amazon Bedrock AgentCore Web Search](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html)
+> gives agents current information from the web with a source URL on every result, authenticated with your AWS credentials rather than a search API key.
```
lnhsingh:
```suggestion
[Amazon Bedrock AgentCore Web Search](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html) gives agents current information from the web with a source URL on every result, authenticated with your AWS credentials rather than a search API key.
```

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
sundargthb: Removed. The floor versions in the first sentence cover it.

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
sundargthb: Changed to ❌. Checked it rather: `WebSearchToolkit` is a plain class, and the tools are `StructuredTool`, which inherits the `Serializable` default of `False` without overriding it.

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
sundargthb: Applied your suggestion. Your reading matches the source: `DEFAULT_REGION` is only filled in when neither `region` nor `gateway_arn` is passed, so passing a region alongside an ARN can only point the client at a region the gateway is not in.

src/oss/python/integrations/tools/bedrock_agentcore_web_search.mdx
sundargthb: Applied. Confirmed the same format in the blog sample, so the example was wrong.

src/oss/python/integrations/providers/aws.mdx
sundargthb: Applied.

sundargthb:
Thanks @michaelnchin and @lnhsingh. All five comments applied in 73b4f69.

michaelnchin (commented):
LGTM 👍

lnhsingh (approved):

